PR #4, opened 2020-04-21
update example for webpack

simskij:


simskij:
@ppcano looks like someone removed my privileges in the repo, so I cant merge it myself. Should be merged to dev and then to master.

ppcano:
@simskij you should have permission now to both `branches`.

PR #8, opened 2020-04-22
Updated sample script URL

hubail:
Older URL had no scheme and gave 404 due to GitHub shifting user contents to a separate domain.


imiric:
Hi, thanks for submitting this, though this change is not necessary as k6 has special script loading for github.com that resolves to the full URL. So even though `github.com/loadimpact/k6/samples/http_get.js` returns a 404 in the browser, these examples will work fine with k6.

hubail:
Got it, thanks for clarification, I was inspecting the script before running.

PR #13, opened 2020-04-29
fixed broken id generation if title starts with a digit

BolotskyDev:
This PR resolves this problem 
image


ppcano:
@BolotskyDev I have merged this PR but it hasn't fixed the problem.

PR #16, opened 2020-05-04
Feature: docs banner update

BolotskyDev:
**Subject**
This PR introduces design updates of a docs banner component according to latest changes motivated by Notion task. 

A bit more verbose:
- introduced `cta-doc` shared component which encapsulates banner logic including 'with image' option
- updated `cta` block
- updated `doc-welcome-page/cloud` component to use `cta-doc`
- updated `cloud` page to use `cta-doc`

**Screenshots**
image
image

**Note**: 
As far as I can tell, this is perfectly mergeable PR, but I am not sure that `k6-docs` contains only 2 entries of banner component. Would be nice if we replace them all before merging. Hope @ppcano can point out the rest of the pages with banners if there are such.


ppcano:
Thanks @BolotskyDev 

>  but I am not sure that k6-docs contains only 2 entries of banner component.

Merged. There were only those two.

PR #18, opened 2020-05-04
Fix Values to match Images

bt-dd:
There were some issues in the documentation where the config description didn't match, believe it should be corrected with these changes?


simskij:
Hi there @bt-dd!

Is there any specific reason why you closed this PR? At a glance this seems like a reasonable change.

Thanks,
Simme

bt-dd:
@simskij I made set target to master but saw other PR's were into develop, made a new PR and it got merged here :) #20

simskij:
Ah, my bad! Awesome then! 😅👍🏻

PR #20, opened 2020-05-04
Fix Load Testing example Values to match Images

bt-dd:
There were some issues in the documentation where the config description didn't match, believe it should be corrected with these changes?


ppcano:
Thanks @bt-dd!

PR #35, opened 2020-06-23
Add per-URL data sent/recv example

robingustafsson:
This PR adds an example script to the `Counter` custom metric docs to show how to track data sent and received per-URL.

As discussed in Slack this was added in-context rather than as an example in the "Examples" section.

### DIFF
--- a/src/data/markdown/docs/01 guides/02 Using k6/02 Metrics.md
+++ b/src/data/markdown/docs/01 guides/02 Using k6/02 Metrics.md
@@ -131,6 +131,53 @@ The value of `my_counter` will be 3 (if you run it one single iteration - i.e. w
 
 Note that there is currently no way of accessing the value of any custom metric from within JavaScript. Note also that counters that have value zero (`0`) at the end of a test are a special case - they will **NOT** be printed to the stdout summary.
 
+**Examples**
+
+*Traking data sent and received per-URL*:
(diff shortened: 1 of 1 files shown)


ppcano:
If we want to place this example in context - I think - it should be placed at https://k6.io/docs/javascript-api/k6-metrics/counter#examples.

But regarding the logic of the `Counter` metric,  correct me if I am wrong,  this example does not include anything that is not already included in the `Advanced Thresholds` example.

- tagging the metric value
- threshold on the counter metric
- threshold on the counter metric with tags

Because of this reason, I think this example should NOT be added as an example of the Counter metric 🙏🙏


--------------------------------

Then I realized that we cannot know the transmitted data per URL/request because `data_sent` and `data_received` metrics are not "tagged" with the URL. And this example provides a solution to this problem.


Now, I think this example should be added to the Examples list including its description. 🙏🙏

> Two custom metrics to track data sent and received. We will tag data points added with the corresponding URL
 so we can filter these metrics down to see the data for individual URLs and set threshold across all or per-URL as well.


Additionally, we could reference this example when describing the built-in metrics `data_received` and `data_sent`. Thoughts?

Screen Shot 2020-06-25 at 1 18 41 AM

ppcano:
Supercedes by #40

PR #47, opened 2020-06-26
Cleanup options tables, add default column

imiric:
The default values in the Options page were just listed after the table, but it makes more sense to make them a part of it.

In some cases it will cause the table to render squished columns, or to need scrolling, but it's not a big problem IMO.


ppcano:
@sniku would you mind to review/merge this one? Thanks.

PR #58, opened 2020-07-14
[refactor] Suggestions for changes

mostafa:


Menda:
Thanks for also adding my suggestion!

PR #68, opened 2020-07-29
Breaking changes: Improving perf and bugfixes

BolotskyDev:
### Shortly
This PR contains changes related to performance improvements. They were made possible by adjusting the way of how we are processing doc pages content: new custom components API, adjustments of content rendering with a usage of `MDXProvider`,  reducing number of to-be-hot-replaced components, fixing a bunch of `md` files to match new API. 

Additionally this PR fixes a few old annoying errors and a bunch of minor improvements (style, certain component's structure etc)

#### Major
The main reason why do our content-rich pages like `Options` became somewhat problematic in terms of performance is switching from `gatsby-transformer-remark`, which handles simple `md` pages, to `gatsby-plugin-mdx`, whose job is to handle `mdx` pages with custom `jsx` content. 

This is how the flow looked like with the former (`gatsby-transformer-remark`):
1. plugin had been giving us a correct `html` representation of `md` content during build time
2. the page was generated using that content
3. during runtime we were hot-replacing all the custom components we need

This is how the flow looks like with the latter (`gatsby-plugin-mdx`):
1. plugin no more gives us `html` representation, but rather a function, which we have to pass to `MDXRenderer` in order to display our content correctly. 
2. the page was generated, but only partially
3. during runtime `MDXRenderer` runs this function and displays us initial content
4. then we are coming to hot replace our custom component.

_**When it comes to really heavy pages like aforementioned `Options`, steps 3-4 is our bottleneck, since we are rendering everything twice right away**_

In order to solve it I've started to use `MDXProvider` exposed by the same plugin. It takes a mapping object selector-component and replaces everything it can match at step 3, thus taking all the heavy lifting from step 4, _declaratively_. 

I had to change most of the custom components we are using in hot replacement in order to make it work, adjust the way we are using them in `md` files and make some changes in hooks replacement logic as well.

_**Unfortunately, I wasn't able to exclude from hot replacement our code blocks at the moment for the following reasons:**_ 

1) Specifying `code` component for native replacement causes processing of `inline-code` as well, though it shouldn't. `CodeGroup` can not handle `inline-code`, and I am not able to programmatically distinguish what comes as a content prop.
 
2) In order to handle wrapped in `div` code groups I'd have to replace every single entry of `div` with `jsx` component, which is really expensive task since we have so many of them across the project 

3) Even if I had have replaced them all, there is one more problem: during processing of a `code` node by provider it removes wrapper, so the data looks like a ton of html tags (look inside any code component we have in prod to understand what I mean). It is not a problem if you know for sure that those tags belong to a single code blocks, but imagine processing of code group with 3 code blocks - I still have no idea how to handle that mess 😞   

Given that, I'd like to get back to this problem after we update our projects deps. I expect updated deps to fix 1), so I'd be able to work on handling 2) and 3), if that would make sense by this time. Moving code blocks out of hot replacement will give us a tremendous perf boost
 
#### Minor 
- fixed a bunch of syntax errors in `md` pages, replaced old API usages with new ones. See affected pages in diff.

#### Bug fixes
- resolved this long annoying client error with `select` component that we are using for mobile navigation
- resolved error that @sniku caught with Sentry: `Sticky` component couldn't invoke a `getBoundingClientRect` on `TableOfContent` inside when we were hiding it if no landmarks (h2 headers) happened to be on the page 

## Notes

_Sorry for the giant PR, but I couldn't deliver it incrementally since almost everything is highly coupled and won't work properly otherwise. I did my best resolving all the possible issues, md custom components syntax mismatches and keeping the branch up-to-date, so that the only thing you have to do is a skim check of git diff and a local manual test to ensure everything is still on track (and, of course, even better)._

#### Common
1) You don't need to import `jsx` components in `md` files anymore. They are provided globally with `MDXProvider`, here is how it works:

Before:

```mdx
// some frontmatter thing
import LdScript from '...'
<LdScript script=""/>
```

After:

```mdx
// some frontmatter thing
<LdScript script=""/>
```

In order to expose some new components you have to pass them into `components` prop of `MDXProvider`. More on that see in [gatsby docs](https://www.gatsbyjs.org/docs/mdx/customizing-components/). Do not expose though if it is like one-off case, just import directly. 

**Important**: please, do not paste pieces of generated html code from pages in production directly into `md` file! That would cause multiple issues in case there is a svg or two. Abstract away ( even if it is one-off case) this piece of code in source component and import directly into `md`. See how it is done with `IntegrationCiIconBlock` [here](https://github.com/loadimpact/k6-docs/blob/develop/src/data/markdown/docs/01%20guides/05%20Testing%20Guides/02%20Automated%20performance%20testing.md#L10).

#### Blockquote

Old API:

```md
// default quote
> Quote
// warning quote
<div class="doc-blockquote" props='{"mod": "warning"}'>

> Quote

</div>
```

New API:

```md
// default quote
> Quote
// warning quote
<Blockquote mod="warning">

> Quote

</Blockquote>
```

_Also keep in mind that we have extended range of headers that being transformed to anchors, `h2` -> `h2`, `h3`, so please do not use `h3` headers for blockquotes. Use `h4`, we also added styles for this one._

#### LdScript

An example of usage:

```md
// some other mdx content
<LdScript script='{"@context": "http://schema.org", "@type": "HowTo", "name": "How to automate performance testing", ....}'/>
```

## Steps to test (finally)

1. Skim check the PR diff, pay attention to changes in `md`files  in order to prevent possible information regressions
2. Pull the changes, run `npm install && npm start`
3. Open dev console, see that no error related to `select` component popped out.
4. Move to any api related page where no ToC is generated, see that no error related to `Sticky` popped out as well
5. Move around, interact with UI a bit, check for possible UI/UI-interaction regressions.


BolotskyDev:
@ppcano Thank you, nice catches! 

1. Minor problem with props, resolved

2. This is interesting: in `notifications` page there are some headers that wrapped in backticks, like "\`test\`". During processing this header is treated like `inline-code`, so our `heading` component can't handle that. I've escaped them and adjusted `slugify` function, so they are working now. 

I guess we are good to test it in `staging` if you confirm that the bugs are resolved indeed 👍

BolotskyDev:
Also this PR fixes `gatsby-plugin-sass` settings which caused that bug with links on hover you've reported recently. I'll fix them in other projects as well.

BolotskyDev:
The last reported problem with broken pages that contain `http` path is resolved, merging into `develop`.

PR #92, opened 2020-08-19
Update apt-key for linux installation

nayema:
The initial installation suggestion `wget -q -O - https://bintray.com/user/downloadSubjectPublicKey?username=bintray | sudo apt-key add -` is not suitable for trying to install behind firewalls. This was highly difficult when I was trying to make a docker image with k6 installed. I was able to get around this by following bintray suggestion from https://bintray.com/cublinux/deb.

### DIFF
--- a/src/data/markdown/docs/01 guides/01 Getting started/02 Installation.md
+++ b/src/data/markdown/docs/01 guides/01 Getting started/02 Installation.md
@@ -2,16 +2,35 @@
 title: 'Installation'
 ---
 
-## Linux (deb and rpm packages)
+## Linux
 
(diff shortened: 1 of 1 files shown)


na--:
Hmm I think we shouldn't replace the original instructions, given that they feature the actual key fingerprint so are likely much more secure. The alternative instruction should be a troubleshooting hint somewhere, not replace the original ones which will work for most people.

simskij:
@na--
My thought as well. 👍🏻

@nayema 
I will modify it a bit to include both alternatives before merging it. Thank you so much for your contribution, its super helpful! 🙏🏼

simskij:
Thanks again for your contribution! 🥳  🙌

nayema:
Thank you @simskij and @na-- !

PR #94, opened 2020-08-26
Update code snippet to match text

jmcduffie32:
In the description of using failed checks to add an error metric, the text does not match the code snippet. The text reads `Note also the expression `check() || errorRate.add(1).` but that code is not used. This PR updates the code snippet to match the description.


simskij:
Thank you for your contribution!

I need to check the behavior before merging this as I have a vague recollection of check always returning `true` unless the check itself throws.

simskij:
It indeed looks like `check` should return false in case a check expression is evaluated to false. I must've been thinking of something else! 😅 

With that said, it also needs to add `0` to the rate in case it returns true, so I'm doing a commit to fix that before merging. Thanks again!

PR #107, opened 2020-09-08
Correct the syntax of the "entry" webpack config

Senney:
The provided webpack sample had invalid javascript in it. I've updated the example to match the syntax required per https://webpack.js.org/concepts/entry-points/


na-- (approved):
LGTM, though I haven't actually tried running webpack with it.

simskij:
That's awesome! No clue how the parenthesis made it in. 😅

Thank you for your contribution!

PR #114, opened 2020-09-23
[chore] deploy a version of docs when PR is created

allansson:
This PR sets up GitHub Actions to deploy a version of the docs everytime a PR is created.

A nice comment will be posted with the URL where the docs can be view, as seen 👇.

### DIFF
--- a/.github/workflows/pr-deploy.yml
+++ b/.github/workflows/pr-deploy.yml
@@ -1,9 +1,9 @@
 name: Pull Request CI
 
 env:
-  PATH_PREFIX: /docs/pull-requests/${{ github.ref }}
+  PATH_PREFIX: /docs/${{ github.ref }}
   GATSBY_DEFAULT_MAIN_URL: https://staging.k6.io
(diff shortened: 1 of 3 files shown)


simskij (changes requested):
All in all, the code looks good! However, the link posted is dead. We should likely tend to that before merging 😅

allansson:
> All in all, the code looks good! However, the link posted is dead. We should likely tend to that before merging 😅

So it turns out that I can't pin gatsby-plugin-s3 to the commit with the necessary "retainObjectsPattern"-feature because they build the plugin on pre-publish. We'll have to wait for the next release of the plugin before merging this PR. Will keep an eye on the package, and fix it whenever they decide to release.

allansson:
@simskij PR:s are now deployed to https://mdr-ci.staging.k6.io/docs. Try out the link above and you should see the deployed version. I think all the links are working as expected, but I might have missed some. External links to e.g. the app or the k6 homepage should go to the staging versions.

Now I don't want to play devops for a year or two. 😅

simskij (approved):

PR #121, opened 2020-09-29
WIP: order of precedence, config options and backlinks

mostafa:
1. Revised order of precedence paragraph.
2. Added backlink from `Config` to `config.json` example on the same page.
3. Added placeholder for order of precedence flowchart, that is going to be drawn by @majavojnoska.
4. Added default config path for different OSes.


MStoykov (approved):

simskij (approved):
I'm not sure whether a flowchart would add any additional value here. In my opinion, these changes are good enough on their own 👍

PR #122, opened 2020-09-30
Follow doc convention for `insecureSkipTLSVerify` option

ppcano:
@MStoykov This change corrects the `insecureSkipTLSVerify` option instruction to follow the format. 

Double checking here - in case I am not aware of something.

### DIFF
--- a/src/data/markdown/docs/01 guides/02 Using k6/05 Options.md
+++ b/src/data/markdown/docs/01 guides/02 Using k6/05 Options.md
@@ -7,52 +7,52 @@ Options allow you to configure how k6 will behave during test execution.
 
 ## List of Options
 
-| Option                                                    | Description                                                                         |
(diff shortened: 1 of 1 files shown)


MStoykov (approved):

src/data/markdown/docs/01 guides/02 Using k6/05 Options.md
```diff
@@ -405,6 +405,26 @@ $ k6 run --include-system-env-vars ~/script.js
 
 </div>
 
+### Insecure Skip TLS Verify
+
+A boolean, true or false. When this option is enabled (set to true), all of the verifications that
+would otherwise be done to establish trust in a server provided TLS certificate will be ignored.
+Available in `k6 run` and `k6 cloud` commands
```
MStoykov, line 412: Maybe an additional comment that this is only about requests done by the script, it doesn't apply for example for requests of outputs (some them have a different option)

PR #132, opened 2020-10-09
Docs: mentions node version number supported

nicosommi:
If the correct version is not set, the process will silently fail unless the user has explicitly set `engine-strict` on their local `npm` configuration.
The errors are confusing, like `flatMap of undefined` or similar (related fix https://github.com/loadimpact/har-to-k6/pull/43/files)


ppcano:
@nicosommi, thank you for your contribution!

PR #137, opened 2020-10-16
Breaking change: mdx processing update

BolotskyDev:
**Describe the changes**
This PR brings a long-awaited update to the way we process mdx: `Code` and `CodeGroup` now being rendered via native mdx renderer as the opposite to previously hot replacement which caused performance issues and was blocking some features to come out.

**Verbose**
- migrated from `gatsby-remark-prismjs` to `prismjs-react-renderer` for max flexibility during custom code block render
- refactored `Code` and `CodeGroup` components
- added postponed earlier collapse/expand mechanism for `Code` component ([task in Notion](https://www.notion.so/pixelpoint/Add-scrolling-option-to-code-blocks-b9f7e7d073e04e8ea5e72152ad9160cd))
- added `CodeInline` component
- updated code blocks usage in every `.md` file

**Side**
- Fixed `Blockquote` component styling bugs described [here](https://github.com/loadimpact/k6-docs/commit/b657e01b7ca3980337571682c839f15a35814008#commitcomment-43306583)
- Since `Cloud API` section has been published, I've enabled Algolia indexing for them back 

**Steps to test**
1. Skim check the diff in `md` files, make sure nothing important got deleted accidentally
2. Make sure it works locally both in dev and prod 
3. Make some random page with code blocks checks, see everything is working as expected
4. Find any page with large code blocks, test expand/collapse mechanism (or just copy past content of a small one temporarily)

PR #138, opened 2020-10-20
Breaking change: mdx processing update

BolotskyDev:
**Describe the changes**
This PR brings a long-awaited update to the way we process mdx: `Code` and `CodeGroup` now being rendered via native mdx renderer as the opposite to previously hot replacement which caused performance issues and was blocking some features to come out.

**Verbose**
- migrated from `gatsby-remark-prismjs` to `prismjs-react-renderer` for max flexibility during custom code block render
- refactored `Code` and `CodeGroup` components
- added postponed earlier collapse/expand mechanism for `Code` component ([task in Notion](https://www.notion.so/pixelpoint/Add-scrolling-option-to-code-blocks-b9f7e7d073e04e8ea5e72152ad9160cd))
- added `CodeInline` component
- updated code blocks usage in every `.md` file

**Side**
- Fixed `Blockquote` component styling bugs described [here](https://github.com/loadimpact/k6-docs/commit/b657e01b7ca3980337571682c839f15a35814008#commitcomment-43306583)

**Steps to test**
1. Skim check the diff in `md` files, make sure nothing important got deleted accidentally
2. Make sure it works locally both in dev and prod 
3. Make some random page with code blocks checks, see everything is working as expected
4. Find any page with large code blocks, test expand/collapse mechanism (or just copy past content of a small one temporarily)

**References**
Recreation from #137


BolotskyDev:
@ppcano This time all conflicts are resolved for merging and rebasing 🙂 Code blocks expand/collapse behavior is reversed as we discussed earlier 👍

ppcano:
@BolotskyDev One issue & One discussion and I think we can merge it to `master`.

- Block code with `javaScript` or `JavaScript` formats are not styled.



Screen Shot 2020-10-20 at 1 02 52 PM



- Block code omitting the title does not show the top bar.

Screen Shot 2020-10-20 at 1 03 46 PM
Screen Shot 2020-10-20 at 1 03 33 PM

I like more the style with the empty bar but this is a design issue. Up to them.

BolotskyDev:
@ppcano 

1) Yeah, this new processing library is kinda stricter that previous, so it requires language declaration to be exact match to one of the [following list](https://github.com/FormidableLabs/prism-react-renderer/blob/master/src/vendor/prism/includeLangs.js). I think we can go around the problem implementing a) language coercion mechanism that will handle case errors and ambiguous declarations like `sh`, `shell`, `bash` b) client side console notifications if the language ain't supported at all (has no match or alternative in the list mentioned above)

2) Not a problem, but let's clarify something: do you want _all_ code blocks to get a terminal header, even those not wrapped in `CodeBlock` or those which were wrapped, but `labels` prop is omitted?

ppcano:
@BolotskyDev 

1 - We can update all the examples if that is your preferred option. I guess we will have to do the same on the blog later.

2 - Ah!, I didn't remember there were 3 options: No labels, empty labels, and labels. Let's continue with the same behavior as before.

BolotskyDev:
@ppcano 
1) I wrapped my head around the problem a little and here is what I've come with: I do not want to rely on code processing library and bring tons of changes every time list of supported languages or the library itself will change. My suggestion is to define a strict list of supported languages in contributing docs for authors to use based on current usage that will be always supported no matter how we might decide process code in a future.

For example, we define [`javascript`, `bash`, `html`, `json`] with explicit contract that those will be supported no matter what happens under the hood, and everything else will be rendered as a plain text (no highlight). This list could grow in size any moment we feel necessity to support highlight another language.

That would increase consistency tremendously, since at the moment there is a clear ambiguity in usage that can confuse any content writer:
- for terminal code there are `sh`, `shell`, `bash`, `powershell` declarations
- for javascript there are `javascript`, `js`, `JS`, `JavaScript` declarations

Also there is a number of other languages, like `sql`, `xml`, `python` which is declared, but never highlighted. Great opportunity to solve highlighting problems once and for all.

Wdyt? 

2) Done, in the last commit I made sure behavior is preserved: non-wrapped code blocks will be headerless, and `CodeGroup`-ed code blocks will have a header even if no `labels` prop is provided

ppcano:
1 - Agree. Let me know how to proceed.

BolotskyDev:
@ppcano Okay, I believe that's it!

Final list of languages includes: `bash`, `javascript`, `json`, `markup` (pretty much html and xml), `diff`, `go`, `plain`. Md files were updated accordingly. Contribution docs now have `Language` [subsection](https://github.com/loadimpact/k6-docs/blob/refactor/mdx-processing-update/CONTRIBUTING_FILE_FORMAT.md#languages).

PR #146, opened 2020-11-03
Removed unnecessary '$' to allow copying and pasting

joshuaj004:
When you copy the command, it copies the '$' as well which causes an error when you paste it into your terminal. If you look at the 'Adding more VUs' section, these commands omit the '$' so I believe this is the correct behavior. 

Btw I love using k6!

### DIFF
--- a/src/data/markdown/docs/01 guides/01 Getting started/03 Running k6.md
+++ b/src/data/markdown/docs/01 guides/01 Getting started/03 Running k6.md
@@ -27,7 +27,7 @@ Then run k6 using this command:
 <CodeGroup labels={["CLI", "Docker", "Docker in Win PowerShell"]}>
 
 ```bash
-$ k6 run script.js
+k6 run script.js
 ```
 
 ```bash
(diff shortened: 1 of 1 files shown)


joshuaj004:
I didn't modify anything relating to the deployment process, so I think you should be able to re-run it and have it work. I'm not sure how to do this, thoughts?

ppcano:
@joshuaj004 Thank you for pointing out and the PR.

@simskij This is a common issue in the Docs. 


Screen Shot 2020-11-09 at 4 15 16 PM


If we want to prepend the symbol - which I am not sure - we should do it in CSS and skip it when copying.

BolotskyDev:
@ppcano FYI, I can implement that fairly easy, but some of our `bash` code blocks contain `powershell` code and I don't think I'd be able to distinguish it programmatically with current setup. That means, some powershell code will be prepended with `$` as well, which could be quite confusing. 

Solution in current setup: replace `bash` with `plain` in powershell code blocks.
Solution with refactoring: well, add `powershell` support.

ppcano:
Thank @BolotskyDev for informing us. 

@simskij was discussing this issue internally and will inform the decision.

simskij:
Hi Josh,

Thank you for the suggestion! We've decided to keep the dollar signs in the scripts for now. However, we would like to proceed with @BolotskyDev's idea about omitting them when copying. I'm creating a separate issue for that.

Closing this PR without merging. With that said, I'd like to thank you again for the excellent contribution. It definitely got us thinking through the UX.

Best,
Simme

PR #161, opened 2020-11-25
feat: bash input signs are no more parts of copy btn content

### LINKED ISSUE #158: Strip prompts from copied shell scripts.
> Hi Josh,
> 
> Thank you for the suggestion! We've decided to keep the dollar signs in the scripts for now. However, we would like to proceed with @BolotskyDev's idea about omitting them when copying. I'm creating a separate issue for that.
> 
> Closing this PR without merging. With that said, I'd like to thank you again for the excellent contribution. It definitely got us thinking through the UX.
> 
> Best,
> Simme

_Originally posted by @simskij in https://github.com/loadimpact/k6-docs/issues/146#issuecomment-730471676_

I think you are most suited to decide the best path forward, @BolotskyDev, so I'm happy to leave this in your hands.

BolotskyDev:
**Describe changes**
This commit makes sure terminal input signs are no more parts of copy btn content in code blocks.

<details>

<summary>Verbose</summary>

Unfortunately, I failed to set up attaching `$` signs automatically as planned initially for two reasons:

1. First try: attach to every single line of a bash code block

Quickly realised it was a bad idea since bash blocks may include not only bash commands, but output as well and `$` signs would be redundant and look weird. 

2. Second try: use the only possible way to distinguish terminal input from terminal output programmatically.

Lines with input starts with `function` token type whereas output tokens have token type of `plain`. I set up to attach `$` sign to a beginning of a code line if it starts with `function`, but It turned out this isn't an option as well: prism parser doesn't recognize such commands as `brew`, `docker`, `k6` etc, so lines starting with those wouldn't have `$` sign and we have no capabilities to extend the dictionary.

Given that, I came up with the following solution: we keep `$` signs in bash terminals, add them manually where we need it, but copy button would be stripped of every `$` that is a first character of a code line.

</details>

**Side**:
- adjusted `CodeGroup` component similarly how it was done on blog to prevent buggy behavior
- added `$` signs to bash blocks on Installation page

**Steps to test**
1. [Open up preview](https://mdr-ci.staging.k6.io/docs/refs/pull/161/merge), find any bash terminal with `$` signs indicating the bash input
2. Click on `Copy` button, paste the buffer content
3. Assure `$` chars were stripped from lines indeed

**References**
Closes #158

### DIFF
--- a/src/components/shared/code/code-group.js
+++ b/src/components/shared/code/code-group.js
@@ -1,13 +1,11 @@
 import PropTypes from 'prop-types';
 import React, { useState } from 'react';
-import { getRandomKey } from 'utils';
 
 import Code from './code';
 import styles from './code-group.module.scss';
 
 const CodeGroup = ({ children, labels, lineNumbers, heightTogglers }) => {
   const [currentIndex, setCurrentIndex] = useState(0);
-  const randomKey = getRandomKey();
   return (
     <div className={styles.wrapper}>
       <div className={styles.header}>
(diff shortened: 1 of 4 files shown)


simskij:
Ace! Good job!

simskij (approved):

PR #178, opened 2020-12-24
Update 01 Overview.md

mjraadi:
changed "write" to "right" because "write" didn't make sense in the context


ppcano:
@mjraadi, thanks for your contribution!

PR #196, opened 2021-01-22
SharedArray documentation

MStoykov:


src/data/markdown/docs/02 javascript api/10 k6-data.md
```diff
@@ -0,0 +1,10 @@
+---
+title: "k6/data"
+excerpt: "k6 data API"
+---
+
+The data module provides a helpers to work with data.
```
imiric, line 6:
```suggestion
The data module provides helpers to work with data.
```

src/data/markdown/docs/02 javascript api/10 k6-data/10-SharedArray.md
```diff
@@ -0,0 +1,44 @@
+---
+title: SharedArray
+---
+
+`SharedArray` is an array-like object that shares the underlying memory between VUs. It constructor
```
imiric, line 5:
```suggestion
`SharedArray` is an array-like object that shares the underlying memory between VUs. Its constructor
```

src/data/markdown/docs/02 javascript api/10 k6-data/10-SharedArray.md
```diff
@@ -0,0 +1,44 @@
...
+takes a name for the `SharedArray` and a function which needs to return an array object itself. The
+function will be executed only once and it's result will be then be saved in memory once and copies
```
imiric, line 7:
```suggestion
function will be executed only once and its result will then be saved in memory once and copies
```

src/data/markdown/docs/02 javascript api/10 k6-data/10-SharedArray.md
```diff
@@ -0,0 +1,44 @@
...
+of the elements will be given when requested. The name is needed as VUs are completely separate js
```
imiric, line 8:
```suggestion
of the elements will be given when requested. The name is needed as VUs are completely separate JS
```

src/data/markdown/docs/02 javascript api/10 k6-data/10-SharedArray.md
```diff
@@ -0,0 +1,44 @@
...
+
+This does mean that you can have multiple such ones and even only load some of them for given VUs, although that is
+unlikely to have any performance benefit.
+
+Everything about `SharedArray` is read-only once it is constructed, so it is not possible to
+communicate between VUs using it.
+
+Supported operations include:
+1. getting the number of elements with `length`
+2. getting an element by it's index using the normal syntax `array[index]`
```
imiric, line 19:
```suggestion
2. getting an element by its index using the normal syntax `array[index]`
```

src/data/markdown/docs/02 javascript api/10 k6-data/10-SharedArray.md
```diff
@@ -0,0 +1,44 @@
...
+3. using `for-of` loops
```
imiric, line 20: Let's mention the performance penalty of using `for-of` loops and that a standard C-like loop is preferred.

src/data/markdown/docs/02 javascript api/10 k6-data/10-SharedArray.md
```diff
@@ -0,0 +1,44 @@
...
+
+var data = new SharedArray("some name", function() {
+    // here you can open files, and then do additional processing on them or just generate the data dynamically 
+    var f = JSON.parse(open("./somefile.json")).;
+    return f; // f must be an array
+});
+
+export default () => {
+    var element = data[Math.floor(Math.random() * data.length)]
+    // do something with element
```
imiric, line 40: Let's include a more complex example showing the common use case of splitting the array among N VUs.

src/data/markdown/docs/02 javascript api/10 k6-data/10-SharedArray.md
MStoykov: I don't think the performance penalty is interesting the moment you do anything with that data. And I am pretty sure the majority of the penalty comes from the copying of the data, not the actual semantics

That will also change, given that I will be optimizing it using the recently merged changes in goja, on the next update. And also that the goja author might implement something like what we did directly in goja and then we will move to it - either way.

src/data/markdown/docs/02 javascript api/10 k6-data/10-SharedArray.md
MStoykov: Do you mean the csv example from https://k6.io/docs/examples/data-parameterization#handling-bigger-data-files (which I need to update as well) ?  Because that was a hack that is no longer needed, which the whole point was to only get parts of the array in each VU so that you don't load most of it. There is no need for that, now. Arguably having them split will be *worse* not better, so I will need to update the linked page, to specifically recommend not doing that.

imiric: Sure, that specific example is not needed anymore, and we should just remove it in favor of this. But most people will want to get a subset of the array anyway and not a random element as shown in this example, in which case `__VU` would still be relevant. So showing that particular real-world scenario would be useful.

src/data/markdown/docs/02 javascript api/10 k6-data/10-SharedArray.md
imiric: OK, but _right now_ there's a performance penalty we should mention, so that people who need it use the plain loop instead.

MStoykov: I am still of the opinion that while you might be (look below for why you are not) technically correct - there just isn't a practical reason to not use it, as the performance difference will disappear the moment you use the data for anything. 

I also think you are confusing the difference in the for-of if we don't use SharedArray at all -where there is an extremely big performance difference, but again, that IMO has no practical applications, as you still need to do something with the data.. 

I did quick not very scientific benchmark where the test file is:
```
{"user": "myuser", "password": "password12345"},
```
100k times I ran with 1 VU 10 iterations... and:L

```javascript
    var len  = s.length
    for (var i = 0; i < len; i++) { s[i] }
```
did:
```
     iteration_duration...: avg=2.75s min=2.51s med=2.67s max=3.25s p(90)=3.2s p(95)=3.23s
     iterations...........: 10  0.362695/s
```


```javascript
    for (var i = 0; i < s.length; i++) { s[i] }
```
did:
```
     iteration_duration...: avg=3.29s min=3.2s med=3.29s max=3.38s p(90)=3.35s p(95)=3.36s
     iterations...........: 10  0.303433/s
```

```javascript
    for (var v of s) {}
```
did:
```
     iteration_duration...: avg=2.28s min=1.96s med=2.14s max=2.8s p(90)=2.69s p(95)=2.74s
     iterations...........: 10  0.437979/s
```
So the for-of one is faster (barely ... possibly because `s[i]` taking the array element needs to go through more ... checks. But again IMO the whole thing doesn't matter because this is iterating over 100k elements and doing nothing but just getting them which is not useful. So accessing 1 element takes 28ms, which while not great IMO, is nothing compared to how long it will take to make an HTTP request. 
And again this is not the for-of loop ... this is just the copying of the element (and all the other things that need to be done in order to get the element).

MStoykov: I did the for-of loop directly on the underlying array, not on the SharedArray and got
```
     iteration_duration...: avg=106.94ms min=78.2ms med=93.9ms max=161.07ms p(90)=149.06ms p(95)=155.07ms
     iterations...........: 10  4.745281/s
```
so it's more than 10x faster ... again not really useful given that you will need to do something with that data.

MStoykov: I guess we can add a note in the vein of "accessing elements has a bigger cost  than with a normal array, but that will balance out in favor of SharedArray for arrays of considerable size, and is likely to be unnoticeable for smaller arrays as well"

imiric: Hhmm then I was confused by [this comment](https://github.com/dop251/goja/issues/232#issuecomment-745210887) and the discussion in #1739, as my impression was that the wrapping done to implement the iterator would be much less performant, but now I'm still confused as to how it can be faster...

Anyway, if you think none of this should be mentioned I'm OK with it.

MStoykov: see https://github.com/loadimpact/k6-docs/pull/196/commits/f90b0b3553138290abd63aaf9f98645f07ad5ee8

src/data/markdown/docs/02 javascript api/10 k6-data/10-SharedArray.md
MStoykov: I have never seen people wanting a subset of the array per iteration. They might want 1-2 elements from it, but I at least have not seen someone asking for a consecutive subset of the array. 

Also, this is not specific to the SharedArray, so I am kind of against having 1 more example.

ppcano (approved):

src/data/markdown/docs/02 javascript api/10 k6-data/10-SharedArray.md
```diff
@@ -0,0 +1,107 @@
...
+| 1000       | true   | 2:03:00   | 86-92%   | 509-519MB   | 92007-95234   |
+| 1000       | false  | 2:02:60   | 78-80%   | 621-630MB   | 92814-94526   |
+| 10000      | true   | 2:02:70   | 88-95%   | 515-523MB   | 92936-94997   |
+| 10000      | false  | 2:03:80   | 81-85%   | 1650-1675MB | 92339-95083   |
+| 100000     | true   | 2:04:50   | 89-91%   | 528-531MB   | 92274-93987   |
+| 100000     | false  | 2:15:00   | 115-123% | 8.9-9.5GB   | 90416-94817   |
+
+The CPU/memory data comes from using `/usr/bin/time` and the raw data can be find [here](https://gist.github.com/MStoykov/1181cfa6f00bc56b90915155f885e2bb)
+
+This numbers are purely illustrative as if the script is doing more with an element that it gets from the `SharedArray`, or an output is in use, or it gets multiple elements, or - all of those things add up so, while `SharedArray` has some CPU usage it might turn out that it will be negligible in a given situation with just 10 elements or more problematic then the memory usage for a 100k elements. So if in doubt you should probably run some benchmarks.
```
imiric, line 107:
```suggestion
These numbers are purely illustrative as the performance can be affected by any additional processing of the element retrieved from the `SharedArray`, or if an output is in use, or it gets multiple elements, etc. While `SharedArray` has some CPU usage it might turn out that it will be negligible in a given situation with just 10 elements or more problematic than the memory usage for a 100k elements. So if in doubt you should probably run some benchmarks and decide which tradeoffs are more important for your use case.
```

src/data/markdown/docs/05 Examples/01 Examples/05 data-parameterization.md
```diff
@@ -16,14 +16,21 @@ This will, in turn, make your test more realistic.
...
-Cloud service allots 8GB of memory for every 300VUs.
-When executing cloud tests, make sure your data files aren't exceeding this limit or your test run may get aborted.
-
-Starting with k6 v0.27.0, there are some [tricks that can be used to better handle bigger data files](#handling-bigger-data-files).
+Each VU in k6 is a separate JS VM so in order to not have multiple copies of the whole data file
+[SharedArray](/javascript-api/k6-data/sharedarray) was added. It does have overhead in accessing elements compared to a normal non shared
+array, but the difference is negligible compared to the time it takes to make requests. And is
+especially not comparable as not using it with big files basically means k6 will be using too much
+memory to run. Which might lead to your script not being able to run at all or aborting in the
+middle.
```
imiric, line 24:
```suggestion
[SharedArray](/javascript-api/k6-data/sharedarray) was added. It does have some CPU overhead in accessing elements compared to a normal non shared
array, but the difference is negligible compared to the time it takes to make requests. This becomes
even less of an issue compared to not using it with large files, as k6 would otherwise use too much memory to run, which might lead to your script not being able to run at all or aborting in the middle if the system resources are exhausted.
```

src/data/markdown/docs/05 Examples/01 Examples/05 data-parameterization.md
```diff
@@ -16,14 +16,21 @@ This will, in turn, make your test more realistic.
...
+
+For example, the Cloud service allots 8GB of memory for every 300VUs. So if you are files are big
+enough and you are not using [SharedArray](/javascript-api/k6-data/sharedarray), that might mean that your script will run out of memory at
+some point. Additionally even if there is enough memory, k6 has a garbage collector(as it's written
+in golang) and it will walk through all accessible objects (including JS ones) and figure out which
+need to be garbage collected. For big JS arrays copied hundreds of times this adds quite a lot of
+additional work.
```
imiric, line 31: The second sentence just repeats what the last sentence of the previous paragraph said, and mentioning the garbage collector seems like too much detail for this documentation.

Mentioning the Cloud limits might be relevant, but we'll have to keep it updated if that changes, and I'm not sure if it's worth bringing it up at all or if this is the right place for it. So I would scrap this paragraph entirely, but if not some minor suggestions:

```suggestion
For example, the Cloud service allocates 8GB of memory for every 300 VUs. So if your files are large
enough and you are not using [SharedArray](/javascript-api/k6-data/sharedarray), that might mean that your script will run out of memory at
some point. Additionally even if there is enough memory, k6 has a garbage collector (as it's written
in golang) and it will walk through all accessible objects (including JS ones) and figure out which
need to be garbage collected. For big JS arrays copied hundreds of times this adds quite a lot of
additional work.
```

src/data/markdown/docs/05 Examples/01 Examples/05 data-parameterization.md
```diff
@@ -16,14 +16,21 @@ This will, in turn, make your test more realistic.
...
+
+A note on performance characteristics of `SharedArray` can be found within it's [API documentation](/javascript-api/k6/data/sharedarray#performance-characteristics).
```
imiric, line 33:
```suggestion
A note on performance characteristics of `SharedArray` can be found within its [API documentation](/javascript-api/k6/data/sharedarray#performance-characteristics).
```

src/data/markdown/docs/05 Examples/01 Examples/05 data-parameterization.md
```diff
@@ -45,7 +52,11 @@ Starting with k6 v0.27.0, there are some [tricks that can be used to better hand
 <CodeGroup labels={["parse-json.js"]} lineNumbers={[true]}>
 
 ```javascript
-const data = JSON.parse(open('./data.json'));
+import { SharedArray } from "k6/data";
+// not using ShareArray here will mean that the code in the function call (that is what loads and
```
imiric, line 56:
```suggestion
// not using SharedArray here will mean that the code in the function call (that is what loads and
```

src/data/markdown/docs/05 Examples/01 Examples/05 data-parameterization.md
```diff
@@ -66,8 +77,14 @@ You can download the library and import it locally like this:
 
 ```javascript
 import papaparse from './papaparse.js';
-
-const csvData = papaparse.parse(open('./data.csv'), { header: true });
+import { SharedArray } from "k6/data";
+// not using ShareArray here will mean that the code in the function call (that is what loads and
```
imiric, line 81:
```suggestion
// not using SharedArray here will mean that the code in the function call (that is what loads and
```

src/data/markdown/docs/05 Examples/01 Examples/05 data-parameterization.md
```diff
@@ -82,9 +99,15 @@ Or you can grab it directly from [jslib.k6.io](https://jslib.k6.io/) like this.
 
 ```javascript
 import papaparse from 'https://jslib.k6.io/papaparse/5.1.1/index.js';
+import { SharedArray } from "k6/data";
 
-// Load CSV file and parse it using Papa Parse
-const csvData = papaparse.parse(open('./data.csv'), { header: true });
+// not using ShareArray here will mean that the code in the function call (that is what loads and
```
imiric, line 104:
```suggestion
// not using SharedArray here will mean that the code in the function call (that is what loads and
```

src/data/markdown/docs/05 Examples/01 Examples/05 data-parameterization.md
```diff
@@ -100,26 +123,31 @@ data to login to the test.k6.io test site:
...
+test_user,1234
 */
 import http from 'k6/http';
 import { check, sleep } from 'k6';
+import { SharedArray } from 'k6/data'
 import papaparse from 'https://jslib.k6.io/papaparse/5.1.1/index.js';
 
-// Load CSV file and parse it using Papa Parse
-const csvData = papaparse.parse(open('./data.csv'), { header: true }).data;
+// not using ShareArray here will mean that the code in the function call (that is what loads and
```
imiric, line 135:
```suggestion
// not using SharedArray here will mean that the code in the function call (that is what loads and
```

src/data/markdown/docs/05 Examples/01 Examples/05 data-parameterization.md
```diff
@@ -144,12 +173,22 @@ export default function () {
...
-## Handling bigger data files
+## Generating data
 
-In k6 version v0.27.0, while there is still no way
-to share memory between VUs, the `__VU` variable is now defined during the init
-context which means that we can split the data between the VUs during initialization
-and not have multiple copies of it during the test run.
+See [this example project on GitHub](https://github.com/k6io/example-data-generation) showing how to use faker.js to generate realistic data at runtime.
+
+## Old hacks
```
imiric, line 180: :)

```suggestion
## Old workarounds
```

src/data/markdown/docs/02 javascript api/10 k6-data/10-SharedArray.md
```diff
@@ -0,0 +1,107 @@
...
+
+</div>
+
+## Performance characteristics
+
+As the `SharedArray` is keeping the data marshalled as JSON and unmarshal elements when requested it
+does:
+
+1. take additional time to unmarshal JSONs
+2. generate objects that then need to be garbage collected
```
imiric, line 52: This doesn't really require a list, so it would be more legible as:

```suggestion
As the `SharedArray` is keeping the data marshalled as JSON and unmarshals elements when requested it
does take additional time to unmarshal JSONs and generate objects that then need to be garbage collected.
```

imiric (commented):
Mostly LGTM, save for some minor grammar issues.

Also, thanks for the benchmark. It does give a sense of the memory savings that can be expected.

src/data/markdown/docs/05 Examples/01 Examples/05 data-parameterization.md
MStoykov: The cloud limits were here from before. I am not certain this is the correct place for them as well, but it is definetely good to have them somewhere. Given that if a cloud user runs out of memory they might be directed to this page, I am for this actually staying. It also shows how much memory we have decided to allocate to a given number of VUs.

As to change ... things changes, so we will need to update it if we ever  change it in the cloud :)

imiric (approved):

na-- (approved):

PR #197, opened 2021-01-22
Misc k6 v0.30.0 documentation

imiric:
This should cover all changes in k6 v0.30.0, besides the `SharedArray` and end-of-test summary, which are in progress.

Let me know if I missed anything.


src/data/markdown/docs/02 javascript api/06 k6-http/10-post- url- -body-- -params- -.md
```diff
@@ -22,18 +22,29 @@ description: 'Issue an HTTP POST request.'
...
+  res = http.post(url, data);
   console.log(JSON.parse(res.body).form.name);
+
+  // Using an ArrayBuffer as body, allowing more flexibility when
+  // building requests with binary data.
+  // Make sure to open() the file as binary (with the 'b' argument),
+  // then wrap the raw binary array in a TypedArray view, and pass the
+  // underlying ArrayBuffer to http.post().
+  data = new Uint8Array(logoBin);
+  http.post(url, data.buffer, { headers: { 'Content-Type': 'image/png' }});
```
MStoykov, line 47: Not certain that this is a good example. If you just want to send the file - no need to use ArrayBuffer ;).

So I propose we add 2 examples:
1. just use open and send that
2. build something in ArrayBuffer and send it, also I would argue it should be noted that you can't just send the TypedArray, but only the underlying ArrayBuffer, so people don't confuse it. (also possibly important to note that the underlying ArrayBuffer might be bigger than what is used by the TypedArray ... which also puts a wrench in a simple complete example, but let's go with simple and add more later ?)

MStoykov (commented):
Seem good to me, but I would prefer to not tell users to wrap []byte in ArrayBuffer, just to do it.
Also, not really certain it's a great idea to have 2 identical examples using a slightly different syntax (the request and post examples) :thinking:

src/data/markdown/docs/02 javascript api/06 k6-http/10-post- url- -body-- -params- -.md
imiric: Agreed that we need a more complex example, but building an ArrayBuffer manually is not a real-world scenario many people will find useful, so not sure it's a good fit here.

I would've liked to include the `multipart/form-data` example, but that would require us stabilizing the polyfill I was testing and maybe make it part of the jslib site. Then it should probably live on a separate docs page as there is a lot of interest for that use case.

WYT?

MStoykov: My argument is more in the vein that showing the wrapping of `open()` is really not all that good of an example, because it's pointless. So having even a not so useful example of how to generate 10 bytes of binary data is a better example in my opinion. 

Showing bad examples that shouldn't be copied is just inviting someone to just do that - copy them ;).

And yes, as we've agreed in other places the FormData polyfill is important and was literally the idea behind this being pushed as it is blocking a lot of people, so :100: for it having it's own documentation page, which is also linked and recommended here instead of using the "magical" object->form-data k6 transformation. Obviously with some note on possible performance differences.

imiric: Resolved by dfcd9cd.

While I agree that wrapping the value of `open()` in a TypedArray is not a useful example, it's at least somewhat practical, whereas I don't foresee anyone needing to build a binary array manually as shown in dfcd9cd, but :man_shrugging:. I think it's fine either way.

imiric:
This should be ready now.

@MStoykov I removed the binary/ArrayBuffer examples from `http.request()`, so they're not the same anymore, but these examples were identical for `http.post()` even before this change. I don't mind some duplication for the `http.request()` examples, since the other methods are just wrappers for it.

MStoykov (approved):

na-- (approved):
LGTM

PR #200, opened 2021-01-26
Add randomIntBetween example

tom-miseur:
randomIntBetween(min,max) is useful!


ppcano:
@tom-miseur And `randomItem`?

@simskij Let us discuss what approach to follow for documenting `jslib` more extensively.

simskij (approved):

PR #206, opened 2021-02-01
Advanced multipart request example + minor fixes

### LINKED ISSUE #203: Add FormData example
After https://github.com/loadimpact/k6/pull/1800 and https://github.com/loadimpact/jslib.k6.io/pull/21 are merged (i.e. k6 v0.31.0), we should add an example that demonstrates how to build advanced `multipart/form-data` request bodies with them. As evidenced by https://github.com/loadimpact/k6/issues/747, https://github.com/loadimpact/k6/issues/843, https://github.com/loadimpact/k6/issues/1375, https://github.com/loadimpact/k6/issues/1382, https://github.com/loadimpact/k6/issues/1571, https://github.com/loadimpact/k6/issues/1784, the current k6 HTTP API is quite restricted and can't cover a lot of corner cases... :disappointed: We'll close these issues (https://github.com/loadimpact/k6/pull/1800#issuecomment-768157584), so we probably should also add warning messages in https://k6.io/docs/javascript-api/k6-http/file-data-filename-contenttype and https://k6.io/docs/javascript-api/k6-http/filedata to use the JS polyfill for the corner cases the Go API can't cover.

At some future point, we'll try to fix all of these deficiencies with a new and better k6 JS API for HTTP requests, but until then this polyfill will likely be sufficient for 99% of users, with almost no downsides. So, we should document it properly.

imiric:
Closes #203

We should merge https://github.com/loadimpact/jslib.k6.io/pull/21 and deploy first, otherwise the `FormData` links won't work.

### DIFF
--- a/src/data/markdown/docs/05 Examples/01 Examples/09 data-uploads.md
+++ b/src/data/markdown/docs/05 Examples/01 Examples/09 data-uploads.md
@@ -86,3 +86,52 @@ and sent.
 
 - [open(filePath, [mode])](/javascript-api/init-context/open-filepath-mode)
 - [http.file(data, [filename], [contentType])](/javascript-api/k6-http/file-data-filename-contenttype)
+
+
+## Advanced multipart request
+
+The previous multipart request example has some limitations:
+
+- It's not possible to assemble the parts in a specific order, because of the
+  unordered nature of JS objects when they're converted to Golang maps, which k6 uses internally.
+  Uploading files in a specific order is a requirement for some APIs (e.g. AWS S3).
+- It's not possible to upload multiple files as part of the same form field, because
+  JS object keys must be unique.
+
+To address this we suggest using the [`FormData` polyfill for k6](https://jslib.k6.io/formdata/0.0.1/index.js).
+
+Here's an example of uploading several binary files and a text file using the polyfill:
+
+<CodeGroup labels={["Advanced multipart request example"]} lineNumbers={[true]}>
+
+```javascript
+import http from 'k6/http';
+import { check } from 'k6';
+import { FormData } from 'https://jslib.k6.io/formdata/0.0.1/index.js';
+
+const img1 = open('/path/to/image1.png', 'b');
+const img2 = open('/path/to/image2.jpg', 'b');
+const txt = open('/path/to/text.txt');
+
+export default function() {
+  const fd = new FormData();
+  fd.append('images', http.file(img1, 'image1.png', 'image/png'));
+  fd.append('images', http.file(img2, 'image2.jpg', 'image/jpeg'));
+  fd.append('text', http.file(txt, 'text.txt', 'text/plain'));
+
+  const res = http.post('https://httpbin.test.k6.io/post', fd.body(),
+    { headers: { 'Content-Type': 'multipart/form-data; boundary=' + fd.boundary }});
+  check(res, {
+    'is status 200': (r) => r.status === 200,
+  });
+}
+```
+
+</CodeGroup>
+
+Note that:
+
+- Both binary files will be uploaded under the `images` form field, and the text file
+  will appear last in the request under the `text` form field.
+- It's required to specify the multipart boundary in the `Content-Type` header,
+  so you must assemble the header manually as shown.
--- a/src/data/markdown/docs/05 Examples/01 Examples/09 data-uploads.md
+++ b/src/data/markdown/docs/05 Examples/01 Examples/09 data-uploads.md
@@ -1,9 +1,9 @@
 ---
 title: 'Data Uploads'
-excerpt: 'Scripting examples on how to execute a load test that will upload a file to the System Under Test(SUT).'
+excerpt: 'Scripting examples on how to execute a load test that will upload a file to the System Under Test (SUT).'
 ---
 
-Example to execute a load test that will upload a file to the System Under Test(SUT).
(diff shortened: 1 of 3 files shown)


src/data/markdown/docs/05 Examples/01 Examples/09 data-uploads.md
```diff
@@ -86,3 +86,52 @@ and sent.
...
 - [open(filePath, [mode])](/javascript-api/init-context/open-filepath-mode)
 - [http.file(data, [filename], [contentType])](/javascript-api/k6-http/file-data-filename-contenttype)
+
+
+## Advanced multipart request
+
+The previous multipart request example has some limitations:
+
+- It's not possible to assemble the parts in a specific order, because of the
+  unordered nature of JS objects when they're converted to Golang maps, which k6 uses internally.
```
MStoykov, line 96: nit: This makes it sound like JS objects don't have an order to their keys ... but they do. I am pretty sure that even in goja that works. The problem is that they are converted to golang maps, which don't, but I am not certain this is made clear from the wording here. I tried fixing it ... but with no success :(

src/data/markdown/docs/05 Examples/01 Examples/09 data-uploads.md
```diff
@@ -86,3 +86,52 @@ and sent.
...
+import http from 'k6/http';
+import { check } from 'k6';
+import { FormData } from 'https://jslib.k6.io/formdata/0.0.1/index.js';
+
+const img1 = open('/path/to/image1.png', 'b');
+const img2 = open('/path/to/image2.jpg', 'b');
+const txt = open('/path/to/text.txt');
+
+export default function() {
+  let fd = new FormData();
```
MStoykov, line 117: nit: this can also be `const` as `const` is only about the identifier - it prevents the identifier to point/identify another object, after initialization.

MStoykov (approved):
LGTM. 

I do think we should probably add a warning to users that using the Object body has the problems you've outlined everywhere. But:
1. this can be done in another PR
2. it will probably be better done after we have tested the FormData somewhat more :) and have gotten at least some reports that it works great for users as well :)

ppcano (approved):

src/data/markdown/docs/05 Examples/01 Examples/09 data-uploads.md
imiric: AFAIK key order was unspecified in <=ES5, and browsers did their own implementation. Even in >=ES6 [the situation is messy](https://stackoverflow.com/a/30919039/96213), and integer keys are ordered numerically, while others are ordered by insertion.

I'm not sure how goja handles this internally, but for our purposes (and I'd argue in general for JS) it's clearer to not rely on objects for order, and use something like `Map` which explicitly supports it.

I thought the phrasing here is clear that it's an issue when they're converted to Golang maps and didn't want to go into any more detail. So suggestions for how to change it is welcome.

src/data/markdown/docs/05 Examples/01 Examples/09 data-uploads.md
imiric: Thanks, I was aware of that, but wasn't thorough enough with the usage. Will change it.

src/data/markdown/docs/05 Examples/01 Examples/09 data-uploads.md
MStoykov: It just has the property names as a list  ;) 
https://github.com/dop251/goja/blob/f5884268f0c0a0eb94423dcf7c2eda14907375e6/object.go#L191-L199

as I said, this is a nitpick. Unless somebody else comes up with better wording, as I can't myself as well, it is fine to just merge it as is.

src/data/markdown/docs/05 Examples/01 Examples/09 data-uploads.md
imiric: Resolved in 397babd.

src/data/markdown/docs/05 Examples/01 Examples/09 data-uploads.md
```diff
@@ -86,3 +86,52 @@ and sent.
...
+  unordered nature of JS objects when they're converted to Golang maps, which k6 uses internally.
+  Uploading files in a specific order is a requirement for some APIs (e.g. AWS S3).
+- It's not possible to upload multiple files as part of the same form field, because
+  JS object keys must be unique.
+
+To address this we suggest using the [`FormData` polyfill for k6](https://jslib.k6.io/formdata/0.0.1/index.js).
+
+Here's an example of uploading several binary files and a text file using the polyfill:
+
+<div class="code-group" data-props='{ "labels": ["multipart-upload-formdata.js"], "lineNumbers": [true] }'>
```
simskij, line 105: Please use the `CodeGroup` component instead:

```
<CodeGroup labels={["Nice code!", "This one is better", "Oh my.."]} lineNumbers={[true, true, true]}>
   ...
</CodeGroup
```

simskij (changes requested):
One minor detail. After that, feel free to merge without a second review. 👍🏼

src/data/markdown/docs/05 Examples/01 Examples/09 data-uploads.md
imiric: Thanks, I think copied this from an old part of the docs. Will keep it in mind for the future.

PR #212, opened 2021-02-05
handleSummary() and other fixes

na--:
Closes https://github.com/loadimpact/k6-docs/issues/184, closes https://github.com/loadimpact/k6-docs/issues/191, closes https://github.com/loadimpact/k6-docs/issues/30, closes https://github.com/loadimpact/k6-docs/issues/167, closes https://github.com/loadimpact/k6-docs/issues/187


simskij:
Deploybot being deploybot. URL: https://mdr-ci.staging.k6.io/docs/refs/pull/212/merge/

na--:
This should be ready for review - I split it into sane commits and made a few minor improvements.

src/data/markdown/docs/01 guides/04 Results visualization/00 End-of-test Summary.md
```diff
@@ -0,0 +1,199 @@
+---
+title: 'End-of-test Summary'
+excerpt: 'By default, k6 prints a summary report containing the aggregated results at the end of the test. Since k6 v0.30.0, the new `handleSummary() callback allows the report to be completely customized, including the generation and saving of JSON, HTML, XML (e.g. JUnit), etc. reports to files.'
```
MStoykov, line 3:
```suggestion
excerpt: 'By default, k6 prints a summary report containing the aggregated results at the end of the test. Since k6 v0.30.0, the new `handleSummary()` callback allows the report to be completely customized, including the generation and saving of JSON, HTML, XML (e.g. JUnit), etc. reports to files.'
```

src/data/markdown/docs/01 guides/04 Results visualization/00 End-of-test Summary.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Summary export to a JSON file
+
+Since k6 v0.26.0, k6 has had the [`--summary-export=path/to/file.json` option](/using-k6/options#summary-export) for local test runs. It exports some of the summary report data to a JSON file format.
+
+Unfortunately, the exported format is somewhat limited and has a few confusing peculiarities. For example, groups and checks are unordered. Threshold values are also somewhat unintuitive - they signify whether the threshold has been crossed. So, `true` is the "bad" threshold value, i.e. when the threshold has failed, and `false` is the "good" value...
+
+We couldn't change the `--summary-export` data format because it would have broken backwards compatibility in a feature that people depended on in CI, so it still works how it used to. However, in k6 v0.30.0, we introduced `handleSummary()` - a new and better way to make JSON exports of the summary data, as well as any other format (CSV, XML (JUnit/xUnit/etc.), HTML, TXT, etc.) that may be required. We strongly recommend everyone to use `handleSummary()` instead of `--sumary-export`. For more details, see the next section in this document...
```
MStoykov, line 53:
```suggestion
We couldn't change the `--summary-export` data format because it would have broken backwards compatibility in a feature that people depended on in CI, so it still works how it used to. However, in k6 v0.30.0, we introduced `handleSummary()` - a new and better way to make JSON exports of the summary data, as well as any other format (CSV, XML (JUnit/xUnit/etc.), HTML, TXT, etc.) that may be required. We strongly recommend everyone to use `handleSummary()` instead of `--summary-export`. For more details, see the next section in this document...
```

src/data/markdown/docs/01 guides/04 Results visualization/00 End-of-test Summary.md
```diff
@@ -0,0 +1,199 @@
...
+
+## handleSummary() callback
+
+Starting with k6 v0.30.0, users can now completely customize the end-of-test summary report!
+
+You can now `export` a function called `handleSummary()` and k6 will call it at the end of the test run, after even [`teardown()`](/docs/using-k6/test-life-cycle#setup-and-teardown-stages). `handleSummary()` will be called with a JS object containing the same information that is used to generate the end-of-test summary and `--summary-export`, and allows users to completely customize how the end-of-test summary looks like.
```
MStoykov, line 59:
```suggestion
You can now `export` a function called `handleSummary()` and k6 will call it at the end of the test run, even after [`teardown()`](/docs/using-k6/test-life-cycle#setup-and-teardown-stages). `handleSummary()` will be called with a JS object containing the same information that is used to generate the end-of-test summary and `--summary-export`, and allows users to completely customize how the end-of-test summary looks like.
```

src/data/markdown/docs/01 guides/04 Results visualization/06 JSON.md
```diff
@@ -100,147 +100,7 @@ For more advanced cases, check out the [jq Manual][jq_manual_url]
...
-        "passes": 435,
-        "fails": 0
-      }
-    }
-  }
-}
-```
-
-</CodeGroup>
+If you don't care about each individual metric measurement and instead want to see only the aggregated data, exporting the end-of-test summary data to a JSON file might be a better choice than using the JSON output described here. For more details, see `--export-summary` and `handleSummary()` in the [end-of-test summary docs](/results-visualization/end-of-test-summary).
```
MStoykov, line 103: Can we then keep the `--export-summary` format from above in the new documentation about it ?

src/data/markdown/docs/01 guides/04 Results visualization/00 End-of-test Summary.md
```diff
@@ -0,0 +1,198 @@
...
+</CodeGroup>
+
+A few options can affect how this report behaves:
+- The [`--summary-trend-stats` option](/using-k6/options#summary-trend-stats) allows you to define which stats for [Trend metrics](/javascript-api/k6-metrics/trend) will be calculated and shown.
+- The [`--summary-time-unit` option](/using-k6/options#summary-time-unit) forces k6 to use a fixed time unit for _all_ time values in the summary.
+- [`no-summary`](/using-k6/options#no-summary) completely disables the report generation. Since k6 v0.30.0 that includes `--summary-export` and `handleSummary()` (see below).
+
+## Summary export to a JSON file
+
+Since k6 v0.26.0, k6 has had the [`--summary-export=path/to/file.json` option](/using-k6/options#summary-export) for local test runs. It exports some of the summary report data to a JSON file format.
```
imiric, line 49:
```suggestion
Since version 0.26.0 k6 has had the [`--summary-export=path/to/file.json` option](/using-k6/options#summary-export) for local test runs. It exports some of the summary report data to a JSON file format.
```

src/data/markdown/docs/01 guides/04 Results visualization/00 End-of-test Summary.md
```diff
@@ -0,0 +1,198 @@
...
+            "thresholds": {
+                "p(95)<500": {
+                    "ok": false
+                }
+            }
+        },
+        "http_req_duration{staticAsset:yes}": { // sub-metric from threshold
+            "contains": "time",
+            "values": {
+                // actual keys depend depend on summaryTrendStats
```
imiric, line 177:
```suggestion
                // actual keys depend on summaryTrendStats
```

imiric (commented):
Stellar work @na--! :clap: Everything is clear, well written and engaging, I don't have any major suggestions.

I also like the clarifications, deprecation notices and warnings. Things like the `--vus` and `--duration` being shortcuts for specific executors are much clearer now. :+1:

ppcano:
@na-- 

General feedback, I think we should clarify more the distinction between the `End-of-test Summary`  & `Metric Results`. 

A few suggestions in this direction:

1 - Start introducing the distinction between both result types earlier on the Results output page. 

2 - Give a unique name to both types of reports and be consistent all around. `Summary Result` & `Metric Results` was the first idea that came to my mind. I find `End-of-test Summary` to be a little wordy.

3 - The Results output page could include a section for each result type to make them visible on the Table of Content, the right sidebar.

na--:
Re: @ppcano

> 1 - Start introducing the distinction between both result types earlier on the Results output page.

:+1: I completely agree with that, will add a short explanation with internal links at the start of the page.

> 2 - Give a unique name to both types of reports and be consistent all around. Summary Result & Metric Results was the first idea that came to my mind. I find End-of-test Summary to be a little wordy.

While I also somewhat dislike `End-of-test Summary`, I like `Summary Result` even less :sweat_smile: But I'm open to other suggestions. For example, how about simply `Test Summary`?

Regarding `Metric Results`, it doesn't sound right to me at all, i.e. it doesn't really make sense or reflect what they are. Also, why do we need to change `External outputs` or `Metric outputs`, what is the problem with them?

> 3 - The Results output page could include a section for each result type to make them visible on the Table of Content, the right sidebar.

Sorry, I don't understand what you mean here

ppcano:
@na-- 

2 - If we want to describe that there are two types of "something". I think we should use the same name for both options.

Given your suggestion, what about `Summary outputs` and `Metric outputs`?



3 - I meant to use the Table of Content to highlight these concepts.

Screen Shot 2021-02-11 at 11 03 23 AM

na--:
@ppcano, but we're not describing "two types of something" :confused: The end-of-test summary is something completely different and distinct from the metric outputs... So, having "Summary outputs" would just confuse people, instead of helping them.

ppcano:
@na-- Correct me I am mistaken here. But I interpret the goal of the features as two different ways to get the results of your test. Either you look at the summary or a more-detailed result/s that include "all" the aggregated metric points.

na--:
Yes and no. You are right, they are two different ways to get the results of your test, but the key word there is _different_... So we don't need to figure out some way to shoehorn them under the same label. We already mention both in the ["Results output" page](https://mdr-ci.staging.k6.io/docs/refs/pull/212/merge/getting-started/results-output) and under the "Results visualization" menu header. This is plenty good enough, there is no need to try and force the two different concepts to have a common word in the name...

To spell out the differences clearly, external metric outputs are:
- external - k6 sends metrics to some other service or process
- granular - we either send every single metric measurement, or if there's some aggregation, it has small time periods
- inflexible - beyond some bare minimum of configuration (e.g. changing the address to which we send metrics), users can't control the external outputs
- dumb - we don't really process things like sub-metrics or thresholds or groups or checks, we just send the raw data

The end-of-test summary is:
- internal - everything happens inside the k6 process
- aggregate - you get aggregated metrics from the beginning of the test, absolutely no fine-grained metrics
- flexible - with k6 v0.30.0 and `handleSummary()`, users have almost total control on what happens with the summary
- "smart" - k6 itself knows that the `checks` and `groups` metrics are special and pre-processes them before passing them to `handleSummary()`, we also automatically calculate threshold values and separate out sub-metrics into their own sinks, etc.

So, despite both components dealing with "results", they are as different as you can make them...

ppcano:
@na-- Consider adding these differences to the docs. 

You understood well my feedback here. It's fine to disagree. Please, proceed with how you think is best.

ppcano (approved):

na--:
:+1: I'll add them to the start of the results output page

imiric (approved):

MStoykov (approved):

PR #219, opened 2021-02-12
Tweak `Results output` Table of Content

ppcano:
@na-- Previously commented at https://github.com/k6io/docs/pull/212#issuecomment-777331346

This change only updates the Table of Content.

Screen Shot 2021-02-12 at 4 37 56 AM


na-- (approved):
Ah, sorry, forgot about this in the other discussion. What do you think about bumping the `Summary export` so it's also visible in the menu?

ppcano:
I don't think so. 

I think we should not give relevance to something that is discouraged.

> the --summary-export flag, though its use is now discouraged  , 


I am merging the PR but feel free to include it if you think it is appropriated.

PR #229, opened 2021-03-02
Update the performance of SharedArray for v0.31.0

MStoykov:


src/data/markdown/docs/02 javascript api/04 k6-data/1-SharedArray.md
```diff
@@ -86,18 +86,24 @@ export default function () {
...
+| 1000       | false  | 2:01:90   | 75-77%   | 333-339MB   | 98069-98951   |
+| 10000      | true   | 2:01:70   | 78-79%   | 213-217MB   | 97953-98735   |
+| 10000      | false  | 2:03:00   | 80-83%   | 1364-1400MB | 96816-98852   |
+| 100000     | true   | 2:02:20   | 78-79%   | 238-275MB   | 98103-98540   |
+| 100000     | false  | 2:14:00   | 120-124% | 8.3-9.1GB   | 96003-97802   |
+
+In v0.30.0 the difference in CPU usage at lower numbers was around 10-15%, but it also started to
+even out at around 10k data lines and was a clear winner at 100k.
 
 The CPU/memory data comes from using `/usr/bin/time` and the raw data can be found [here](https://gist.github.com/MStoykov/1181cfa6f00bc56b90915155f885e2bb).
```
na--, line 108: You have to update this as well, or maybe even remove the gist link? I doubt it's that important to show the raw data when you have the script above...

MStoykov: done

src/data/markdown/docs/02 javascript api/04 k6-data/1-SharedArray.md
```diff
@@ -86,18 +86,24 @@ export default function () {
 
 </div>
 
-Which was ran with v0.30.0 and 100 VUs started to even out the CPU usage around 10k elements, but also was using 1/3 of the memory. At 100k `SharedArray` was the clear winner, while for lower numbers it is possible that not using it will help with performance.
+Which was ran with v0.31.0 and 100 VUs. As can be seen there isn't much of a difference at lower
+number of data lines, and it can be argued that before around 1k there is little benefit in memory
+usage. But also there is little to no difference in CPU usage as well. At 10k and above the memory
+savings start to heavily translate to CPU ones.
```
na--, line 92:
```suggestion
Which was ran with v0.31.0 and 100 VUs. As can be seen from the table below, there isn't much of a difference at lower
numbers of data lines - up until around 1000 data lines there is little benefit in memory
usage. But also there is little to no difference in CPU usage as well. At 10k and above, the memory
savings start to heavily translate to CPU ones.
```

Also, why do you insist on manually wrapping long text lines? Editors and viewers (i.e. GitHub UI) are perfectly capable of soft-wrapping long text lines, they don't need your help... The way you're split this makes it a huge PITA to view _and_ to edit it in the future...

For example, if I tile my browser window to one side of the screen, here's how shitty the new text looks here:
newlines

And notice how, if you accept the suggested edits above, the text would become super crooked, since some lines would be shorted and some longer. And even if I configure my text editor to wrap lines at the same length as yours (which I don't know, and in any case, can't do in the GitHub PR review interface...), the diff will become needlessly large. If I delete a single word at the start of a paragraph, it might bring up words from every subsequent line to fill up the "gap", i.e. which would be super confusing in the diff.

na--: In summary, this is much better, please don't wrap long lines manually:
```suggestion
Which was ran with v0.31.0 and 100 VUs. As can be seen from the table below, there isn't much of a difference at lower numbers of data lines - up until around 1000 data lines there is little benefit in memory usage. But also there is little to no difference in CPU usage as well. At 10k and above, the memory savings start to heavily translate to CPU ones.
```

cc @simskij @ppcano, I'd go as far as asking to adopt this as an official linter rule in the docs... :sweat_smile:

imiric: :thinking: I'm also guilty of wrapping long lines, and follow a similar 80-100 char convention as with code. Main reason being that it standardizes the page width for everything, and it's more pleasant to read in raw/unrendered form. After all, you run into the same issue with raw tables: there's a specific width and if your editor doesn't match it, it's going to wrap around.

But considering this isn't an issue with rendered Markdown, and you make good points about the diff, I'll stop wrapping them manually.

MStoykov: mostly as vim did it automatically. I have now disabled this for `.md` files

MStoykov: do you want to also unwrap all the other lines in this file?

na--: > Main reason being that it standardizes the page width for everything

And that is a good thing why? :sweat_smile: We don't have the same size of screens (especially @MStoykov :rofl:), or reading preferences, so why should we standardize the line width? Everyone can configure soft line wrapping however they want locally, same as tab widths and font sizes.

> and it's more pleasant to read in raw/unrendered form.

Citation needed :smile: It's definitely not more present for me - I prefer shorter line length than the above ones, for example. And every time I start editing such a sentence, I almost rage quit :sweat_smile: I am not sure how vim handles editing in the middle of a hard-line-wrapped paragraph, but my editor definitely doesn't do it well and I have to manually fiddle with newlines.

> do you want to also unwrap all the other lines in this file?

Yes please :pray:

imiric (approved):

na-- (approved):

PR #230, opened 2021-03-07
Documentation for jslib.k6.io

sniku:
- structure for jslib.k6.io documentation
 - functional lib
 - httpx lib
 - examples

WIP:
 - spelling / grammar
 - cleanup


src/data/markdown/docs/02 javascript api/11 jslib/01 httpx.md
```diff
@@ -0,0 +1,102 @@
...
+</Blockquote>
+
+
+### Example
+
+<CodeGroup labels={["httpx session used together with functional library"]}>
+
+```javascript
+import { test } from 'https://jslib.k6.io/functional/0.0.1/index.js';
+import { Httpx } from 'https://jslib.k6.io/httpx/0.0.3/index.js';
```
MStoykov, line 34: why not `httpx` ?

src/data/markdown/docs/02 javascript api/11 jslib/01 httpx.md
```diff
@@ -0,0 +1,102 @@
...
+      last_name: 'Owner',
+      username: USERNAME,
+      password: PASSWORD,
+    });
+
+    t.expect(resp.status).as("status").toEqual(201)
+      .and(resp).toHaveValidJson();
+  })
+
+  &&
```
MStoykov, line 65: maybe a comment that this is important ?

src/data/markdown/docs/02 javascript api/11 jslib/01 httpx.md
simskij: Capitalization of class names is pretty standard in JS, so from that perspective it makes sense. However, as we already have a lot of APIs which are all lowercased, maybe lowercasing would give it a more unified look.

MStoykov: `HTTPx` ? 

 I mean ... not capitalizing the whole acronym is a pretty big no no ... anywhere ... I think ?

src/data/markdown/docs/02 javascript api/11 jslib/02 functional.md
```diff
@@ -0,0 +1,38 @@
...
+
+More advanced examples can be found in the [examples section](/examples/functional-testing)
+
+
+| Function | Description |
+| -------- | ----------- |
+| [test(name, function)](/javascript-api/jslib/functional/test-name-function)  | Entry point for creating tests.  |
+| [expect(value)](/javascript-api/jslib/functional/expect-value)  | expect(value) sets the value to be used in comparison by the next function in the chain |
+| [and(value)](/javascript-api/jslib/functional/and-value)  | and(value) is similar to expect(value), but can be used in a chain. |
+| [as(alias)](/javascript-api/jslib/functional/as-string)  | as(alias) sets a textual representation of the value passed to `expect` or `and`. |
```
simskij, line 30: why is this needed? shouldn't it use the variable name as the textual representation? looks a bit like gold-plating to me.

src/data/markdown/docs/02 javascript api/11 jslib/02 functional.md
```diff
@@ -0,0 +1,38 @@
...
+
+
+See [getting started with functional.js](/javascript-api/jslib/functional/getting-started-with-functional-js) for usage examples.
+
+More advanced examples can be found in the [examples section](/examples/functional-testing)
+
+
+| Function | Description |
+| -------- | ----------- |
+| [test(name, function)](/javascript-api/jslib/functional/test-name-function)  | Entry point for creating tests.  |
```
simskij, line 27: Is there a particular reason why you opted for `test` over the BDD-style function names (`describe`, `when`, `it`) that most other js testing tools use?

src/data/markdown/docs/02 javascript api/11 jslib/01 httpx.md
simskij: Why is it that `&&` is even needed? I would expect this to become the most common source of errors for users of this lib.

src/data/markdown/docs/02 javascript api/11 jslib/functional/01 Intruduction.md
```diff
@@ -0,0 +1,92 @@
...
+
+```javascript
+import { test } from 'https://jslib.k6.io/functional/0.0.2/index.js';
+import http from 'k6/http';
+
+export default function testSuite() {
+  test('Basic API test', (t) => {
+    let response = http.get("https://test-api.k6.io/public/crocodiles")
+
+    t.expect(response.status).as("API status code").toEqual(200);
```
simskij, line 41: Why is `expect` a member of `t` rather than a separate import?

simskij (changes requested):
This mostly turned into a review of the actual libraries. When it comes to how it's documented and where you've placed it, you've managed to get it spot-on to what I was going to suggest. 😅  Do with the comments as you want.

src/data/markdown/docs/02 javascript api/11 jslib/01 httpx.md
simskij: No, that's mainly an issue in Go. In JS it's not only fine but even encouraged to capitalize it just the way @sniku did. :)

src/data/markdown/docs/02 javascript api/11 jslib/functional/10 error handling.md
```diff
@@ -0,0 +1,73 @@
...
+  at file:///home/user/happy-path-check.js:5:22(17)  executor=per-vu-iterations scenario=default source=stacktrace
+```
+
+In this example, the system was overloaded, and the load balancer returned a 503 response that did not have a valid JSON body. k6 has thrown a JavaScript exception and restarted execution from the beginning. 
+This test code is fragile to failing SUT because the first `check` does not prevent the second check from executing. 
+It's possible to rewrite this code to be less fragile, but that will make it longer and less readable. 
+
+Error handling of this type happens automatically when using the `functional.js` library.
+When the first `expect` fails, the remaining checks in the chain are not executed, and the test is marked as failed — the execution proceeds to the next `test()` instead of restarting from the top.
+
```
MStoykov, line 52: I think the naming `test` is super confusing ... so I am with @simskij on https://github.com/k6io/docs/pull/230#discussion_r589320904

src/data/markdown/docs/02 javascript api/11 jslib/01 httpx.md
MStoykov: I ... don't agree with this, based on the mdn web docs: 
examples: 
* [ignoreBOM](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoderStream/ignoreBOM)
* [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL_API) and [createObjectURL](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL) , [toJSON](https://developer.mozilla.org/en-US/docs/Web/API/URL/toJSON)
* [CSSPositionValue](https://developer.mozilla.org/en-US/docs/Web/API/CSSPositionValue) and all the other things starting with CSS 
* everything [here](https://developer.mozilla.org/en-US/docs/Web/API#interfaces) that starts with an acronym has the acronym uppercase 

Of course, there is also everybody's favorite [XMLHttpRequest](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest) which has the second acronym in not uppercase as it needs to separate it ... but we aren't in this case so :P

simskij: To my knowledge the Ecmascript standard does not concern itself with how acronyms are written.

The closest I come to some kind of trace as to why it's XMLHttp and not XmlHttp or XMLHTTP is a style-guide by Microsoft that makes a difference between long (>3) and short (<= 3) acronyms where the short ones should be uppercase while the long ones should be capitalized. 

Google's style guide seem to propose something similar: https://google.github.io/styleguide/jsguide.html#naming

src/data/markdown/docs/02 javascript api/11 jslib/02 functional.md
sniku: I had the same idea, but sadly in JS you can't use reflection to extract the variable name. I tried to work around it by stringifying the function call and regex out the variable name and that worked :smile:  until it ran through babel and babel minified the code and changed my variable name to `a`.  :sob: 

In other words, it's not possible to extract "response.status" string from code like this `t.expect(response.status).toBe(200)` 

To the solution is to add `.as("response code")` to the chain.

src/data/markdown/docs/02 javascript api/11 jslib/01 httpx.md
sniku: I will create a repo for `httpx` and `functional` so issues like this can be discussed inside the repo. 
Maybe even we should have a template repo for jslib libraries developed for k6? :thinking: What do you think @simskij ?

src/data/markdown/docs/02 javascript api/11 jslib/02 functional.md
sniku: I was inspired by `ava` more than other libs. I agree with you and @MStoykov that `test` is too ambiguous. Maybe `describe` is better here :thinking: 
I'll create a repo for this library and we can discuss it there.

src/data/markdown/docs/02 javascript api/11 jslib/functional/01 Intruduction.md
sniku: expect() raises a very specific Exception when the condition is false. The `test()` catches that exception and processes the result. If `expect` is called outside of `test()`, the exception will abort iteration and produce a very confusing stack trace. This is why `expect` is not available outside of `test`. 
Ava does it the same way, so that's what I opted for :-)

src/data/markdown/docs/02 javascript api/11 jslib/functional/10 error handling.md
sniku: Agreed. I'll rename it to something less ambiguous.

src/data/markdown/docs/02 javascript api/11 jslib/01 httpx.md
sniku: Fair point, it either needs to be documented or changed. There's a good reason for having `&&` but maybe this coding style is too confusing.

sniku:
@simskij is our auto-deploy on PR not working anymore? :thinking:

sniku:
We now have repos for both libraries
- httpx https://github.com/k6io/k6-jslib-httpx/
- functional: https://github.com/k6io/k6-jslib-functional

Let's discuss the library code there.

src/data/markdown/docs/02 javascript api/11 jslib/01 httpx.md
simskij: That could definitely be interesting! What would you envision that template to contain?

src/data/markdown/docs/02 javascript api/11 jslib/01 httpx.md
simskij: I think so. Adding a way to chain them using fluent syntax would probably be less confusing if you still want the `&&` behavior? Like `test().and.test().and.test().and.test()`?

src/data/markdown/docs/02 javascript api/11 jslib/02 functional.md
simskij: Alright. Thanks for the clarification! Babel **should** retain symbol names, however. @MStoykov do you know why this happens? Nothing in https://github.com/loadimpact/k6/blob/a85664b610d11c8b5ab22724bd38cfaa58d84ce3/js/compiler/compiler.go#L96 indicates that it would mangle symbol names, from what I can tell.

src/data/markdown/docs/02 javascript api/11 jslib/02 functional.md
simskij: > @simskij is our auto-deploy on PR not working anymore? 🤔

Our auto-deploy works, but it will only post the URL if the first build works, unfortunately. 😭 
It's either that or a new comment from the bot for every push, I'm afraid. But I'd be happy to change it to that if that's preferable.

simskij:
> We now have repos for both libraries
> 
> * httpx [k6io/k6-jslib-httpx](https://github.com/k6io/k6-jslib-httpx/)
> * functional: [k6io/k6-jslib-functional](https://github.com/k6io/k6-jslib-functional)
> 
> Let's discuss the library code there.

Awesome! Thanks, @sniku - appreciated! With that said, it sounded like you wanted to do some minor adjustments before merging? Once you're satisfied, I'm ok with having it merged.

simskij (approved):

src/data/markdown/docs/02 javascript api/11 jslib/02 functional.md
MStoykov: @simskij I don't think it has anything to do with babel. 
As far as I can understand you both want:
```javascript
var something = 12;
expect(something)
func expect(a) {
  // somehow get that a's name before the call was "something" or "response.status" here
}
```
which seems to *not* be possible in js. 

"The workaround" that is proposed around the internet is to use an object with 1 key that is the name of the variable such as 
```javascript
var something = 1241
function expect(a) {
        console.log(Object.keys(a)[0]);
}
expect({something});
var response = {status:200}

expect({response.status}) // this doesn't work though
```
Unless I am missing something or I misunderstood, this seems like a thing that we won't be able to UX around just with some code and I don't have any better proposals for the API or way to do it as I don't think `as({response_status: response.status})` is at all better so :shrug:

simskij: Yeah, I might have a look at some point, cause I know other js testing tools do it, but I'm totally fine with how it looks right now. Sorry for opening pandora's box 😂  👍🏼

src/data/markdown/docs/02 javascript api/11 jslib/01 httpx.md
ppcano: > coding style is too confusing
I think that too.

PR #235, opened 2021-03-15
Add ArrayBuffer documentation for k6 v0.31.0

### LINKED ISSUE #233: v0.31: Expanded ArrayBuffer support in most internal modules
In [v0.31](https://github.com/loadimpact/k6/releases/tag/v0.31.0)

> JS: Expanded ArrayBuffer support in most internal modules, so now you can pass ArrayBuffer to http.file(), in k6/encoding and k6/crypto functions.

Relates to https://github.com/loadimpact/k6/pull/1800

imiric:
Closes #233

### DIFF
--- a/src/data/markdown/docs/02 javascript api/01 Init context/open.md
+++ b/src/data/markdown/docs/02 javascript api/01 Init context/open.md
@@ -26,9 +26,9 @@ See example further down on this page. For more in-depth description see [Runnin
 
 ### Returns
 
-| Type           | Description                                                                              |
-| -------------- | ---------------------------------------------------------------------------------------- |
-| string / bytes | The contents of the file, read as text or bytes (if `b` has been specified as the mode). |
+| Type           | Description                                                                                          |
+| -------------- | ---------------------------------------------------------------------------------------------------- |
+| string / Array | The contents of the file, returned as string or array of numbers (if `b` was specified as the mode). |
 
 <CodeGroup labels={["users.json"]}>
 
--- a/src/data/markdown/docs/02 javascript api/03 k6-crypto/02-createHMAC- algorithm- secret -.md
+++ b/src/data/markdown/docs/02 javascript api/03 k6-crypto/02-createHMAC- algorithm- secret -.md
@@ -8,7 +8,7 @@ Creates a HMAC hashing object that can then be fed with data repeatedly, and fro
 | Parameter |  Type  | Description                                                                                                                         |
 | --------- | :----: | :---------------------------------------------------------------------------------------------------------------------------------- |
 | algorithm | string | The hashing algorithm to use. One of `md4`, `md5`, `sha1`, `sha256`, `sha384`, `sha512`, `sha512_224`, `sha512_256` or `ripemd160`. |
-| secret    | string | A shared secret used to sign the data.                                                                                              |
+| secret    | string / ArrayBuffer | A shared secret used to sign the data.                                                                                |
 
 ### Returns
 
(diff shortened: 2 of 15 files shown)


src/data/markdown/docs/02 javascript api/03 k6-crypto/03-hexEncode- input -.md
```diff
@@ -0,0 +1,42 @@
+---
+title: 'hexEncode( input )'
+description: 'Encode input data into a hexadecimal string.'
```
imiric, line 3: This was never documented, though I'm wondering if we should leave it undocumented and move it to `k6/encoding` where it arguably belongs, while also adding a `hexDecode()` function.

ppcano:
I've created the [TypeScript PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51761) related to this changes. Please, @imiric review the TypeScript changes. 



Previously discussed with @imiric in Slack; [open](https://k6.io/docs/javascript-api/init-context/open-filepath-mode) can return a  `bytes` type that is valid for [File.data](https://k6.io/docs/javascript-api/k6-http/filedata).

The doc changes in this PR removes this case and needs to be handled.

ppcano (changes requested):
https://github.com/k6io/docs/pull/235#issuecomment-799577573

imiric:
@ppcano See 6f29b53.

ppcano (approved):

src/data/markdown/docs/02 javascript api/03 k6-crypto/03-hexEncode- input -.md
na--: hmm yeah, this doesn't seem like something that should be a part of the k6 core, it should be very easy to implement in JS :confused: @MStoykov, do you remember why this was added in https://github.com/loadimpact/k6/pull/952? Because if we haven't documented it, I personally think we should just silently un-export it :sweat_smile:

MStoykov: I seemed to have used it in the tests.

 I don't use it any of my PoC awsv4 signings and https://github.com/loadimpact/k6/issues/856 has nothing specific on it. 
 
I am not really for dropping it or keeping it for that matter. I think it probably has some good use cases and we can probably re-add it if/when we hit them if we decide to drop ti

MStoykov (approved):

na-- (commented):
Besides not mentioning `hexEncode()`, the only other comment I can make here is that we might want to signify in some way that `ArrayBuffer` is supported only starting from k6 v0.31.0. We don't have versioned API documentation yet (https://github.com/k6io/docs/issues/242), so I'm not sure how we can do it in a clean way yet...

MStoykov:
> is supported only starting from k6 v0.31.0.

Like `http_req_failed` [here](https://k6.io/docs/using-k6/metrics#http-specific-built-in-metrics)

src/data/markdown/docs/02 javascript api/03 k6-crypto/03-hexEncode- input -.md
imiric: OK, I'll remove the docs for it then until we decide what to do with it.

imiric:
OK, let me know if b8bca07 is clear.

MStoykov (approved):
LGTM! 
I have no idea if using the escape is better :shrug:

na-- (approved):

PR #240, opened 2021-03-17
Update compatiblity-mode information around v0.31.0

MStoykov:


src/data/markdown/docs/01 guides/02 Using k6/19 Javascript Compatibility Mode.md
```diff
@@ -117,10 +117,9 @@ $ K6_COMPATIBILITY_MODE=extended k6 run script.js
 
 </CodeGroup>
 
-Transforms scripts to ES5.1 using Babel with ES2015 preset, and loads
-[core.js](https://github.com/zloirock/core-js) v2 polyfills. This is
-the default mode, providing greater JavaScript compatibility, at the
-expense of startup time, RAM usage and performance.
+In case of syntax/parsing error, the script will be transformed using Babel with specific plugins bringing the compatibility to ES2015(ES6)+. This means that features such as classes and arrow functions can be used. This does take some time to transpile and the produced code has slightly different line/column numbers.
```
imiric, line 120:
```suggestion
In case of syntax/parsing errors, the script will be transformed using Babel with specific plugins bringing the compatibility to ES2015(ES6)+. This means that features such as classes and arrow functions can be used. This does take some time to transpile and the produced code has slightly different line/column numbers. 
```

src/data/markdown/docs/01 guides/02 Using k6/19 Javascript Compatibility Mode.md
```diff
@@ -117,10 +117,9 @@ $ K6_COMPATIBILITY_MODE=extended k6 run script.js
 
 </CodeGroup>
 
-Transforms scripts to ES5.1 using Babel with ES2015 preset, and loads
-[core.js](https://github.com/zloirock/core-js) v2 polyfills. This is
-the default mode, providing greater JavaScript compatibility, at the
-expense of startup time, RAM usage and performance.
+In case of syntax/parsing error, the script will be transformed using Babel with specific plugins bringing the compatibility to ES2015(ES6)+. This means that features such as classes and arrow functions can be used. This does take some time to transpile and the produced code has slightly different line/column numbers. 
+
+Up to v0.31.0 it also include [core.js](https://github.com/zloirock/core-js) v2 and even more babel plugins. The core-js added around 2mb per VU of memory usage and some of the transformations (generators, async/await) of babel were not useful as they were still not enough for k6 to just work with those features. So up to this using `compatibility-mode` `base` was a significant improvement on memory usage which also translated to some CPU gains.
```
imiric, line 122:
```suggestion
Until v0.31.0 it also included [core.js](https://github.com/zloirock/core-js) v2 and even more Babel plugins. The core-js added around 2mb per VU of memory usage and some of the transformations (generators, async/await) of Babel were not useful as they were still not enough for k6 to just work with those features. So up to this using `compatibility-mode` `base` was a significant improvement on memory usage which also translated to some CPU gains.
```

I don't understand the second part of this paragraph, though (from "some of the transformations..."). Did you mean to imply that `--compatibility-mode=base` no longer has as much CPU/memory benefits? I think it should be reworded a bit.

MStoykov: I would argue it has no practical memory benefits ... babel takes some MB in memory but this is unlikely to matter and the transpiled code is unlikely to be bigger enough to matter (or the fact we have it transpiled or not).

Babel transpiling is not *fast* but it is also rarely above 1s or something like that for a simple script. Also if you just run ES5.1 tests with `extended and babel doesn't transpile anything ... it is the same as running with `compatibility-mode=base` as it will just not transpile with babel ...

src/data/markdown/docs/01 guides/02 Using k6/19 Javascript Compatibility Mode.md
```diff
@@ -117,10 +117,9 @@ $ K6_COMPATIBILITY_MODE=extended k6 run script.js
 
 </CodeGroup>
 
-Transforms scripts to ES5.1 using Babel with ES2015 preset, and loads
-[core.js](https://github.com/zloirock/core-js) v2 polyfills. This is
-the default mode, providing greater JavaScript compatibility, at the
-expense of startup time, RAM usage and performance.
+In case of syntax/parsing errors, the script will be transformed using Babel with specific plugins bringing the compatibility to ES2015(ES6)+. This means that features such as classes and arrow functions can be used. This does take some time to transpile and the produced code has slightly different line/column numbers. 
+
+Until v0.31.0 it also included [core.js](https://github.com/zloirock/core-js) v2 and even more Babel plugins. The core-js added around 2mb per VU of memory usage and some of the transformations (generators, async/await) of Babel were not useful as they were still not enough for k6 to just work with those features. So up to this using `compatibility-mode` `base` was a significant improvement on memory usage which also translated to some CPU gains.
```
na--, line 122:
```suggestion
Before k6 v0.31.0, we also included [core.js](https://github.com/zloirock/core-js) v2 and even more Babel plugins. This added around 2MB extra memory usage per VU and some of the transformations (generators, async/await) of Babel were not useful as they were still not enough for k6 to just work with those features. So before v0.31.0, using `--compatibility-mode=base` was a significant improvement on memory usage, which also translated to some CPU gains.
```

MStoykov: I don't like using `we` to mean `k6` in this context I am okay with the rest

na--: fair enough, edited

imiric (approved):

na-- (approved):

PR #243, opened 2021-03-18
Update 04 Thresholds.md to correct discrepancy.

reub-the-cube:
Correct discrepancy between 200ms (comment) and 500ms (actual code) to be 200ms on both.


na-- (approved):
Thanks!

PR #280, opened 2021-04-28
Update 02 Installation.md

aramtor:


simskij:
Thank you for the contribution!

@na--, feel free to decide whether we should include this. I've never had to do this myself as most common distros contain both gnupg and ca-certificates now a days, but since you've got a taste for weird, niche distros you're better suited to make the call (although I would assume most of those don't include `apt` 😂 )

imiric:
> I've never had to do this myself as most common distros contain both gnupg and ca-certificates now a days

I think so too. This would only be needed if building a Docker image since neither `ubuntu` or `debian` carry it in the base image, but they are included in the desktop distributions AFAIK.

So we might want to add it as a note for Docker, but it's not needed as part of the main instructions.

aramtor:
> So we might want to add it as a note for Docker, but it's not needed as part of the main instructions.

That would be awesome, thanks guys!

simskij:
Ok, @imiric will have to serve as @na-- for now. I've rephrased it and will merge it. Thanks again @aramtor 🙏🏼

PR #281, opened 2021-04-30
Fix errors in code example

coryhouse:
The current example doesn't run. This resolves the issues.


simskij:
Thank you for your contribution! 🙏🏼

PR #285, opened 2021-05-03
Linking to examples

tom-miseur:
Adds a link to data correlation examples in the `Response.body` description


src/data/markdown/docs/02 javascript api/07 k6-http/61 Response.md
```diff
@@ -10,7 +10,7 @@ Note that in the case of redirects, all the information in the Response object w
...
+| `Response.body`                                                                                | string   | Response body content, often used to extract dynamic data (see examples [here](/examples/correlation-and-dynamic-data)) and when verifying the presence of content using [checks](/javascript-api/k6/check-val-sets-tags).<br /><br />See [Params.responseType](/javascript-api/k6-http/params) and [options.discardResponseBodies](/using-k6/options) for how to discard the body when it is not needed (and to save memory) or when handling bodies with binary data.                                                     |
```
ppcano, line 13: @tom-miseur 

Perhaps it is the layout and format but I am not fully convinced about adding so many related references. But I also think additional links to related info are helpful. 

IMO, I would skip the `and when verifying the presence of content using [checks]` but feel free to merge it.

tom-miseur: Maybe each property should have an associated sub-page like [Response.clickLink( [params] )](https://mdr-ci.staging.k6.io/docs/refs/pull/285/merge/javascript-api/k6-http/response/response-clicklink-params) where we can delve into examples instead?

ppcano: @tom-miseur 

For each property, I think it will be too much. 

If we want to extend significantly the documentation about one or many properties, we can create a new page that is visible or invisible from the sidebar. 

Anyway, I was just sharing openly my initial thoughts and I think the changes are good improvements. Go ahead with what you think is best.

ppcano:
@tom-miseur can we move this forward?

PR #290, opened 2021-05-10
Add netdata.md

OdysLam:
Thus PR adds instructions on using [netdata](https://netdata.cloud) with k6. 

Netdata already supports K6 out-of-the-box, meaning that the user doesn't have to do anything. Instead, the user can focus on running the k6 experiments, while Netdata will visualize everything.


simskij:
Great work @OdysLam! Really excited to see this!

simskij:
@ppcano, how should we approach this? In this case, we have new content that does not have a Spanish translation, hence the build failing. It shouldn't really be a requirement that everything is translated for the build to succeed, right?

ppcano:
@simskij Yes. We can duplicate it in English while we figure out and fix it.

OdysLam:
Thanks, guys for the merge. Thanks to you I spent some time in our awesome StatsD collector. 

I really loved K6 and its simplicity. Keep up the awesome work :)

PR #294, opened 2021-05-14
feat: added sudo on install

johnnyvaz1:


imiric:
Thanks for this! Would you mind updating the [Spanish version](https://github.com/k6io/docs/blob/6028a7b23a9d0af530cbf6dd0d80877e40d2f298/src/data/markdown/translated-guides/es/01%20Getting%20started/02%20Installation.md) as well?

PR #309, opened 2021-05-28
feat: added xk6-dotenv

szkiba:
[xk6-dotenv](https://github.com/szkiba/xk6-dotenv) is a k6 extension that loads env vars from a .env file

### DIFF
--- a/src/data/ecosystem/extensions.js
+++ b/src/data/ecosystem/extensions.js
@@ -445,6 +445,18 @@ const extensions = [
     official: false,
     categories: ['Misc'],
   },
+  {
+    name: 'xk6-dotenv',
+    description: 'A k6 extension that loads env vars from a .env file',
+    url: 'https://github.com/szkiba/xk6-dotenv',
+    logo: '',
+    author: {
+      name: 'Iván Szkiba',
+      url: 'https://github.com/szkiba',
+    },
+    official: false,
+    categories: ['Misc'],
+  },
 ];
 
 export default extensions;


imiric (approved):
You're too productive @szkiba! :smile: :clap:

Thanks for this and the amount of work you're putting into k6! :bow:

szkiba:
> You're too productive @szkiba!  
> 
> Thanks for this and the amount of work you're putting into k6!

Hi @imiric , this is just a quick development. Today I started to write a k6 test for one of my service, and noticed that I miss this feature :) (set credentials from .env file for example)

imiric:
BTW you can also do this outside of k6 with some [shell tricks](https://stackoverflow.com/questions/19331497/set-environment-variables-from-file-of-key-value-pairs). This is not cross platform or very user friendly, and it doesn't have the hierarchy feature, so I'm sure many will find your extension useful, but personally I'd prefer not having to compile an extension for it. :)

szkiba:
> BTW you can also do this outside of k6 with some [shell tricks](https://stackoverflow.com/questions/19331497/set-environment-variables-from-file-of-key-value-pairs). This is not cross platform or very user friendly, and it doesn't have the hierarchy feature, so I'm sure many will find your extension useful, but personally I'd prefer not having to compile an extension for it. :)

Yes, I used several shell/runner solution in past. These has some advantages (ie when you leave the dir, variables unset, etc). But yes, supporting it from app itself also has advantages :) and I already have to use custom k6 build (becasue of other extensions) so  it is not a problem :)

PR #319, opened 2021-06-09
Light versioning builds

vi-nastya:
This PR adds `JS_API_VERSIONS_TO_BUILD` env param to limit number of versions for the build (e.g. if it's set to `2`, then the build will contain the latest version and the most recent of the older versions).

Note: in `production` this option is ignored and all versions are included.

**How to use it for local builds**
Please make sure that your `.env.development` is up to date with the `.env.example` and has value for  `JS_API_VERSIONS_TO_BUILD`


src/utils/versioning.js
```diff
@@ -3,4 +3,10 @@ const SUPPORTED_VERSIONS = ['v0.31'];
 /** latest version number for URLs without version prefix */
 const LATEST_VERSION = 'v0.32';
 
-module.exports = { SUPPORTED_VERSIONS, LATEST_VERSION };
+const DEFAULT_JS_API_VERSIONS_TO_BUILD = 2;
```
ppcano, line 6: @vi-nastya FYI.  I have added this constant to avoid that the contributors have to define the `JS_API_VERSIONS_TO_BUILD` env variable on the `env.development` file.

PR #340, opened 2021-06-14
Update 09 StatsD.md

dmytro-bazavluk-imanage:


dmytro-bazavluk-imanage:
the pr is wrong, and the actual default prefix is `k6.` which is replaced with the value from K6_STATSD_NAMESPACE, including the dot (`.`). Probably bug in the code and not in the implementation?

PR #344, opened 2021-06-29
Archive v0.32.0 API docs, fix archive script

imiric:
I split this apart from the upcoming v0.33.0 docs PR to make it somewhat easier to review. Review by commit, and mostly ignore faa875a. We do really need to get rid of the copying...


src/utils/versioning.js
```diff
@@ -3,7 +3,7 @@ const SUPPORTED_VERSIONS = ['v0.31', 'v0.32'];
 /** latest version number for URLs without version prefix */
 const LATEST_VERSION = 'v0.33';
 
-const DEFAULT_JS_API_VERSIONS_TO_BUILD = 2;
+const DEFAULT_JS_API_VERSIONS_TO_BUILD = 10;
```
ppcano, line 6: @imiric I will rollback this change later. 

The goal is to improve building time by building only the latest two versions for the `development` and `staging` environments.  For `production`, it builds all the existing versions.

imiric: Hmm OK, I'll remove it. I wasn't sure why I wasn't seeing the older versions locally, so I'll add a comment instead.

MStoykov (approved):

na-- (approved):
LGTM, though I didn't realize we had to hardcode versions in URLs :sob: this is the first time I'm seeing `archive-js-api-version.sh` :scream:

PR #372, opened 2021-08-10
Add info about __VU variable max values based on instance tiers. 

### LINKED ISSUE #332: add 'Cloud considerations' in __VU and __ITER article
This is for https://k6.io/docs/using-k6/execution-context-variables/ , under the __VU and __ITER subheader.

Tier 1 is used when there are 1-999 VUs in a load zone
Tier 2 is used when there are 1000-4001 VUs in a load zone
Tier 3 is used when there are more than 4001 VUs in a load zone.
The tier 1 server handles up to 300VUs
The tier 2 server handles up to 1200VUs
The tier 3 server handles up to 5000VUs.
Regardless of the tier, the amount of resources (CPU, Memory, Network) per VU is the same.

For example, if you start a test with 900VUs, we will use 3x Tier 1 servers.
If you start a test with 1000VUs in a single load zone, we will use 1x Tier 2 server. If the same test is started in 2 load zones, there will be 500VUs per load zone and 4x Tier 1 servers will be used.

Focus on the crossing point to where multiple servers are used, rather than the mathematical exercise above.

Griatch:
Resolves #332.

I ended up adding more of the 'math' to this (despite the recommendation in the ticket) since that seemed prudent to understand what was going on. 

As a note - the 'staging' version of the docs doesn't seem to have a freeform search? It's often a nightmare to figure out just where these pages are actually ending up in the final documentation and the naming scheme of individual files is beyond confusing ...

### DIFF
--- a/src/data/markdown/translated-guides/en/02 Using k6/12 Execution context variables.md
+++ b/src/data/markdown/translated-guides/en/02 Using k6/12 Execution context variables.md
@@ -41,21 +41,20 @@ A numeric counter with the current iteration number for a specific VU. Zero-base
 
 Current VU number in use. The value is assigned incrementally for each new VU instance, starting from one. The variable will be 0 while executing the setup and teardown functions.
 
-It's important to remember that this value is shown _per server instance running you test_. How many servers we use depends on the size of your test and how you spread it across different load zones. So for a larger test, the value of `__VU` will likely not reflect the actual _total_ number of VUs in your test at a given moment.
+### ⚠️ To note when running in the k6 Cloud
 
-- For tests using **1-999** VUs per load zone, we use server instances that handle 300 VU's per machine. So for a 900 VU test using one load zone, there will be three instances and the value of `__VU` will vary from 1-300 per machine.
-- For tests using **1000-4001** VUs per load zone, we use server instances that handle 1200 VUs per machine. So for a 1000 VU, 1-load-zone test, your `__VU` value will go all the way to 1000.
+When you run your tests in the [k6 Cloud](/cloud) you will have additional [environment variables](/using-k6/environment-variables) that will tell you on which server, load zone and distribution of the test you are currently executing. You can read more about them [here](/using-k6/environment-variables). 
+
+The **\_\_VU** environment variable may also behave differently while running a test in the cloud. This is because the **\_\_VU** value is shown _per server instance running you test_. How many servers we use depends on the size of your test and how you spread it across different load zones. So for a larger test, the value of **\_\_VU** will likely not reflect the actual _total_ number of VUs in your test at a given moment.
+
+- For tests using **1-999** VUs per load zone, we use server instances that handle 300 VU's per machine. So for a 900 VU test using one load zone, there will be three instances and the value of **\_\_VU** will vary from 1-300 per machine.
+- For tests using **1000-4001** VUs per load zone, we use server instances that handle 1200 VUs per machine. So for a 1000 VU, 1-load-zone test, your **\_\_VU** value will go all the way to 1000.
 - For tests with **>4001** VUs per load zone, we use instances that handle 5000 VUs per machine.
 
-As an example, if you run a 1000 VU test evenly distributed over two load zones, there are 500 VUs per load zone. Since this is in the 1-999 VU bracket, we'll use server instances that handle 300VUs apiece. This means 2 instances per load zone and your individual `__VU` values will thus vary from 1 to 250.
+> Note that these are the _defaults_. If you have specific requirements, please contact k6 support for a custom solution.
 
+As an example, if you run a 1000 VU test evenly distributed over two load zones, there are 500 VUs per load zone. Since this is in the 1-999 VU bracket, we'll use server instances that handle 300VUs apiece. This means 2 instances per load zone and your individual **\_\_VU** values will thus vary from 1 to 250.
 
-> ### ⚠️ Additional context information available in the k6 Cloud
->
-> If you are running a test in [k6 Cloud](/cloud) you will have additional
-> [environment variables](/using-k6/environment-variables) that will tell you on which server, load zone
-> and distribution of the test you are currently executing. You can read more about them
-> [here](/using-k6/environment-variables)
 
 ## k6 Test Coordinator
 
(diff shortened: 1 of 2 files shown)


src/data/markdown/translated-guides/en/02 Using k6/12 Execution context variables.md
```diff
@@ -31,17 +31,24 @@ $ k6 run --vus 10 --iterations 100 script.js
...
 
 A numeric counter with the current iteration number for a specific VU. Zero-based.
 
 ### \_\_VU
 
-Current VU number. The value is assigned incrementally for each new VU instance. One-based.
-However, VU number is 0 while executing the setup and teardown functions.
+Current VU number in use. The value is assigned incrementally for each new VU instance, starting from one. The variable will be 0 while executing the setup and teardown functions.
+
+It's important to remember that this value is shown _per server instance running you test_. How many servers we use depends on the size of your test and how you spread it across different load zones. So for a larger test, the value of `__VU` will likely not reflect the actual _total_ number of VUs in your test at a given moment.
```
MStoykov, line 44: This makes it sound as k6 OSS supports multiple servers running. 
I do think this should be in a *sub* section that is called "Cloud considerations" or something like that so it's pretty obvious it is currently only about the cloud.

src/data/markdown/translated-guides/en/02 Using k6/12 Execution context variables.md
```diff
@@ -31,17 +31,24 @@ $ k6 run --vus 10 --iterations 100 script.js
...
+
+- For tests using **1-999** VUs per load zone, we use server instances that handle 300 VU's per machine. So for a 900 VU test using one load zone, there will be three instances and the value of `__VU` will vary from 1-300 per machine.
+- For tests using **1000-4001** VUs per load zone, we use server instances that handle 1200 VUs per machine. So for a 1000 VU, 1-load-zone test, your `__VU` value will go all the way to 1000.
+- For tests with **>4001** VUs per load zone, we use instances that handle 5000 VUs per machine.
```
MStoykov, line 48: I would add a note that this is the *default* and we can change it so it might be different for clients with different needs.

MStoykov (commented):
Overall pretty good, I just thing we should note that this is about cloud specifically

src/data/markdown/translated-guides/en/02 Using k6/12 Execution context variables.md
Griatch: Good point. I restructured the text to include all the k6-cloud specific execution context envvar info under one header.

src/data/markdown/translated-guides/en/02 Using k6/12 Execution context variables.md
```diff
@@ -31,24 +31,30 @@ $ k6 run --vus 10 --iterations 100 script.js
...
+
+### ⚠️ To note when running in the k6 Cloud
+
+When you run your tests in the [k6 Cloud](/cloud) you will have additional [environment variables](/using-k6/environment-variables) that will tell you on which server, load zone and distribution of the test you are currently executing. You can read more about them [here](/using-k6/environment-variables). 
+
+The **\_\_VU** environment variable may also behave differently while running a test in the cloud. This is because the **\_\_VU** value is shown _per server instance running you test_. How many servers we use depends on the size of your test and how you spread it across different load zones. So for a larger test, the value of **\_\_VU** will likely not reflect the actual _total_ number of VUs in your test at a given moment.
+
+- For tests using **1-999** VUs per load zone, we use server instances that handle 300 VU's per machine. So for a 900 VU test using one load zone, there will be three instances and the value of **\_\_VU** will vary from 1-300 per machine.
```
ppcano, line 50: @Griatch 

I suggest moving all this part out of the k6 OSS guides. 

I think it is better to link to the k6 Cloud docs where we document this part. Otherwise, the information is redundant in both sections and harder to maintain.  I suggest something like:



> You can read [here](https://k6.io/docs/cloud/cloud-faq/general-questions/#how-many-vus-can-be-run-from-the-same-dedicated-ip) how the k6 Cloud allocates VUs per server (aka load generator).

src/data/markdown/translated-guides/en/02 Using k6/12 Execution context variables.md
```diff
@@ -31,24 +31,30 @@ $ k6 run --vus 10 --iterations 100 script.js
...
 
 A numeric counter with the current iteration number for a specific VU. Zero-based.
 
 ### \_\_VU
 
-Current VU number. The value is assigned incrementally for each new VU instance. One-based.
-However, VU number is 0 while executing the setup and teardown functions.
+Current VU number in use. The value is assigned incrementally for each new VU instance, starting from one. The variable will be 0 while executing the setup and teardown functions.
+
+### ⚠️ To note when running in the k6 Cloud
```
ppcano, line 44: If the section becomes shorter, I suggest wrapping the whole new addition within a blockquote

> **Running in the k6 Cloud**
> ......

src/data/markdown/translated-guides/en/02 Using k6/12 Execution context variables.md
```diff
@@ -31,24 +31,30 @@ $ k6 run --vus 10 --iterations 100 script.js
...
+
+When you run your tests in the [k6 Cloud](/cloud) you will have additional [environment variables](/using-k6/environment-variables) that will tell you on which server, load zone and distribution of the test you are currently executing. You can read more about them [here](/using-k6/environment-variables).
```
ppcano, line 46: The link about the additional variables is wrong -> https://k6.io/docs/cloud/creating-and-running-a-test/cloud-tests-from-the-cli/#environment-variables

ppcano (changes requested):

src/data/markdown/translated-guides/en/02 Using k6/12 Execution context variables.md
Griatch: Ok, I removed this and linked. I added Mihail's mention of these values 'just being defaults' to the main text though, it seems to make sense there too.

src/data/markdown/translated-guides/en/02 Using k6/12 Execution context variables.md
Griatch: Made it a block quote indeed.

ppcano:
This PR was accidentally ported to #374. It is already merged. Closing it.

PR #379, opened 2021-08-17
Typo

Lamera:
Fix the heading, so that it appears in the navigation as the other executors.


ppcano:
@Lamera, thanks for catching this one and the PR. 💯

PR #380, opened 2021-08-18
Additional info about sleep(t)

rhires:
Copied from a community forum post by one of the k6 team members, and slightly updated text

### DIFF
--- a/src/data/markdown/docs/02 javascript api/02 k6/sleep- t -.md
+++ b/src/data/markdown/docs/02 javascript api/02 k6/sleep- t -.md
@@ -10,7 +10,27 @@ Suspend VU execution for the specified duration.
 | --------- | ------ | --------------------- |
 | t         | number | Duration, in seconds. |
 
+### More information
+Sleep doesn’t impact the test duration. It will impact how quickly the VUs can complete iterations of the default function and thus how many requests are made.
+
+Consider the example below. If it for 60 seconds with 1 VU, it would make roughly 60 total requests/iterations of the script. If sleep(t) is changed to 60 seconds, it would make 1 iteration/request. 
+
+Sleep is an idle time for VUs. It’s mostly used for script pacing, emulating user behavior, and just load generation stability. VUs will naturally execute as quickly as possible so if you get too aggressive, you can eat up all the resources on the load generation and skew results.
+
 ### Examples
+<CodeGroup labels={[]}>
+
+```javascript
+import http from "k6/http";
+import { sleep } from "k6";
+
+export default function() {
+  let res = http.get("https://loadimpact.com");
+  sleep(1);
+};
+```
+
+</CodeGroup>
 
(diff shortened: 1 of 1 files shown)


ppcano:
@tom-miseur supersedes by https://github.com/grafana/k6-docs/pull/381 ?

src/data/markdown/docs/02 javascript api/02 k6/sleep- t -.md
```diff
@@ -10,7 +10,27 @@ Suspend VU execution for the specified duration.
 | --------- | ------ | --------------------- |
 | t         | number | Duration, in seconds. |
 
+### More information
+Sleep doesn’t impact the test duration. It will impact how quickly the VUs can complete iterations of the default function and thus how many requests are made.
+
+Consider the example below. If it for 60 seconds with 1 VU, it would make roughly 60 total requests/iterations of the script. If sleep(t) is changed to 60 seconds, it would make 1 iteration/request. 
+
+Sleep is an idle time for VUs. It’s mostly used for script pacing, emulating user behavior, and just load generation stability. VUs will naturally execute as quickly as possible so if you get too aggressive, you can eat up all the resources on the load generation and skew results.
+
```
MStoykov, line 19: Hi  @rhires, thanks for the PR, but I think https://github.com/grafana/k6-docs/pull/381/files is more accurate description.

> Sleep doesn’t impact the test duration.

I mean ... it technically does - if you run just 10 iterations with 1 VU (for example) if you don't have sleep it will just do the request(s) and stop. But with a sleep it will sleep (and wait) on that line until the specified time elapses. So if we sleep 1 second it will take at least 10s for the test to finish and it's duration will be at least 10s while if the requests are fast (100ms let say) it will finish in around 1s. 

While this might not be true for duration based executors, where this will mostly change the pace of starting requests (as mentioned in the other PR and the final paragraph here)

>  Consider the example below. If it for 60 seconds with 1 VU, it would make roughly 60 total requests/iterations of the script. If sleep(t) is changed to 60 seconds, it would make 1 iteration/request. 

this also is technically wrong. k6 mostly does iterations which are executions of the `default` functions. In your case you have one request and a sleep. If you increase the sleep the only thing that will change is that iterations will take *longer*. Depending on the executor that means that you will make less iterations or k6 (with arrival rate executors) will try to make the same amount of iterations but they will just take longer. But the iterations/request will still be 1 as you have 1 request in the iteration so I would mostly consider it 1iteration/request ;). 

I am not really certain if we don't need to explain something more as obviously there is some confusion around this

tom-miseur:
> @tom-miseur, Is this PR superseded by #381 ?

Yes. I felt like the answer from the forum had some inaccuracies as pointed out by @MStoykov

ppcano:
Closing in favor of #381.  

@rhires, it is very appreciated that you try to help here. If you want to add any additional improvement, please, feel free to open another PR on top of the latest changes (#381).

Issue #321, opened 2021-06-10
Add: Example JSON output with Docker


FloorD:
We have had a couple of users asking about this process, it would be good to have it documented. Could be adding a section to the existing JSON output article: https://k6.io/docs/results-visualization/json

PR #384, opened 2021-08-19
Show docker example for output JSON data

Griatch:
This aims to resolve https://github.com/grafana/k6-docs/issues/321 .

This adds a pane explaining how to output JSON data when running k6 from docker. I also tested this but made the instruction OS-generic by generalizing the names for input/outputdirs (the `$PWD` doesn't exist for Windows shells in the same way after all).


ppcano:
Thanks, @Griatch! 💯

Issue #74, opened 2020-08-12
Mention support for HTTP_PROXY, HTTPS_PROXY, NO_PROXY


na--:
https://community.k6.io/t/k6-http-header-anomaly-how-to-config-fiddler-for-k6/137/6 made me look this up, and we don't seem to mention proxy support at all... 

k6 [uses](https://github.com/loadimpact/k6/blob/0e94653c9954c17b2b1a480b456e23ec120feff2/js/runner.go#L171) Go's `http.ProxyFromEnvironment`, so we can pretty much copy and adapt this documentation: https://golang.org/pkg/net/http/#ProxyFromEnvironment

na--:
As evidenced by https://github.com/loadimpact/k6/issues/1062, we should mention that the proxy configs also work in other contexts, e.g. sending metrics to our cloud or other outputs.

ppcano:
Documented at [k6 Load Testing Debugging Using a Web Proxy](https://k6.io/blog/k6-load-testing-debugging-using-a-web-proxy/)

simskij:
@ppcano this should probably also be documented in a dryer tone in the docs

na--:
Agree with @simskij, we should at least have a mention in the docs about these variables, possibly with a reference to the blog article, so that when someone searches `proxy` in the docs, they are not left hanging. 

Moreover, HTTP proxies are not used only for debugging, so the blog article doesn't cover all use cases... k6 users in corporate environments often need to use a proxy to develop their load tests, due to security requirements.

PR #385, opened 2021-08-19
Add mentions of HTTP_PROXY envvars. 

Griatch:
This aims to resolve issue #74 by adding the previously unmentioned `PROXY` envvars also supported by k6 to docs. I have not tested using any of these, but I presume k6 people know what they support ;)

### DIFF
--- a/src/data/markdown/translated-guides/en/02 Using k6/11 Environment variables.md
+++ b/src/data/markdown/translated-guides/en/02 Using k6/11 Environment variables.md
@@ -93,3 +93,12 @@ behavior of k6:
 | `K6_CLOUD_TOKEN`     | An authentication token to use in API calls to the cloud service, when the --cloud results output option is specified. |
 | `K6_NO_USAGE_REPORT` | Define this to disable [usage reports](/misc/usage-collection).                                                        |
 | `K6_OUT`             | Specify results output, same as --out command-line option.                                                             |
+
+k6 also allows proxying via its use of [Golang's standard http library](https://pkg.go.dev/net/http#ProxyFromEnvironment).
+This is useful e.g. when wanting to send data to [k6 cloud](/cloud) from a k6 running inside a restricted network.
+
+| Name                 | Description                                                          |
+| -------------------- | ---------------------------------------------------------------------|
+| `HTTP_PROXY`         | Given either as complete URL or as `host:port` for HTTP requests.    |
+| `HTTPS_PROXY`        | Given either as complete URL or as `host:port` for HTTPS requests.   |
+| `NO_PROXY`           | Don't use any proxy for requests.                                    |


ppcano:
I have some doubts and want to bring @tom-miseur to the conversation. 

As far as I read on the original issue, the common use case would be to proxy to Fiddler.  We discussed creating a Fiddler tutorial. Perhaps, it is good to add the docs when working on the tutorial. 

Also, I am unsure about the section: `Environment variables k6 will read automatically`. Should not these options be moved to the Options page?

Griatch:
@ppcano I appears like https://github.com/grafana/k6-docs/blob/master/src/data/markdown/translated-guides/en/02%20Using%20k6/05%20Options.md doesn't deal with cloud execution at all today. I could see the point of moving these into `options`, but I suspect the cloud options will be quite hidden away in that huge options-page though.

ppcano:
> I suspect the cloud options will be quite hidden away in that huge options-page though

Agree. But I consider it a different issue to solve.

MStoykov:
I don't think that those should be moved to the Options - they are technically something coming from go not k6 and I prefer not to mingle those together.

I have "checked" and it seems to work. By check here I mean that I set it and I got errors for the correct endpoints with `-o cloud` which means that it at least went to the proxy ... the fact there wasn't one is a different manner.

ppcano:
@MStoykov I meant the section [Environment variables k6 will read automatically](https://k6.io/docs/using-k6/environment-variables/#environment-variables-k6-will-read-automatically) listing the options: `K6_OUTPUT`, `K6_CLOUD_HOST`, `K6_CLOUD_TOKEN `, and `K6_NO_USAGE_REPORT`. wdyt?

tom-miseur:
One thing I found is that passing in proxy settings _via_ k6, i.e. through the `-e` flag, does not work; they actually need to be set against the OS for them to take effect. I guess that might be obvious to some, but we do call things passed in via `-e` environment variables as well, so there may be some users who think it might be possible (happened to me!).

I will make a note of that on the blog post, but maybe it's worth noting here as well?

Griatch:
@tom-miseur I think the docs refer to these as `execution variables`, that is, things that will be made available inside the script later. I can link to the `environment variables` page just to make this clearer. But @ppcano needs to make the call if this even belongs on this page or not first.

tom-miseur:
> @tom-miseur I think the docs refer to these as `execution variables`

Not sure about that - https://k6.io/docs/using-k6/options#supply-environment-variables is the section that talks about them, and indeed the flag itself can be either `-env` or `-e`. I would much prefer we called them `execution variables` though to clear up any confusion!

Griatch:
@tom-miseur You are right, I was thinking of [execution context variables](https://k6.io/docs/using-k6/execution-context-variables/), but I see that page doesn't actually mention the use of `-e`.

Griatch:
@tom-miseur In lieu of figuring out any new names for things, I just added a mention to the page to make it clear these are system-level envvars and not for passing with `-e`.

src/data/markdown/translated-guides/en/02 Using k6/11 Environment variables.md
```diff
@@ -84,12 +84,22 @@ system environment variables with the same name.
 
 ## Environment variables k6 will read automatically
 
-k6 will also try to read some specific environment variables that the user can set to change the
-behavior of k6:
+k6 will also try to read some specific environment variables that the user can
+set to change the behavior of k6. These must be supplied to k6 at the system level
+(so not using `-e` or `--env`).
```
MStoykov, line 89: can you link this to https://k6.io/docs/using-k6/options/#supply-environment-variables

src/data/markdown/translated-guides/en/02 Using k6/11 Environment variables.md
```diff
@@ -84,12 +84,22 @@ system environment variables with the same name.
...
 | `K6_OUT`             | Specify results output, same as --out command-line option.                                                             |
+
+k6 also allows proxying via its use of [Golang's standard http library](https://pkg.go.dev/net/http#ProxyFromEnvironment).
+This is useful e.g. when wanting to send data to [k6 cloud](/cloud) from a k6 running inside a restricted network.
+
+| Name                 | Description                                                          |
+| -------------------- | ---------------------------------------------------------------------|
+| `HTTP_PROXY`         | Given either as complete URL or as `host:port` for HTTP requests.    |
+| `HTTPS_PROXY`        | Given either as complete URL or as `host:port` for HTTPS requests.   |
+| `NO_PROXY`           | Don't use any proxy for requests.                                    |
```
MStoykov, line 105: Maybe we should expand on the values for NO_PROXY based on https://pkg.go.dev/golang.org/x/net/http/httpproxy#Config

src/data/markdown/translated-guides/en/02 Using k6/11 Environment variables.md
```diff
@@ -84,12 +84,22 @@ system environment variables with the same name.
...
 | `K6_NO_USAGE_REPORT` | Define this to disable [usage reports](/misc/usage-collection).                                                        |
 | `K6_OUT`             | Specify results output, same as --out command-line option.                                                             |
+
+k6 also allows proxying via its use of [Golang's standard http library](https://pkg.go.dev/net/http#ProxyFromEnvironment).
+This is useful e.g. when wanting to send data to [k6 cloud](/cloud) from a k6 running inside a restricted network.
+
+| Name                 | Description                                                          |
+| -------------------- | ---------------------------------------------------------------------|
+| `HTTP_PROXY`         | Given either as complete URL or as `host:port` for HTTP requests.    |
+| `HTTPS_PROXY`        | Given either as complete URL or as `host:port` for HTTPS requests.   |
```
MStoykov, line 104:
```suggestion
| `HTTP_PROXY`         | The proxy address given either as complete URL or as `host:port` that will be used for HTTP requests unless overwritten by NO_PROXY.    |
| `HTTPS_PROXY`        |  The proxy address given either as complete URL or as `host:port` that will be used for HTTPS requests unless overwritten by NO_PROXY.     |
```

Griatch: I was too simplistic with the `NO_PROXY` indeed. I'll update, thanks!

src/data/markdown/translated-guides/en/02 Using k6/11 Environment variables.md
Griatch: Yes that makes sense.

src/data/markdown/translated-guides/en/02 Using k6/11 Environment variables.md
```diff
@@ -84,12 +84,27 @@ system environment variables with the same name.
...
 | `K6_OUT`             | Specify results output, same as --out command-line option.                                                             |
+
+k6 also allows proxying via its use of [Golang's standard http library](https://pkg.go.dev/net/http#ProxyFromEnvironment).
+This is useful e.g. when wanting to send data to [k6 cloud](/cloud) from a k6 running inside a restricted network.
+
+| Name                 | Description                                                          |
+| -------------------- | ---------------------------------------------------------------------|
+| `HTTP_PROXY`         | Proxy address (URL or `host:port`) to use for HTTP requests, unless overridden by `NO_PROXY`.    |
+| `HTTPS_PROXY`        | Proxy address (URL or `host:port`) to use for HTTPS requests, unless overridden by `NO_PROXY`.   |
+| `NO_PROXY`           | Hosts that should bypass proxies. This is a comma-separated string of IP addresses and/or host names. See [the httpproxy docs][httpproxyconfig] for more info. Example: `"1.2.3.4, 2.3.4.5:80, myhost.com"`.|
```
MStoykov, line 105: not sure about the spaces in the example :thinking:

Griatch:
Addressed @MStoykov's feedback; added more details around how to spec `NO_PROXY` and also cleaned up a little more of that table.

MStoykov (approved):
Looks good ,but you are the only person that uses those types of links, which I am not against, I am just worried they will get broken by someone :rofl:

src/data/markdown/translated-guides/en/02 Using k6/11 Environment variables.md
Griatch: Fair enough, I removed them to be on the safe side. :sweat_smile:

Griatch:
@MStoykov 

> Looks good ,but you are the only person that uses those types of links, which I am not against, I am just worried they will get broken by someone rofl

They make sense in a table I think, just to make it a little easier/shorter to edit (but I didn't translate them everywhere). It's standard markdown so it should not be an issue I hope.

ppcano:
I think this PR should be superseded by #387 and [k6 Load Testing Debugging Using a Web Proxy](https://k6.io/blog/k6-load-testing-debugging-using-a-web-proxy/)

The blog post describes the Proxy settings like in this PR.  Next, we can think about how to reorganize the docs regarding Debugging and the HTTP Proxy settings in new PR.

I suggest closing this PR and merging #387

MStoykov:
I am fine with that @ppcano but you will need to update the blog post with the better explanation about NO_PROXY ;)

ppcano:
@MStoykov I missed including the latest changes. I will correct it now. Thank you!!

PR #403, opened 2021-09-10
Fix example for externally controlled executor

yorugac:


ppcano:
Thanks so much for helping with the k6 docs, @yorugac.

Issue #325, opened 2021-06-11
Document `config.json`


FloorD:
For https://k6.io/docs/using-k6/options#config add an example file as "An example of a config file is available here" currently leads nowhere.

MStoykov:
It probably needs to go to the `config.json` file which is the second example in this [section](https://k6.io/docs/using-k6/options/#using-options)

monizb:
Hi! Can I take up this issue?

ppcano:
@monizb, you are welcome! Let us know if you have any questions.

monizb:
> @monizb, you are welcome! Let us know if you have any questions.

Thank you! I did indeed have a question about this, do you want me to link that part to the [section](https://k6.io/docs/using-k6/options/#using-options) or to the config.json file itself? If its the latter, would that be possible without a header tag?

ppcano:
Yes. Ideally, to the `config.json` example. 

I think you can wrap the previous description around a `<div id="config-json-example"` and use an anchor link.

PR #416, opened 2021-10-01
Fixed broken Anchor Link

monizb:
The link https://k6.io/docs/using-k6/options#config was leading to nowhere, fixed it so that It points to https://k6.io/docs/using-k6/options/#config-json-example

Fixes: #325


ppcano:
Thanks, @monizb

Issue #315, opened 2021-06-03
Document --quiet option


ppcano:
The [k6 options docs](https://k6.io/docs/using-k6/options/) should document the `--quiet/-q` option that disables the progress update bar on the console output.

See the difference when running a simple script.

```
k6 run script.js -d 20s
k6 run script.js -d 20s --quiet
```

The command is valid for `k6 run` and `k6 cloud` commands.

imrishabh18:
@FloorD Can I take this up? And can you guide me what should I look into to resolve this issue please.

ppcano:
@imrishabh18 updated the issue description.

ppcano:
closed by #417

PR #417, opened 2021-10-01
docs: added the quiet options in the doc

imrishabh18:
Closes - #315


imiric (approved):
Thanks for this!

We should also add it to the Spanish translation at `src/data/markdown/translated-guides/es/02 Using k6/05 Options.md`, but we can do this ourselves. FYI @ppcano

ppcano:
Thanks, @imrishabh18  💯.  Merging.

PR #419, opened 2021-10-02
Broken URL Path

krutarththakkar:
Identified some Broken links.
They were not pointing to the desired location.
Figured out the correct URL path and fixed the links.


ppcano:
Hi, these broken URLs are examples. It is fine to link to 404 resources.

PR #437, opened 2021-10-11
feat: support alternative content for JS API top level page

vi-nastya:
This PR adds ability to provide custom content for top-level `Javascript API` page.

**How it works**
1. Create  `.md` files for all section you want to add custom content for inside `src/data/markdown/docs/02 javascript api/alternative main modules`.
2. Make sure that these files have the same names and frontmatter `title` as corresponding files in `src/data/markdown/docs/02 javascript api` (e.g. `06 k6-execution.md`)
3. Set `shouldCreatePage: false` in frontmatter for all alternative files (need this to prevent generating separate pages based on these files)
4. Build the project and make sure that the alternative content is displayed in `/javascript-api` page sections.

The same approach could be used for versioned folders.

### DIFF
--- a/src/templates/docs/javascript-api.js
+++ b/src/templates/docs/javascript-api.js
@@ -34,12 +34,17 @@ const componentsForNativeReplacement = {
 
 const getContent = (nodes, alternativeNodes, sidebarTree) =>
   // eslint-disable-next-line array-callback-return,consistent-return
-  nodes.map(({ id, children: [entity] }) => {
+  nodes.map(({ id, relativeDirectory, children: [entity] }) => {
     const {
       frontmatter: { title },
       body,
     } = entity;
     if (title.replace(/\//g, '-') in sidebarTree.children) {
+      // skip found alternative content to avoid rendering it twice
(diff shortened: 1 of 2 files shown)


BolotskyDev:
LGTM 💯  It is very thoughtful of you @vi-nastya to handle versioned apis as well. The only thing to do is to fix failed builds - it is happening due to currently absent `shouldCreatePage` anywhere in our md data. I believe this can be fixed if we'll add a single `shouldCreatePage: true` to any of existing articles OR define this type [manually](https://www.gatsbyjs.com/docs/reference/graphql-data-layer/schema-customization/#creating-type-definitions). More on this you can read by clicking `details` in failed PR build check

ppcano:
@vi-nastya I have added the content on https://github.com/grafana/k6-docs/pull/437/commits/a92ff535cf60fc031c5192ed0a4186b2271ce052. 

The content of the alternative sections will be shown twice as you can see on the Table of Content. 
Screenshot 2021-10-14 at 12 59 50

The rest looks good. We can merge it after the fix. 

Initially, we won't update the versioned content. I have tested it and it works except for the issue mentioned above.

Issue #465, opened 2021-11-01
[Feature Request]: Add Documentation Template as per new Github Issue forms


Rutam21:
### Feature Description

GitHub has recently rolled out a public beta for their issue forms feature. This would allow us to create interactive issue templates and validate them.

K6 Docs currently doesn't use a documentation template format. The task is to create a GitHub issue form for documentation for this repository.

### Suggested Solution (optional)

A `documentation.yaml` file should be added to be used as the default template for raising any documentation requests for the repository.

 - [x] Prepare documentation issue form in .github/ISSUE_TEMPLATE/documentation.yaml

ppcano:
Closing it for the reasons mentioned at #467

PR #467, opened 2021-11-01
FEAT: Issue 465: Add Documentation Template as per new Github Issue Forms

Rutam21:
# Linked Issue
It fixes https://github.com/grafana/k6-docs/issues/465

# Description
This PR adds the Issue template for `Documentation` requests to use the latest Github Issue forms.

# Methodology
A respective `yaml` file has been added that would be used to provide the newest template for any issue requests.

### DIFF
--- a/.github/ISSUE_TEMPLATE/documentation.yaml
+++ b/.github/ISSUE_TEMPLATE/documentation.yaml
@@ -1,0 +1,32 @@
+name: "📚 Documentation"
+description: "Report an issue related to documentation"
+title: "📚 Documentation: "
+labels: [documentation]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to make our documentation better 🙏
+  - type: textarea
+    id: issue-description
+    validations:
+      required: true
+    attributes:
+      label: "💭 Description"
+      description: "A clear and concise description of what the issue is."
+      placeholder: "Documentation should ..."
+  - type: checkboxes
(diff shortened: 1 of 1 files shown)


Rutam21:
@na-- Please review and suggest changes, if any. Thanks.

Rutam21:
This form is just for the documentation issues/requests. This is a doc-only repo that is hosted on the web, right? So, if anyone finds any content changes or wants something new related to K6 to be documented, then they would definitely raise one and this would be standard formatting to make it look better with validations as well. I didn't add Feature/Bug templates here because of this as that would be taken care of in the K6 repo itself. We can add a config file the same way and add the K6 repo link for those here as well. Let me know if it makes sense now @na--

na--:
My point is that I am not sure what value we'll get out of a rigid issue template form when the repo is focused entirely on documentation... :man_shrugging: In the k6 repo, they somewhat make sense, since you can have multiple types of issues, but most of the benefit is lost if you don't need that, I think. Still, I don't have a strong opinion either way, @ppcano can decide if he wants this :sweat_smile:

ppcano:
Hi Rutam21, 

We have discussed this internally. We agreed with Ned's view - we haven't defined different types of issues in this repo, so we don't see a need for templates now.

We might want to include a template in the future, but this requires internal alignment with other projects, so we let #465 open for future considerations. 

I am closing PR as we are not merging it now. 

Thanks so much for your interest in helping and contributing to the k6 project.

PR #469, opened 2021-11-04
Updated default Tags

codebien:


src/data/markdown/translated-guides/en/02 Using k6/08 Tags and Groups.md
```diff
@@ -120,7 +120,7 @@ Currently, k6 automatically creates the following tags by default:
 | `tls_version` | the [TLS](/using-k6/protocols/ssl-tls) version                                                                                                                                    |
 | `scenario`    | the name of the scenario where the metric was emitted                                                                                                                             |
 | `service`     | the RPC service name for gRPC                                                                                                                                                     |
-| `rpc_type`    | one of `unary`, `server_streaming`, `client_streaming` or `bidirectional_streaming` for gRPC. *Note:*  only `unary` requests are currently supported.                             |
+| `expected_response`    | TODO |
```
codebien, line 123:
```suggestion
| `expected_response`    | If the HTTP response expected or not |
```
@MStoykov do you have a better suggestion?

MStoykov: Sorry for the slow response.
Something like
```suggestion
| `expected_response`    | "true" or "false" based on the [responseCallback](/javascript-api/k6-http/setresponsecallback-callback/) used, by default this checks that the status is 2xx or 3xx |
```
is better IMO - definitely should have a link to the setting of the callback

na-- (approved):

PR #477, opened 2021-11-09
Fixes for xk6-browser docs

robingustafsson:
PR with fixes and improvements to xk6-browser docs.


robingustafsson:
@ppcano I'll work on adding links to Playwright for all supported APIs for the classes that today just says "all APIs are supported except", to make it easier to see what APIs are supported in xk6-browser. Next step after that is adding actual docs for every supported API....but that will take some time so something we can do after today 😅

PR #485, opened 2021-11-15
 Update installation page

ppcano:
The current installation page is a little overloaded with info for edge cases. This PR does a few cosmetic changes.

## Current version

Screenshot 2021-11-15 at 12 44 51

## Proposal

Screenshot 2021-11-15 at 12 59 33

### DIFF
--- a/src/components/shared/collapsible/collapsible.module.scss
+++ b/src/components/shared/collapsible/collapsible.module.scss
@@ -33,7 +33,7 @@
   background: $color-additional-3;
   padding: 12px 50px 13px 25px;
   color: $color-primary;
-  font-size: 25px;
+  font-size: 18px;
   line-height: 35px;
   border-bottom: 1px solid transparent;
   animation: borderFadeIn 0.3s;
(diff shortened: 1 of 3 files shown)


na--:
This is probably ok, but I wonder if, instead of hiding the details, we shouldn't just create a separate install troubleshooting section with even more details? :thinking: Including the error messages that users might see if they need to install gnupg2 or install the key manually? This way we can just link to this section here, while the actual errors and their solutions will be more easily discoverable from search (Google _and_ the built-in docs Algolia search).

ppcano:
@na-- https://github.com/grafana/k6-docs/pull/485/commits/b6eca6faf9dbcc8a88f704c9b8995a4683a35c9b  see below

Screenshot 2021-11-15 at 16 13 22

Screenshot 2021-11-15 at 16 13 32

na-- (approved):
LGTM, though we might want to eventually expand Troubleshooting with the actual error messages people will see and search for.

PR #499, opened 2021-11-19
new documentation for BDD/TDD/functional testing library

sniku:
This PR deprecates the `expect` library in favor of the more capable `k6chaijs` lib based on an established framework. 

https://github.com/grafana/k6-jslib-k6chaijs


tom-miseur (commented):
Looks good! Minor thing about `should` below.

src/data/markdown/docs/05 Examples/01 Examples/19 functional testing.md
```diff
@@ -44,134 +44,1528 @@ This test goes through several steps. It creates a new user account, authenticat
...
+
+  // describe('docs example 209', () => {
+  //   should.fail();
+  //   should.fail("custom error message");
+  //   should.fail(1, 2);
+  //   should.fail(1, 2, "custom error message");
+  //   should.fail(1, 2, "custom error message", ">");
+  //   should.fail(1, 2, undefined, ">");
+
+  // });
```
tom-miseur, line 1565: The example is commented out. Is that expected?

Should we add a quick example for `should`?

The following gave the expected result:

```javascript
import http from 'k6/http';
import { describe, expect, chai } from 'https://jslib.k6.io/k6chaijs/4.3.4.0/index.js';

export default function () {
  describe('Get Public Crocodiles',   () => {
    const res = http.get('https://test-api.k6.io/public/crocodiles');
    chai.should();
    res.should.have.validJsonBody();
});
}
```

It is actually the first style listed on the [ChaiJS](https://www.chaijs.com/) website (although I'm not sure if that means it is widely used - I don't really like this style TBH).

MattDodsonEnglish:
I've made some small edits for grammar, and to make the sentences a little brisker (less passive voice, avoiding unnecessary noun forms).

Overall, I think the doc looks pretty good! I was able to follow along, and the examples seem useful.

The only other thing I have to note is more of a meta-discussion about processes. I wonder how long we should document deprecated features.

That is, when can we remove the documentation of the "expect.js" library? What if all links to those expect.js docs were directed to the new chaiJS library, with a small callout about the express deprecation? Would something bad happen?

 I suppose we can't remove it yet, as it could inconvenience many users. But I also don't think things should be kept up "for historical purposes" until the year 2100, either.

sniku:
> I've made some small edits for grammar, and to make the sentences a little brisker (less passive voice, avoiding unnecessary noun forms).
> 
> Overall, I think the doc looks pretty good! I was able to follow along, and the examples seem useful.
> 
> The only other thing I have to note is more of a meta-discussion about processes. I wonder how long we should document deprecated features.
> 
> That is, when can we remove the documentation of the "expect.js" library? What if all links to those expect.js docs were directed to the new chaiJS library, with a small callout about the express deprecation? Would something bad happen?
> 
> I suppose we can't remove it yet, as it could inconvenience many users. But I also don't think things should be kept up "for historical purposes" until the year 2100, either.

Thanks for the grammar fixes! I hope to keep working on these docs over the next 6 months, so at some point I'll likely remove the `expect` docs and add redirects. 

@ppcano do you think we can merge this?

PR #505, opened 2021-11-24
Exclude setup/teardown from iteration_duration

codebien:
Workaround for excluding  `setup/teardown` from `iteration_duration` metric.

### DIFF
--- a/src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
+++ b/src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
@@ -69,16 +69,79 @@ All the `http_req_...` lines and the ones after them are _built-in_ metrics that
 
 The following _built-in_ metrics will **always** be collected by k6:
 
-| Metric Name        | Type    | Description                                                                                                                                                                                                     |
-| ------------------ | ------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| vus                | Gauge   | Current number of active virtual users                                                                                                                                                                          |
-| vus_max            | Gauge   | Max possible number of virtual users (VU resources are pre-allocated, to ensure performance will not be affected when scaling up the load level)                                                                |
-| iterations         | Counter | The aggregate number of times the VUs in the test have executed the JS script (the `default` function).                                                                                                         |
-| iteration_duration | Trend   | The time it took to complete one full iteration of the default/main function.                                                                                                                                   |
+| Metric Name | Type | Description |
+| ----------- | ---- | ----------- |
+| vus                | Gauge   | Current number of active virtual users |
+| vus_max            | Gauge   | Max possible number of virtual users (VU resources are pre-allocated, to ensure performance will not be affected when scaling up the load level) |
+| iterations         | Counter | The aggregate number of times the VUs in the test have executed the JS script (the `default` function). |
+| iteration_duration | Trend   | The time it took to complete one full iteration. It includes the time spent in `setup` and `teardown` as well. See the [workaround](#workaround-to-calculate-iteration_duration-metric-only-for-a-scenario) in the example below, how to create a sub-metric for getting only the duration of the iteration's function for a specific scenario. |
 | dropped_iterations | Counter | Introduced in k6 v0.27.0, the number of iterations that could not be started due to lack of VUs (for the arrival-rate executors) or lack of time (due to expired maxDuration in the iteration-based executors). |
-| data_received      | Counter | The amount of received data. Read [this example](/examples/track-transmitted-data-per-url) to track data for an individual URL.                                                                                 |
-| data_sent          | Counter | The amount of data sent. Read [this example](/examples/track-transmitted-data-per-url) to track data for an individual URL.                                                                                     |
-| checks             | Rate    | The rate of successful checks.                                                                                                                                                                                  |
+| data_received      | Counter | The amount of received data. Read [this example](/examples/track-transmitted-data-per-url) to track data for an individual URL. |
+| data_sent          | Counter | The amount of data sent. Read [this example](/examples/track-transmitted-data-per-url) to track data for an individual URL. |
+| checks             | Rate    | The rate of successful checks. |
+
+<Collapsible title="Workaround to calculate iteration_duration metric only for a scenario" tag="h2">
+
+A common requested case is to track the `iteration_duration` metric without including time spent for `setup` and `teardown` functions.
+This feature is not yet available but a threshold on `iteration_duration` or any pre-existing metrics can be used as a workaround.
+
+It's based on the concept of creating thresholds for sub-metrics created by tags for the required scope and setting the criteria that always pass. It works with any enabled tags that already works with threshold, for example:
+* `iteration_duration{scenario:default}` generates a sub-metric collecting samples only for the default scenario's iteration. `scenario:default` is used because that's the internal k6 scenario name when we haven't specified `options. `scenarios` explicitly and are just using the execution shortcuts instead.
+* `iteration_duration{group:::setup}` or `iteration_duration{group:::teardown}` create sub-metrics collecting the duration only for `setup` and `teardown`. `k6` implicitly creates [groups](/using-k6/tags-and-groups#groups) for `setup` and `teardown`, and `::` is the group separator.
+* `http_req_duration{scenario:default}` can be useful as well for isolating executed long-running requests.
+
+<CodeGroup lineNumbers={[true]}>
+
+```javascript
+import { sleep } from 'k6'
+import http from 'k6/http'
+
+export const options = {
+    vus: 1,
+    iterations: 1,
+    thresholds: {
+        'iteration_duration{scenario:default}': [`max>=0`],
+        'iteration_duration{group:::setup}': [`max>=0`],
+        'iteration_duration{group:::teardown}': [`max>=0`],
+        'http_req_duration{scenario:default}': [`max>=0`],
+    },
+}
+
+export function setup() {
+    http.get('http://httpbin.test.k6.io/delay/5');
+}
+
+export default function () {
+    http.get('http://test.k6.io/?where=default');
+    sleep(0.5);
+}
+
+export function teardown() {
+    http.get('http://httpbin.test.k6.io/delay/3');
+    sleep(5);
+}
+```
+
+</CodeGroup>
+
+Dedicated sub-metrics have been generated collecting samples only for the scope defined by thresholds:
+
+<CodeGroup lineNumbers={[false]}>
+
+```
+     http_req_duration..............: avg=1.48s    min=101.95ms med=148.4ms  max=5.22s    p(90)=4.21s    p(95)=4.71s
+       { expected_response:true }...: avg=1.48s    min=101.95ms med=148.4ms  max=5.22s    p(90)=4.21s    p(95)=4.71s
+     ✓ { scenario:default }.........: avg=148.4ms  min=103.1ms  med=148.4ms  max=193.7ms  p(90)=184.64ms p(95)=189.17ms
+
+     iteration_duration.............: avg=5.51s    min=1.61s    med=6.13s    max=8.81s    p(90)=8.27s    p(95)=8.54s
+     ✓ { group:::setup }............: avg=6.13s    min=6.13s    med=6.13s    max=6.13s    p(90)=6.13s    p(95)=6.13s
+     ✓ { group:::teardown }.........: avg=8.81s    min=8.81s    med=8.81s    max=8.81s    p(90)=8.81s    p(95)=8.81s
+     ✓ { scenario:default }.........: avg=1.61s    min=1.61s    med=1.61s    max=1.61s    p(90)=1.61s    p(95)=1.61s
+```
+
+</CodeGroup>
+
+</Collapsible>
 
 ## HTTP-specific built-in metrics
 


src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
```diff
@@ -69,16 +69,80 @@ All the `http_req_...` lines and the ones after them are _built-in_ metrics that
...
-| data_sent          | Counter | The amount of data sent. Read [this example](/examples/track-transmitted-data-per-url) to track data for an individual URL.                                                                                     |
-| checks             | Rate    | The rate of successful checks.                                                                                                                                                                                  |
+| data_received      | Counter | The amount of received data. Read [this example](/examples/track-transmitted-data-per-url) to track data for an individual URL. |
+| data_sent          | Counter | The amount of data sent. Read [this example](/examples/track-transmitted-data-per-url) to track data for an individual URL. |
+| checks             | Rate    | The rate of successful checks. |
+
+<Collapsible title="Workaround to isolate `iteration_duration` metric from Setup and Teardown">
+
+A common requested case is to track the `iteration_duration` metric without including time spent for Setup and Teardown functions.
```
MStoykov, line 85:
```suggestion
A common requested case is to track the `iteration_duration` metric without including time spent for `setup` and `teardown` functions.
```

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
```diff
@@ -69,16 +69,80 @@ All the `http_req_...` lines and the ones after them are _built-in_ metrics that
...
-| iterations         | Counter | The aggregate number of times the VUs in the test have executed the JS script (the `default` function).                                                                                                         |
-| iteration_duration | Trend   | The time it took to complete one full iteration of the default/main function.                                                                                                                                   |
+| Metric Name | Type | Description |
+| ----------- | ---- | ----------- |
+| vus                | Gauge   | Current number of active virtual users |
+| vus_max            | Gauge   | Max possible number of virtual users (VU resources are pre-allocated, to ensure performance will not be affected when scaling up the load level) |
+| iterations         | Counter | The aggregate number of times the VUs in the test have executed the JS script (the `default` function). |
+| iteration_duration | Trend   | The time it took to complete one full iteration. There is an available workaround for excluding the time spent in setup/teardown functions. Check the example below. |
```
MStoykov, line 77:
```suggestion
| iteration_duration | Trend   | The time it took to complete one full iteration. This includes the time spent in `setup` and `teardown` as well, there is currently a workaround shown in the example below. |
```
seems better to me to first say that setup and teardown time is included *and* then talk about a workaround

MStoykov (approved):
LGTM, I left some nitpicks

MStoykov (approved):

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
```diff
@@ -69,16 +69,80 @@ All the `http_req_...` lines and the ones after them are _built-in_ metrics that
...
-| iterations         | Counter | The aggregate number of times the VUs in the test have executed the JS script (the `default` function).                                                                                                         |
-| iteration_duration | Trend   | The time it took to complete one full iteration of the default/main function.                                                                                                                                   |
+| Metric Name | Type | Description |
+| ----------- | ---- | ----------- |
+| vus                | Gauge   | Current number of active virtual users |
+| vus_max            | Gauge   | Max possible number of virtual users (VU resources are pre-allocated, to ensure performance will not be affected when scaling up the load level) |
+| iterations         | Counter | The aggregate number of times the VUs in the test have executed the JS script (the `default` function). |
+| iteration_duration | Trend   | The time it took to complete one full iteration. It includes the time spent in `setup` and `teardown` as well, there is currently a workaround for excluding them shown in the example below. |
```
yorugac, line 77:
```suggestion
| iteration_duration | Trend   | The time it took to complete one full iteration. It includes the time spent in `setup` and `teardown` as well. To exclude them, see the workaround in the example below. |
```

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
```diff
@@ -69,16 +69,80 @@ All the `http_req_...` lines and the ones after them are _built-in_ metrics that
...
+import { sleep } from 'k6'
+import http from 'k6/http'
+
+export const options = {
+    vus: 1,
+    iterations: 1,
+    thresholds: {
+        // We're using 'scenario:default' because that's the internal k6
+        // scenario name when we haven't actually specified `options.scenarios`
+        // explicitly and are just using the old execution shortcuts instead...
```
yorugac, line 100:
```suggestion
        // We're using 'scenario:default' because that's the default name of scenario
        // unless `options.scenarios` is specified.
```
?

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
```diff
@@ -69,16 +69,80 @@ All the `http_req_...` lines and the ones after them are _built-in_ metrics that
...
+        'iteration_duration{scenario:default}': [
+            // Some dummy threshold that's always going to pass.
+	    // You don't even need to have something here.
```
yorugac, line 103: Sorry, I don't follow here: what does this mean `don't need to have something here`? :sweat_smile: 
I.e. if there is a simpler way to make a dummy threshold, it's better to just add it in the example.

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
```diff
@@ -69,16 +69,80 @@ All the `http_req_...` lines and the ones after them are _built-in_ metrics that
...
+            `max>=0`,
+        ],
+        // You can isolate setup and teardown metrics, if you are interested in them
+        // for some reason. This looks nasty, but '::' is the group separator.
```
yorugac, line 107:
```suggestion
        // You can remove the time spent only on `setup` or only on `teardown` function too.
        // This looks nasty, but '::' is the group separator.
```

\+ a link to [group docs](https://k6.io/docs/using-k6/tags-and-groups/#groups) here?

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
```diff
@@ -69,16 +69,80 @@ All the `http_req_...` lines and the ones after them are _built-in_ metrics that
...
+        'iteration_duration{group:::setup}': [`max>=0`],
+        'iteration_duration{group:::teardown}': [`max>=0`],
+
+        // And you can also isolate the HTTP metrics, in case of long-running requests
```
yorugac, line 111:
```suggestion
        // And you can also remove the time spent on the HTTP metrics, in case of long-running requests
```
Correct?

I think "isolated" is misleading as this workaround doesn't actually isolate anything: it just allows one to get additional values of duration with specific conditions.

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
```diff
@@ -69,16 +69,80 @@ All the `http_req_...` lines and the ones after them are _built-in_ metrics that
...
+
+export function teardown() {
+    http.get('http://httpbin.test.k6.io/delay/3');
+    sleep(5);
+}
+```
+
+</CodeGroup>
+
+A dedicated metric for the `default` scenario is generated excluding the setup and teardown function:
```
yorugac, line 132:
```suggestion
As a result, a dedicated metric for the `default` scenario is generated excluding the time spent on `setup` and `teardown` functions:
```

Also, shouldn't there be a `http_req_duration` in this output too?

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
```diff
@@ -69,16 +69,80 @@ All the `http_req_...` lines and the ones after them are _built-in_ metrics that
...
-| data_sent          | Counter | The amount of data sent. Read [this example](/examples/track-transmitted-data-per-url) to track data for an individual URL.                                                                                     |
-| checks             | Rate    | The rate of successful checks.                                                                                                                                                                                  |
+| data_received      | Counter | The amount of received data. Read [this example](/examples/track-transmitted-data-per-url) to track data for an individual URL. |
+| data_sent          | Counter | The amount of data sent. Read [this example](/examples/track-transmitted-data-per-url) to track data for an individual URL. |
+| checks             | Rate    | The rate of successful checks. |
+
+<Collapsible title="Workaround to isolate `iteration_duration` metric from Setup and Teardown">
```
yorugac, line 83:
```suggestion
<Collapsible title="Workaround to calculate `iteration_duration` metric without the time spent on `setup` and `teardown`">
```

yorugac (changes requested):
It's clear to me but I think some formulations could be improved since it's the docs. @codebien, please see the comments

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
MStoykov:
```suggestion
        // We're using 'scenario:default' because that's the internal k6
        // scenario name when we haven't actually specified `options.scenarios`
        // explicitly and `default` should be replaced with the appropriate name if `options.scenarios` is used
```

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
MStoykov: I kind of feel `isolate` is better, as it doesn't remove any metrics it just looks it just aggregates only metrics that match. 

Maybe : 
```suggestion
        // And you can do this for all metrics, for example including HTTP metrics
```
and then have a more concrete explanation above?

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
MStoykov: this doesn't *remove* the time spent ... it makes it so the thresholds is calculated only on the provided values, so if anything it "removes" all other values for this metrics :shrug: I od think "isolate" is good enough, but we can have a better explanation *in one place*

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
MStoykov: I would argue this is ... wrongly worded

`itearation_duration{scenario:default}` *does* not actually *exclude* `setup` and `teardown` times. It looks like it, but what it actually does is make a new *submetric* (w/e that means) that is all values of `iteration_duraiton` that have a tag `scanario` with value `default` - theoretically at least (especially with `k6/execution`) you can have metrics with that tag and value that are emitted in `setup` and also you can have another 20 scenarios with different names and this will "exclude" all of them

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
MStoykov:
```suggestion
<Collapsible title="Workaround to calculate `iteration_duration` metric only for a scenario">
```

MStoykov (commented):
I agree with @yorugac that this probably needs better language, but I would argue that the proposed move away from "isolate" to "remove" is the wrong direction :) 

As I am mentioning there `metric_name{tag_name:value}` just makes another (sub)"metric" that is all the values of `metric_name` for which `tag_name` has the value `value`. In the particular example it just so happens that we are talking about `iteration_duration` and 1 "default" function with 1 scenario and some setup/teardown. But if they were multiple scenarios it wouldn't just "remove" the setup/teardown metrics but all of them. So "isolate" seems better but it might need to be worded as "isolate the metric with only the specified tag value" :shrug:

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
codebien:
```suggestion
| iteration_duration | Trend   | The time it took to complete one full iteration. It includes the time spent in `setup` and `teardown` as well. See the workaround in the example below, how to create a sub-metric for getting only the duration of the iteration's function for a specific scenario.  |
```
Could be better in this way?

codebien:
I re-arranged following the previous suggestions, WDYT?

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
```diff
@@ -69,16 +69,78 @@ All the `http_req_...` lines and the ones after them are _built-in_ metrics that
...
-| iteration_duration | Trend   | The time it took to complete one full iteration of the default/main function.                                                                                                                                   |
+| Metric Name | Type | Description |
+| ----------- | ---- | ----------- |
+| vus                | Gauge   | Current number of active virtual users |
+| vus_max            | Gauge   | Max possible number of virtual users (VU resources are pre-allocated, to ensure performance will not be affected when scaling up the load level) |
+| iterations         | Counter | The aggregate number of times the VUs in the test have executed the JS script (the `default` function). |
+| iteration_duration | Trend   | The time it took to complete one full iteration. It includes the time spent in `setup` and `teardown` as well. See the workaround in the example below, how to create a sub-metric for getting only the duration of the iteration's function for a specific scenario. |
```
yorugac, line 77: Looking at other fields in the table, there should be an internal link here, right?

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
```diff
@@ -69,16 +69,78 @@ All the `http_req_...` lines and the ones after them are _built-in_ metrics that
...
+| data_received      | Counter | The amount of received data. Read [this example](/examples/track-transmitted-data-per-url) to track data for an individual URL. |
+| data_sent          | Counter | The amount of data sent. Read [this example](/examples/track-transmitted-data-per-url) to track data for an individual URL. |
+| checks             | Rate    | The rate of successful checks. |
+
+<Collapsible title="Workaround to calculate `iteration_duration` metric only for a scenario">
+
+A common requested case is to track the `iteration_duration` metric without including time spent for `setup` and `teardown` functions.
+This feature is not yet available but a threshold on `iteration_duration` or any pre-existing metrics can be used as a workaround.
+
+It's based on the concept of creating sub-metrics for the required scope and set always-pass thresholds. It mostly works with every filter that already works with Threshold, for example:
```
yorugac, line 88:
```suggestion
It's based on the concept of creating sub-metrics for the required scope and setting thresholds that always pass. It works with any filter that already works with threshold, for example:
```
Also, not fully sure about all docs but I think at least [thresholds page](https://k6.io/docs/using-k6/thresholds/) always uses just "metric" and "submetric" for all cases, not "filter" as here ^^

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
```diff
@@ -69,16 +69,78 @@ All the `http_req_...` lines and the ones after them are _built-in_ metrics that
...
+* 'iteration_duration{scenario:default}' will generate a sub-metric collecting samples only for the default scenario's iteration. 'scenario:default' because that's the internal k6 scenario name when we haven't actually specified `options.scenarios` explicitly and are just using the execution shortcuts instead.
```
yorugac, line 89:
```suggestion
* 'iteration_duration{scenario:default}' generates a sub-metric collecting samples only for the default scenario's iteration. 'scenario:default' is used because that's the internal k6 scenario name when we haven't specified `options.scenarios` explicitly and are just using the execution shortcuts instead.
```

yorugac (approved):
Sorry I got to this PR only now.
I think it's much more clear now! I added a few grammar/typos comments that I noticed on the way, but the overall wording is definitely easier to follow :slightly_smiling_face:

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
```diff
@@ -69,16 +69,79 @@ All the `http_req_...` lines and the ones after them are _built-in_ metrics that
...
-| data_sent          | Counter | The amount of data sent. Read [this example](/examples/track-transmitted-data-per-url) to track data for an individual URL.                                                                                     |
-| checks             | Rate    | The rate of successful checks.                                                                                                                                                                                  |
+| data_received      | Counter | The amount of received data. Read [this example](/examples/track-transmitted-data-per-url) to track data for an individual URL. |
+| data_sent          | Counter | The amount of data sent. Read [this example](/examples/track-transmitted-data-per-url) to track data for an individual URL. |
+| checks             | Rate    | The rate of successful checks. |
+
+<Collapsible title="Workaround to calculate iteration_duration metric only for a scenario" tag="h2">
+
+A common requested case is to track the `iteration_duration` metric without including time spent for `setup` and `teardown` functions.
```
yorugac, line 85:
```suggestion
A common requested case is to track the `iteration_duration` metric without including time spent on `setup` and `teardown` functions.
```

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
```diff
@@ -69,16 +69,79 @@ All the `http_req_...` lines and the ones after them are _built-in_ metrics that
...
+This feature is not yet available but a threshold on `iteration_duration` or any pre-existing metrics can be used as a workaround.
+
+It's based on the concept of creating thresholds for sub-metrics created by tags for the required scope and setting the criteria that always pass. It works with any enabled tags that already works with threshold, for example:
```
yorugac, line 88:
```suggestion
It's based on the concept of creating thresholds for sub-metrics created by tags for the required scope and setting the criteria that always pass. It works with any enabled tags that can be used in a threshold, for example:
```
To avoid double usage of "work" in the same sentence ^^

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
```diff
@@ -69,16 +69,79 @@ All the `http_req_...` lines and the ones after them are _built-in_ metrics that
...
+
+<Collapsible title="Workaround to calculate iteration_duration metric only for a scenario" tag="h2">
+
+A common requested case is to track the `iteration_duration` metric without including time spent on `setup` and `teardown` functions.
+This feature is not yet available but a threshold on `iteration_duration` or any pre-existing metrics can be used as a workaround.
+
+It's based on the concept of creating thresholds for sub-metrics created by tags for the required scope and setting the criteria that always pass. It works with any enabled tags that can be used in a threshold, for example:
+* `iteration_duration{scenario:default}` generates a sub-metric collecting samples only for the default scenario's iteration. `scenario:default` is used because that's the internal k6 scenario name when we haven't specified `options. `scenarios` explicitly and are just using the execution shortcuts instead.
+* `iteration_duration{group:::setup}` or `iteration_duration{group:::teardown}` create sub-metrics collecting the duration only for `setup` and `teardown`. `::` is the group separator syntax and k6 implicitly creates groups for `setup` and `teardown`.
```
MStoykov, line 90:
```suggestion
* `iteration_duration{group:::setup}` or `iteration_duration{group:::teardown}` create sub-metrics collecting the duration only for `setup` and `teardown`.  K6 implicitly creates [groups](/using-k6/tags-and-groups#groups) for `setup` and `teardown`, and `::` is the group separator.
```

MStoykov (approved):
LGTM, I had a final nitpick but we can leave it as is as well

PR #511, opened 2021-11-29
Update structure of the Extension guide + minor content changes

ppcano:
Screenshot 2021-11-29 at 16 38 40


imiric (approved):
LGTM! I like the split into separate pages, it's much easier to read that way.

PR #514, opened 2021-11-30
Fix for ambiguous example on thresholds page

yorugac:
There is a misleading example on "Thresholds" page, in ["Threshold Syntax"](https://k6.io/docs/using-k6/thresholds/#threshold-syntax) section that makes it seem as if it's fully correct to define 2 thresholds for the same metric. While it is possible, it should be avoided as k6 reads thresholds into Golang map and thresholds with the same key (metric name) will be overwritten. So this PR is meant to correct this info by:
- removing long format from "Threshold Syntax" (currently, it doesn't make sense there as the description for fields in long format is in one of the next sections, ["Aborting a test when a threshold is crossed"](https://k6.io/docs/using-k6/thresholds/#aborting-a-test-when-a-threshold-is-crossed)) and leaving that example with just one metric - one threshold
- adding a warning about duplicate definitions in the end of "Threshold Syntax"

### DIFF
--- a/src/data/markdown/translated-guides/en/02 Using k6/04 Thresholds.md
+++ b/src/data/markdown/translated-guides/en/02 Using k6/04 Thresholds.md
@@ -173,30 +173,30 @@ You can find more specific threshold examples on the [Counter](/javascript-api/k
 
 ## Threshold Syntax
 
-Thresholds can be specified in a short or full format.
+A threshold requires at least one `threshold_expression` to be defined:
 
 <CodeGroup labels={["threshold-options.js"]} lineNumbers={[true]}>
 
 ```javascript
 export const options = {
   thresholds: {
-    metric_name1: ['threshold_expression' /* ...*/], // short format
-    metric_name1: [
+    metric_name1: [ 'threshold_expression', ... ], // short format
+    metric_name2: [
       {
         threshold: 'threshold_expression',
         abortOnFail: true, // boolean
         delayAbortEval: '10s', // string
       },
     ], // full format
-  },
+  }
 };
 ```
 
 </CodeGroup>
 
-The above declaration inside a k6 script means that there will be a threshold configured for
-the metric `metric_name1`. To determine if the threshold has failed or passed, the string `'threshold_expression'`
-will be evaluated. The `'threshold_expression'` must follow the following format.
+The above declaration inside a k6 script means that there will be thresholds configured for
+the metrics `metric_name1` and `metric_name2`. To determine if the threshold has failed or passed, the string `'threshold_expression'`
+will be evaluated. The `'threshold_expression'` must follow the following format:
 
 `aggregation_method operator value`
 
@@ -270,6 +270,23 @@ We have these thresholds:
   this metric is that the returned content must be smaller than 4000 bytes.
 - A counter metric that keeps track of the total number of times content returned was **not** OK.
   The success criteria here implies that content can't have been bad more than 99 times.
+
+**⚠️ Common mistake** Do not specify multiple thresholds for the same metric by repeating the same object key:
+
+<CodeGroup labels={["threshold-duplicate-mistake.js"]} lineNumbers={[true]}>
+
+```javascript
+export let options = {
+  thresholds: {
+    // avoid using the same metric more than once here
+    // metric_name: [ 'count<100' ],
+    // metric_name: [ 'rate<50' ],
+  }
+};
+```
+</CodeGroup>
+
+Since thresholds are defined as the properties of a JavaScript object, it's not possible to specify multiple ones with the same property name. Only the last one will remain and the rest will be **silently** ignored. You should specify them with an [array for the same key instead](/using-k6/thresholds/#multiple-thresholds-on-a-single-metric).
 
 ## Thresholds on tags
 


src/data/markdown/translated-guides/en/02 Using k6/04 Thresholds.md
```diff
@@ -173,30 +173,23 @@ You can find more specific threshold examples on the [Counter](/javascript-api/k
...
-    metric_name1: [
-      {
-        threshold: 'threshold_expression',
-        abortOnFail: true, // boolean
-        delayAbortEval: '10s', // string
-      },
-    ], // full format
-  },
+    metric_name: [ 'threshold_expression', ... ], // short format
+  }
```
na--, line 184: Hmm, I'd leave the example with the full format here, just fix it so it has `metric_name1` and `metric_name2`. It doesn't seem like we explain it anywhere else in the page...

src/data/markdown/translated-guides/en/02 Using k6/04 Thresholds.md
```diff
@@ -271,6 +264,23 @@ We have these thresholds:
 - A counter metric that keeps track of the total number of times content returned was **not** OK.
   The success criteria here implies that content can't have been bad more than 99 times.
 
+**⚠️ Bad practice.** Even though it is possible to define a threshold for the same metric more than once like this:
```
na--, line 267: This is not really a "bad practice", this is more of a "common mistake".

na--: And this sentence also doesn't seem right:
> Even though it is possible to define a threshold for the same metric more than once like this

It's not technically correct and I think we can use programming terms here, something like
> Since thresholds are defined as an object, it's not possible to specify different thresholds with the same key (metric name) like this:
> ... <example> ...
> Instead, specify them in the array as shown in https://k6.io/docs/using-k6/thresholds/#multiple-thresholds-on-a-single-metric

ppcano: General rule?

ppcano (approved):

src/data/markdown/translated-guides/en/02 Using k6/04 Thresholds.md
na--: general rule is even worse :sweat_smile: it's pretty much always a mistake

yorugac: @na-- not sure about this:

> Since thresholds are defined as an object, it's not possible to specify different thresholds with the same key (metric name) like this:

I meant that it is possible in the sense that k6 doesn't throw an error on it. And that's why one should pay attention in this case, to avoid hard to catch mistake. I.e. if I were to read "it's not possible to specify", I'd think that it'll be caught at the start and I don't need to worry about it but that's not the case.
Hm, maybe then:
> Even though it is possible to define a threshold for the same metric more than once in the script like this
> ...
> This is not possible in k6 engine since thresholds are defined as an object with unique keys (metric names). Instead, specify them in the array as shown in https://k6.io/docs/using-k6/thresholds/#multiple-thresholds-on-a-single-metric

?

src/data/markdown/translated-guides/en/02 Using k6/04 Thresholds.md
yorugac: Yes, I wondered about this part too.. But the full format is explained in one of the next sections: https://k6.io/docs/using-k6/thresholds/#aborting-a-test-when-a-threshold-is-crossed
I'm fine putting it back with another name, as you suggested too, I just find it a bit strange that description is in another place.

na--: Doesn't hurt to have the example with both formats on the top :man_shrugging:

src/data/markdown/translated-guides/en/02 Using k6/04 Thresholds.md
na--: "though it is possible to define a threshold for the same metric more than once in the script like this" is false - it is not possible to do that in vanilla JavaScript. Yes, it's not a syntax error, but that doesn't mean it's possible. It's not a k6 thing, it's how JS works.

yorugac: Okay, I get your point about complete correctness; so what is your version then? If we take into account that it's best not to create a wrong impression for a reader :slightly_smiling_face:

src/data/markdown/translated-guides/en/02 Using k6/04 Thresholds.md
```diff
@@ -271,6 +264,23 @@ We have these thresholds:
...
+    metric_name: [ 'count<100' ],
+    metric_name: [ 'rate<50' ],
+  }
+};
+```
+
+</CodeGroup>
+
+Such definitions must be avoided so that each metric's threshold is defined only once. Otherwise, only one of the threshold definitions will be executed and there is no guarantee which one it would be. Instead, if the given metric needs more than one threshold expression, put all threshold expressions as part of one array as is shown for `RTT` metric in the previous example.
+
```
na--, line 283:
```suggestion
> **⚠️ Common mistake**: Do not specify multiple thresholds for the same metric by repeating the same object key. Because thresholds are defined as a the properties of a JavaScript object, it's not possible to specify multiple ones with the same property name. Only the last one will remain and the rest will be **silently** ignored. You can specify them with an [array for the same key instead](/using-k6/thresholds/#multiple-thresholds-on-a-single-metric).
```

src/data/markdown/translated-guides/en/02 Using k6/04 Thresholds.md
na--: Here it is https://github.com/grafana/k6-docs/pull/514/files#r759222258

I think we shouldn't even give an example, just in case someone copies that without actually reading the text...

na-- (approved):

Issue #52, opened 2020-07-01
Improve installation instructions, maybe add more details about unofficial packages


simskij:
Like asdf, aur, appimage, snap among others.

Should also include a disclaimer about these methods not being official

simskij:
Same as mentioned in #51. Closing that one as this contains (somewhat) more details. Thanks @MStoykov ! 🙏

na--:
https://repology.org/project/k6/versions

We can mention some of these packages - they're unofficial for us, in that we don't build them ourselves, but they may be official to some of the distributions (i.e. they are officially packaged in their main distribution channels)... So, I think we should distinguish between at least 3 tiers of trust:
1. Managed by us
2. First-class packages in a distribution 
3. Just a random package built by a random person on the internet

The distinction between 2 and 3 for me is whether someone from the distribution has to review the package before it can be updated, or if a single random person on the internet can release a new bogus k6 version with malware via that channel... :smile: Homebrew and nixpkgs probably fall in the second category, while AUR is in the third, I think...

MStoykov:
- the [asdf is one](https://kandi.openweaver.com/shell/grimoh/asdf-k6) year old
- aur is [up to date](https://aur.archlinux.org/packages/k6)
- can't find an appimage
- [snap](https://snapcraft.io/k6) is 3 years old.

Given that I kind of feel like there is barely any reason to make this page. 

I would argue Arch Linux users are aware AUR exists and know how to check it and have only them seems ... strange

So I am for closing this and *not* do anything. So for the purposes on not waiting on this I *am* closing this issue and if someone disagrees they can reopen it.

MStoykov:
The [chocolatey one](https://community.chocolatey.org/packages/k6) seems up to date and has had a [PR](https://github.com/grafana/k6/pull/1027) to be added to the README ~forever~ three years.

na--:
I somewhat disagree here. Maybe we don't need a separate page, but I don't think the [current page with installation instructions](https://k6.io/docs/getting-started/installation/) is very good. So I'll reopen this with a slightly changed title.

MStoykov:
I would prefer if you open a new issue and propose the changes you want or at least describe what you don't like from the current one.

Now this issue has a title and first comment that don't go all that well together

MattDodsonEnglish:
The installation page has instructions for the choco package install now.  Same one documented in the linked PR: https://github.com/grafana/k6/pull/1027/files

Closing this issue.

PR #521, opened 2021-12-06
Include Winget instructions to Update 02 installation.md

Alexandre-Nourissier:
Include the Windows Package Manager instruction to install K6.

This helps clarify which package to install, as a search with "k6" returns multiple results, and could possibly lead to confusion in the future. 

image

### DIFF
--- a/src/data/markdown/translated-guides/en/01 Getting started/02 Installation.md
+++ b/src/data/markdown/translated-guides/en/01 Getting started/02 Installation.md
@@ -42,6 +42,12 @@ If you use the [Chocolatey package manager](https://chocolatey.org/) you can ins
 choco install k6
 ```
 
+or if you use the [Windows Package Manager](https://github.com/microsoft/winget-cli) you can install from the unofficial [k6 manifests](https://github.com/vedantmgoyal2009/winget-pkgs/tree/master/manifests/k/k6/k6) with:  
+
+```
+winget install k6
+```
+
 Otherwise you can manually download and install the [latest official `.msi` package](https://dl.k6.io/msi/k6-latest-amd64.msi).
 
 ## Docker
--- a/src/data/markdown/translated-guides/en/01 Getting started/02 Installation.md
+++ b/src/data/markdown/translated-guides/en/01 Getting started/02 Installation.md
@@ -42,7 +42,7 @@ If you use the [Chocolatey package manager](https://chocolatey.org/) you can ins
 choco install k6
 ```
 
-or if you use the [Windows Package Manager](https://github.com/microsoft/winget-cli) you can install from the unofficial [k6 manifests](https://github.com/vedantmgoyal2009/winget-pkgs/tree/master/manifests/k/k6/k6) with:  
+or if you use the [Windows Package Manager](https://github.com/microsoft/winget-cli) you can install the official packages from the k6 manifests [(created by the community)](https://github.com/microsoft/winget-pkgs/tree/master/manifests/k/k6/k6) with:  
 
 ```
 winget install k6


ppcano:
@Alexandre-Nourissier, thanks for your contribution and support.

There is an issue discussing how to document unofficial packages #52. Given other possible implications, it is better if the k6 core team reviews and decides how to provide unofficial installation instructions.

cc @na--

Alexandre-Nourissier:
> @Alexandre-Nourissier, thanks for your contribution and support.
> 
> There is an issue discussing how to document unofficial packages #52. Given other possible implications, it is better if the k6 core team reviews and decides how to provide unofficial installation instructions.
> 
> cc @na--

This would be ideal. 

I do notice that this wasn't a priority for the k6 team as the issue was abandoned 15 months ago. Bear in mind that malicious or benevolent 3rd parties will not wait for your team to act.  Even if k6 doesn't publicly communicate about a package or doesn't consider itself the owner of this package, if a convenient way to install k6 is available through a distribution channel, people will want to use it.

Sometimes, just like for Winget, it's easy to track down the actual source of the package, but sometimes, it's also pretty hard to have the package's exclusive ownership.

In any case, thank you and your team for this great tool!

na--:
There seems to be an official and up-to-date winget package: https://github.com/microsoft/winget-pkgs/blob/master/manifests/k/k6/k6/0.35.0/k6.k6.installer.yaml

Even though we don't maintain that, it will be fine to include that in our docs, since it's an "official" package in the main winget repo. It has some controls and review on how it's updated, similarly to the homebrew package. 

The weird thing though is that the latest version shown in https://winget.run/pkg/k6/k6 is k6 v0.31.1 :confused:

Alexandre-Nourissier:
I wasn't aware of this website but I assume that the data aggregation routine is offline. I assume that people just use the CLI search feature, or check the GitHub repository directly. But my opinion doesn't matter, it's up to the team to have the ownership. 

That said, if I had to advocate against the WinGet distribution channel, I would say that it can be hard to retain exclusive ownership of the package. At least at some point in history, (don't know if this is still the case) anyone could push a new update that would target a non-approved executable. There is protection against the targeted executable being replaced (hash), but a new version doesn't have to be hosted on the same domain as the previous one.

na-- (changes requested):
> I wasn't aware of this website but I assume that the data aggregation routine is offline. 

Yeah, that's probably likely...

> That said, if I had to advocate against the WinGet distribution channel, I would say that it can be hard to retain exclusive ownership of the package. At least at some point in history, (don't know if this is still the case) anyone could push a new update that would target a non-approved executable. There is protection against the targeted executable being replaced (hash), but a new version doesn't have to be hosted on the same domain as the previous one.

That's fair, but again, we don't want to have an "exclusive ownership of the package". We don't have that for the [homebrew k6 package](https://github.com/Homebrew/homebrew-core/blob/master/Formula/k6.rb) and we still recommend it in the docs. And if some linux distro starts having an official k6 package tomorrow, we'll mention that as well...

We simply want these package managers to have some review process in place, so it's fairly difficult for someone to maliciously replace the k6 package with their own custom build. From what I can see, the official `microsoft/winget-pkgs` repo has this: https://github.com/microsoft/winget-pkgs/pulls?q=is%3Apr+is%3Aclosed

So, if you replace the `vedantmgoyal2009` repo with it and reword the sentence (it's not "unofficial"), I'd be fine with merging this PR.

Alexandre-Nourissier:
@na--  Good catch! Well, that was dangerous... Thanks for keeping up with my shenanigans 🥇. I made the change to include the missing information.

na-- (approved):
Thanks!

ppcano:
Thanks, @Alexandre-Nourissier.

PR #546, opened 2022-01-13
Remove repeated word in summary documentation

federicotdn:
Remove a repeated word.


ppcano:
Muchas gracias, @federicotdn!

PR #556, opened 2022-01-18
Document v0.36 open()'s breaking change

olegbespalov:
Document a breaking change of the `open()` function

### DIFF
--- a/src/data/markdown/docs/02 javascript api/01 Init context/open.md
+++ b/src/data/markdown/docs/02 javascript api/01 Init context/open.md
@@ -23,7 +23,7 @@ See the example further down on this page. For a more in-depth description, see 
 
 #### Breaking change in v0.36.0
 
-Since k6 v0.36.0 we also restricted usage of the `open(...)` under the `__VU` condition. A code like `if (__VU >0) { const arr = open("./arr.json"); }` should cause an error.
+Since k6 v0.36.0, VUs are now restricted to only `open()` files that were also opened in the [init context](https://k6.io/docs/using-k6/test-life-cycle/#init-and-vu-stages) of the first VU - the one that was initialized to get the exported `options` from the JS script (`__VU==0`). This means that the code like `if (__VU > 0) { const arr = open("./arr.json"); }` will result in an error.
 
 </blockquote>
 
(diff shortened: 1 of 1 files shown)


src/data/markdown/docs/02 javascript api/01 Init context/open.md
```diff
@@ -21,6 +21,10 @@ By restricting it to the init context, we can easily determine what local files
 
 See the example further down on this page. For a more in-depth description, see [Running k6](/getting-started/running-k6).
 
+#### Breaking change in v0.36.0
+
+Since k6 v0.36.0 we also restricted usage of the `open(...)` under the `__VU` condition. A code like `if (__VU >0) { const arr = open("./arr.json"); }` should cause an error.
```
yorugac, line 26:
```suggestion
Since k6 v0.36.0 `open(...)` is now restricted to the init context of the VU that's initialized to get the options from the script (`__VU==0`). This means that the code like `if (__VU >0) { const arr = open("./arr.json"); }` will result in error.
```
Trying to improve formulation with rather shameless copy-pasting from https://github.com/grafana/k6/pull/2340 :sweat_smile: But if formulation was done in a satisfying manner there, I think we can re-use it here.

yorugac (changes requested):
Formulation changes: please see the comment.

A related suggestion: perhaps it makes sense to add "good practice hints"? E.g. it's best if you open files like this: a), b), ...
Unless "best practices" section should be in another place and only linked here. WDYT?

src/data/markdown/docs/02 javascript api/01 Init context/open.md
na--:
```suggestion
Since k6 v0.36.0, VUs are now restricted to only `open()` files that were also opened in the init context of the first VU - the one that was initialized to get the exported `options` from the JS script (`__VU==0`). This means that the code like `if (__VU > 0) { const arr = open("./arr.json"); }` will result in an error.
```

olegbespalov:
>A related suggestion: perhaps it makes sense to add "good practice hints"? E.g. it's best if you open files like this: a), b), ...
Unless "best practices" section should be in another place and only linked here. WDYT?

I think it's a good idea, but on the other see that the opening of the files is a hot topic, so maybe it makes sense to add this section of the documentation once we make the improvements (for example, once there is more clarity with [that issue](https://github.com/grafana/k6/issues/2311))

yorugac:
@olegbespalov yes, agreed on returning to this after https://github.com/grafana/k6/issues/2311 :+1:

MStoykov (approved):

PR #559, opened 2022-01-20
JS API for execution.test.abort

codebien:
Docs for `test.abort()` new feature

### DIFF
--- a/src/data/markdown/docs/02 javascript api/06 k6-execution.md
+++ b/src/data/markdown/docs/02 javascript api/06 k6-execution.md
@@ -9,6 +9,7 @@ The `k6/execution` module provides the test execution information with the follo
 
 - [instance](#instance)
 - [scenario](#scenario)
+- [test](#test)
 - [vu](#vu)
 
 <div class="code-group" data-props='{"labels": [], "lineNumbers": [true]}'>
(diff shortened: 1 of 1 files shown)


src/data/markdown/docs/02 javascript api/06 k6-execution.md
```diff
@@ -64,6 +65,14 @@ Meta information and execution details about the current running [scenario](/usi
 | iterationInTest     | integer | The unique and zero-based sequential number of the current iteration in the scenario. It is unique in all k6 execution modes - in local, cloud and distributed/segmented test runs. However, while every instance will get non-overlapping index values in cloud/distributed tests, they might iterate over them at different speeds, so the values won't be sequential across them. |
 
 
+### test
+
+Control the test execution.
+
+| Field | Type | Description |
+|-------|------|-------------|
+| abort | | It immediately aborts a k6 load test run with the exit code `108`. Aborting a test is fully supported in `k6 cloud`, preserving the same user experience for both local and cloud users. |
```
imiric, line 74: I'm wondering if "field" is the right column name... "Property" would be more accurate. Same goes for the other sections. It's unrelated to this PR, but we might as well change it here if you agree.

And perhaps the value of Type here could be "function"?

imiric (changes requested):

imiric (approved):
Great! Thanks a lot for picking this up! :smile:

oleiade (approved):
Nickel Chrome 👍🏻 Cheers for that 🦖

na-- (approved):

ppcano:
@codebien @na-- 

Because the default behavior of k6 is `same user experience for both local and cloud users`,  I suggest removing:

`Aborting a test is fully supported in k6 cloud, preserving the same user experience for both local and cloud users.`

And, only provide the details on the exceptions.

PR #561, opened 2022-01-21
Specify that scenarios run in parallel by default

MStoykov:


oleiade (approved):
👍🏻 🦖 🌮

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios.md
```diff
@@ -18,7 +18,7 @@ Scenarios allow us to make in-depth configurations to how VUs and iterations are
 - Every scenario can use a distinct VU and iteration scheduling pattern,
   powered by a purpose-built [executor](#executors). This enables modeling
   of advanced execution patterns which can better simulate real-world traffic.
-- They can be configured to run in sequence or parallel, or in any mix of the two.
+- They run independently, which means that they also by default run in parallel. But can be made somewhat sequential through setting `startTime` accordingly.
```
na--, line 21:
```suggestion
- They are independent from each other and run in parallel, though can be made to appear sequential through the `startTime` property.
```

MStoykov:
This new wording kind of make it sound like you need to set `startTime` on *all* scenarios, but you definitely can skip it on at least one, but it's definitely better than before so :shrug:

PR #564, opened 2022-01-24
abort: Removed obvious sentence

codebien:


src/data/markdown/docs/02 javascript api/06 k6-execution.md
```diff
@@ -71,7 +71,7 @@ Control the test execution.
 
 | Property | Type | Description |
 |----------|------|-------------|
-| abort | function | It immediately aborts a k6 load test run with the exit code `108`. Aborting a test is fully supported in `k6 cloud`, preserving the same user experience for both local and cloud users. |
+| abort | function | It immediately aborts a k6 load test run with the exit code `108`. |
```
codebien, line 74:
```suggestion
| abort | function | It ~~immediately~~ aborts a k6 load test run with the exit code `108`. |
```
@imiric Is it immediately the right word? Can be considered immediately also for the cloud?

ppcano:
@codebien When reading the docs, I also wondered how the input string affects the output. 

The `exit` code is the same but prints the abort message on the console? Not sure if it is necessary but in case it is relevant, we could explain more about this.

src/data/markdown/docs/02 javascript api/06 k6-execution.md
imiric: It's not quite "immediate" in the cloud as the processing winds down, but it might be worth mentioning here that `teardown()` will still run if `abort()` is called. So in that sense, I would remove "immediately" from here.

codebien, line 74:
```suggestion
| abort([String]) | function | It aborts a `k6` load test run with the exit code `108`, and optionally, a String can be passed for providing a more detailed error's message. `k6` executes the `teardown()` function if it has been defined before stopping the execution. |
```
WDYT?

ppcano: Suggestion:

It aborts the test run with the exit code `108`, and an optional string parameter can provide an error message. Aborting the test will not prevent the [`teardown()`]() execution.

PR #574, opened 2022-02-10
chore: stress the importance of opening inside the SharedArray

olegbespalov:
Stress the importance that any opening of big files should happen inside the SharedArray's callback

### DIFF
--- a/src/data/markdown/docs/02 javascript api/04 k6-data/1-SharedArray.md
+++ b/src/data/markdown/docs/02 javascript api/04 k6-data/1-SharedArray.md
@@ -28,8 +28,8 @@ Which means that for the most part if you currently have an array data structure
 import { SharedArray } from 'k6/data';
 
 const data = new SharedArray('some name', function () {
-  // here you can generate the data dynamically or open files and do additional processing
-  // it's important to open large files inside this function to avoid memory saturation
+  // All heavy work (opening and processing big files for example) should be done inside here.
+  // This way it will happen only once and the result will be shared between all VUs, saving time and memory.
   const f = JSON.parse(open('./somefile.json'));
   return f; // f must be an array
 });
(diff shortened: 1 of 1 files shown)


src/data/markdown/docs/02 javascript api/04 k6-data/1-SharedArray.md
```diff
@@ -28,7 +28,7 @@ Which means that for the most part if you currently have an array data structure
 import { SharedArray } from 'k6/data';
 
 const data = new SharedArray('some name', function () {
-  // here you can open files, and then do additional processing on them or just generate the data dynamically
+  // it's important to open big files (generate a big amount of data) inside this function
```
ppcano, line 31:
```suggestion
  // here you can generate the data dynamically or open files and do additional processing
  // it's important to open large files inside this function to avoid memory saturation
```

ppcano (approved):

src/data/markdown/docs/02 javascript api/04 k6-data/1-SharedArray.md
```diff
@@ -28,7 +28,8 @@ Which means that for the most part if you currently have an array data structure
 import { SharedArray } from 'k6/data';
 
 const data = new SharedArray('some name', function () {
-  // here you can open files, and then do additional processing on them or just generate the data dynamically
+  // here you can generate the data dynamically or open files and do additional processing
+  // it's important to open large files inside this function to avoid memory saturation
```
MStoykov, line 32:
```suggestion
  // All heavy work (opening and processing big files for example) should be done inside here.
  // This way it will happen only once and the result will be shared between all VUs, saving time and memory.
```

MStoykov: I feel like explaining *why* that is the case is equally if not important as I would expect that will be what the user will remember/understand  better.

MStoykov (approved):

oleiade (approved):

PR #577, opened 2022-02-23
Remove callouts as New for older releases

javaducky:
Some of the features called out as "New" are actually rather old (> 1 year) given latest release is v0.37.0.


ppcano:
Thanks for working on this!

Issue #441, opened 2021-10-13
Improve `ramping-arrival-rate` docs and examples


na--:
Prompted by this slack thread: https://k6io.slack.com/archives/C3E688J8Z/p1634033548234300

These pages can definitely be improved with more details on the topic on how to simulate ramping (which we should probably also call "variable", "non-constant" and other synonyms, so people who haven't heard the term "ramping" can find) arrival-rate load:
- https://k6.io/docs/using-k6/scenarios/executors/ramping-arrival-rate
- https://k6.io/docs/using-k6/scenarios/arrival-rate/
- https://k6.io/docs/using-k6/scenarios/advanced-examples/

Copying my answer, something like which we can use as another illustrative example:
> ah, yeah, I see how this can be confusing... :disappointed: I'll add a docs issue so we can better explain it, but in short:
> * `timeUnit` is constant for the whole duration of the scenario, you can't change it for a specific `stage`
> * you can think of the `target` in a specific stage as `targetRate`, i.e. what should the `rate` value be at the end of the stage's `duration`
> 
> so, if you have something like this:
> ```js
> import http from 'k6/http';
> import exec from 'k6/execution';
> 
> export let options = {
>     scenarios: {
>         foo: {
>             executor: 'ramping-arrival-rate',
>             preAllocatedVUs: 2,
> 
>             startRate: 10,
>             timeUnit: '1m',
>             stages: [
>                 { target: 10, duration: '1m' },
>                 { target: 60, duration: '5m' },
>                 { target: 60, duration: '5m' },
>                 { target: 5, duration: '1m' },
>             ],
>         },
>     },
> };
> 
> export default function () {
>     console.log(`[t=${(new Date()) - exec.scenario.startTime}ms] VU{${exec.vu.idInInstance}/${exec.vu.idInTest}} ran scenario iteration ${exec.scenario.> iterationInTest} (${exec.vu.iterationInScenario} for VU) in ${exec.scenario.name}`);
>     http.get('https://test.k6.io/');
> }
> ```
> 
> k6 will:
> * stay at 10 iterations per minute (i.e. 1/6s) for the first minute
> * linearly ramp up to 60 iters/min over the next 5 minutes
> * stay at 60 iters/min for the next 5 minutes
> * ramp down to 5 iters/min over the next minute
> * wait for up to 30s (the default `gracefulStop` value) for any already started iterations to finish, but without starting any new ones
> * run `teardown()` (if specified) and exit

PR #596, opened 2022-03-22
Clarify ramping arrival rate executor's docs

oleiade:
In this PR, I attempt to clarify the documentation section that's
specific to the ramping arrival rate executor, as requested by #441.

I have replaced usages of the word "execution" and "executed" with
"started" to clarify the fact that the executor would not guarantee to
execute a given number of iterations per `timeUnit`, but rather would
start them.

I also expanded the provided example to illustrate its usage more
explicitly, following what was proposed in #441.

## Question Marks/TODO 🔴 

I would need to update the example picture to represent the new example setup. Any ideas how this was generated, and I could update it to reflect the changes?

### DIFF
--- a/src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/01 Executors/06 ramping-arrival-rate.md
+++ b/src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/01 Executors/06 ramping-arrival-rate.md
@@ -1,13 +1,13 @@
 ---
 title: 'Ramping arrival rate'
-excerpt: 'A variable number of iterations are executed in a specified period of time.'
+excerpt: 'A variable number of iterations are started in a specified period of time.'
 ---
 
 ## Description
 
-A variable number of iterations are executed in a specified period of time. This is
-similar to the ramping VUs executor, but for iterations instead, and k6 will attempt
-to dynamically change the number of VUs to achieve the configured iteration rate.
+A variable number of iterations are started in specified periods of time. This is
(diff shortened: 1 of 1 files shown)


ppcano:
> Any ideas how this was generated, and I could update it to reflect the changes?
@javaducky did those running the test on k6 Cloud.

javaducky:
You may want to apply similar updates to all the other executors as well...I made them be all consistent when I previously went through this.

oleiade:
Hey @javaducky,

If I get to apply these changes to the other executors, I would want to do it in other PRs anyway 😸 

Regarding the chart illustration update, could you walk me through your workflow? I'm happy to make it myself, but I'd like to make it consistent with the other pre-existing ones?

javaducky:
Sadly, the creation was very manual. I simply ran the example script in the cloud app, did an area capture of the graphs, then edited the images using SnagIt. 

I only created (and added) those images a couple weeks ago, so they _may not_ have added that much to the docs...I plan to use them for the workshop documentation anyway. So...maybe it's _not_ necessary to have them in the docs for this iteration? 🤷 

Anyway, if you do decide that it's more work to create these images than the value they provide, feel free to drop them from the docs.

oleiade:
Hey folks, I gave a shot at producing a somewhat similar chart, but failed miserably. I'll leave it out for this PR, as I prefer to not have a chart, rather than an ugly inconsistent one. Maybe @javaducky could make one in a separate PR using the example of this one then?

Could I get your eyes on this @ppcano and @EdvinasDaugirdas so we merge it? 😸 👀

ppcano (approved):

PR #607, opened 2022-04-04
Update 02 per-vu-iterations.md

nicolapalavecino:


javaducky:
Hi @nicolapalavecino , thanks for you interest in keeping our docs up-to-date!

Looking at your changes, it appears the original text is in fact correct. If you consider the example on the same page, using `vus: 10` and `iterations: 20`, the total number of iterations performed with the script is 200 as expected and described.

PR #624, opened 2022-04-18
Reworks "life cycle" explanation page

MattDodsonEnglish:
This is a first attempt to clarify the very important explanation page "Stages of a test life cycle."

There is still much room to improve and expand, but I hope this current edit makes it a little easier for first-time users to conceptualize.

Since I added some content, I'd appreciate a second pair of eyes.


src/data/markdown/translated-guides/en/02 Using k6/06 Test life cycle.md
```diff
@@ -1,10 +1,15 @@
...
 
-The four distinct life cycle stages in a k6 test are "init", "setup", "VU" and "teardown"
-Throughout the documentation, you will also see us referring to it as "init code", "VU code" etc.
+A k6 test has four distinct stages, which always run in the same order:
+
+1.  *Init code* initializes VUs
+2. *Setup code* sets up data and the test environment
+3. *VU* code runs the test
+4. *Teardown* cleans up data and stops the test environment
+
```
MStoykov, line 12:
```suggestion
1. *Init code* initializes VUs
2. *Setup code* sets up data and the test environment
3. *VU* code runs the test
4. *Teardown* cleans up data and stops the test environment

```

If we are listing `setup` and `teardown` I guess we should list `handleSummary`

Also, we do have `code` three times, two times bold, one not. Also, we could say `teardown code` the same we can say `setup code` so, which all seems a bit inconsistent.

src/data/markdown/translated-guides/en/02 Using k6/06 Test life cycle.md
```diff
@@ -1,10 +1,15 @@
...
-excerpt: 'The four distinct life cycle stages in a k6 test are "init", "setup", "VU" and "teardown".'
+excerpt: 'The four distinct life cycle stages in a k6 test are "init", "setup", "VU", and "teardown".'
 ---
 
-The four distinct life cycle stages in a k6 test are "init", "setup", "VU" and "teardown"
-Throughout the documentation, you will also see us referring to it as "init code", "VU code" etc.
+A k6 test has four distinct stages, which always run in the same order:
+
+1. *Init code* initializes VUs
+2. *Setup code* sets up data and the test environment
```
ppcano, line 9: might be good to mark it as Optional.

src/data/markdown/translated-guides/en/02 Using k6/06 Test life cycle.md
```diff
@@ -1,10 +1,15 @@
...
+3. *VU code* runs the test
+4. *Teardown code* cleans up data and stops the test environment
```
ppcano, line 11: might be good to mark it as Optional.

src/data/markdown/translated-guides/en/02 Using k6/06 Test life cycle.md
```diff
@@ -41,33 +62,45 @@ export default function () {
...
-execution. In the case of cloud and clustered execution we know which files will be needed, so
-we distribute only those files. We know which modules will be imported, so we can bundle them
-up from the get-go. And, tying into the performance point above, the other nodes don't even
-need writable filesystems - everything can be kept in-memory.
+Why separate init and VU code? There are two chief reasons:
+* Better performance
+* Better design
 
-As an added bonus, you can use this to reuse data between iterations (but only for the same VU):
+### Performance is faster and more accurate
```
ppcano, line 86: I suggest removing both sections. Perhaps a short explanation with the bullet points. 

If we want to expand the topic, we could create an additional section on other page.

MattDodsonEnglish: Now, I've removed the H3s, and made that section its own H2.

What do you think? It seems like it works as its own section. I believe it's good to split content into digestible sections, making it a little easier for the reader to scan, and avoiding the appearance of the dreaded "wall of text".

src/data/markdown/translated-guides/en/02 Using k6/06 Test life cycle.md
MattDodsonEnglish: Thanks! I fixed the list.

I'm not sure about whether to add `handleSummary` here. While it's definitely part of the code, it doesn't seem as much a part of the test's "life cycle". These four stages distinctly influence how the test runs. `handleSummary` influences output, but not the actual test.

Does that make sense to you? This is just my initial opinion here.

src/data/markdown/translated-guides/en/02 Using k6/06 Test life cycle.md
```diff
@@ -26,10 +31,26 @@ export function teardown(data) {
...
 </CodeGroup>
 
+## A quick overview of test stages
+
+This table provides the essential information about each stage.
+The rest of this page goes into deeper technical detail.
+
+| Test stage      | Used to                                                    | Example                                                                                 | Called                                                                             | Required? |
+|-----------------|------------------------------------------------------------|-----------------------------------------------------------------------------------------|------------------------------------------------------------------------------------|-----------|
+| **1. init**     | Load local files, import modules, declare global variables | Open JSON file, Import module                                                           | Once per VU\*                                                                      | Optional  |
```
ppcano, line 41: Init is optional, but a test without importing any k6 API is useless.  Should Init be required ??

src/data/markdown/translated-guides/en/02 Using k6/06 Test life cycle.md
```diff
@@ -83,29 +112,27 @@ export default function () {
...
+might be torn down  (depending on your test configuration options).
 
-> Make sure to use `sleep()` statements to pace your VUs properly. An appropriate amount of
-> sleep/think time at the end of the default function is often needed to properly simulate a
-> user reading content on a page. If you don't have a `sleep()` statement at the end of
-> the default function your VU might be more "aggressive" than you've planned.
+> Make sure to use `sleep()` statements to pace your VUs properly, simulating a user reading content on your page.
+> If you don't have a `sleep()` statement at the end of the default function, your VU might be more "aggressive" than you've planned.
 >
 > VU without any `sleep()` is akin to a user who constantly presses F5 to refresh the page.
```
ppcano, line 125: Consider removing it. It might mislead users to website testing.

src/data/markdown/translated-guides/en/02 Using k6/06 Test life cycle.md
```diff
@@ -188,14 +213,14 @@ export default function (data) {
 
 </CodeGroup>
 
-Note that any requests made in the setup and teardown stages will be counted in the end-of-test
-summary. Those requests will be tagged appropriately with the `::setup` and `::teardown` values
+Note that the [the end-of-test summary](/results-visualization/end-of-test-summary/#summary-export-to-a-json-file) includes requests made in the setup and teardown stages.
+These requests are tagged appropriately with the `::setup` and `::teardown` values
 for the `group` metric tag, so that you can filter them in JSON output or InfluxDB.
 
 ## Skip setup and teardown execution
```
ppcano, line 220: This is an advanced case. I suggest removing it from the ToC.

src/data/markdown/translated-guides/en/02 Using k6/06 Test life cycle.md
```diff
@@ -129,19 +156,18 @@ export function teardown(data) {
...
+You might have noticed the function signatures of the `default` and `teardown` functions
+take an argument, referred to here as `data`.
 
-This `data` will be whatever is returned in the `setup` function, so a mechanism for passing data
-from the setup stage to the subsequent VU and teardown stages in a way that, again, is compatible
-with our goal of supporting local, cloud and clustered execution modes without requiring script
-changes when switching between them. (it might or might not be the same node that runs the setup
-and teardown stages in the cloud or clustered execution mode).
+This `data` is whatever the `setup` function returns.
+Having a mechanism for passing data from the setup stage to the subsequent VU and teardown stages is compatible with our goal of supporting local, cloud, and clustered execution modes without requiring script changes when switching between them.
```
ppcano, line 163: Consider removing it.  Is this really necessary?

src/data/markdown/translated-guides/en/02 Using k6/06 Test life cycle.md
```diff
@@ -83,29 +112,27 @@ export default function () {
...
-at the beginning of the test, after the init stage but before the VU stage (`default` function),
-and `teardown` is called at the end of a test, after the VU stage (`default` function). Therefore,
-VU number is 0 while executing the `setup` and `teardown` functions.
+Like `default`, `setup` and `teardown` functions must be exported functions.
+But unlike the `default` function, k6 calls `setup` and `teardown` only once per test. 
+
+* `setup` is called at the beginning of the test, after the init stage but before the VU stage.
+* `teardown` is called at the end of a test, after the VU stage (`default` function).
+
+The VU number is 0 while the `setup` and `teardown` functions execute.
```
ppcano, line 135: I am unsure this belongs here. It is a very specific detail and looks out of context here.

I would consider it moving it together with:

- Note that the end-of-test summary includes requests made in the setup and teardown stages.
- Skip setup and teardown execution
- The VU number is 0 while the `setup` and `teardown` functions execute

ppcano (commented):
more comments

src/data/markdown/translated-guides/en/02 Using k6/06 Test life cycle.md
ppcano: have doubts. 

2. *Setup code* sets up data and the test environment. Optional. 

Don't hold a strong opinion.

src/data/markdown/translated-guides/en/02 Using k6/06 Test life cycle.md
MStoykov: Init isn't optional. You always need init code as this is the place that defines which function is the default ;).
All the code that is outside of a function will be executed *always* at the begining, including if a function is called

```javascript
function one() {
  // this code is not init context and isn't called so it won't be executed 
}
function two() {
  // this code is not the init context but is called
}

two() // this is in the init context and calles the function `two` so it's code will also be executed

// export default function() {} // if this isn't uncommented k6 will give you error because nothing was defined as the function to be executed 

export function setup() { // this line that is in the "init code" is what defines that there is `setup` to be called
  // some setup code
} 
```
I hope this makes it clear that "init code" is totally not optional ;). 

This actually is one of the reason (IMO) that we don't call it init code, but init context. As that moves the thinking away from code and to *where* the code is being executed (in the init context). 

I would expect there is probably even a better way to explain this using words that are more specific and don't let so much for interpretation. 

WDYT @na--

src/data/markdown/translated-guides/en/02 Using k6/06 Test life cycle.md
```diff
@@ -26,10 +31,26 @@ export function teardown(data) {
...
+|-----------------|------------------------------------------------------------|-----------------------------------------------------------------------------------------|------------------------------------------------------------------------------------|-----------|
+| **1. init**     | Load local files, import modules, declare global variables | Open JSON file, Import module                                                           | Once per VU\*                                                                      | Optional  |
+| **2. Setup**    | Set up data for processing, share data among VUs           | Call API to start test environment                                                      | Once                                                                               | Optional  |
+| **3. VU code**  | Run the test script                                        | Make https requests, validate responses                                                 | Per VU, for as many iterations and as long a duration as the test options required | Required       |
```
MStoykov, line 43:
```suggestion
| **3. VU code**  | Run the test function, usually the `default` one                                        | Make https requests, validate responses                                                 | Once per iteration, as many times  as the test options required | Required       |
```

I would want to make it clear that we are calling *a function*, usually the `default` one, not a "script", which has a definition in ECMAScript which is more or less the whole file. But in practice "run the test script" is probably more technically correct to be said about the "init code' ;) 

Additionally, while the last part is now more vague it is also more correct as otherwise it doesn't really work for arrival rates or `shared-iterations` , especially as the duration is just a limit in the second one, not how long it should run.

src/data/markdown/translated-guides/en/02 Using k6/06 Test life cycle.md
```diff
@@ -26,10 +31,26 @@ export function teardown(data) {
...
+
+\* In cloud scripts, init code might be called more often.
+
+
+
 ## Init and VU stages
 
-Scripts must contain, at the very least, a `default` function - this defines the entry point
-for your VUs, similar to the `main()` function in many other languages:
+Scripts must contain, at the very least, a `default` function.
```
MStoykov, line 52: Technically not - through the use of scenarios' `exec` setting you can define another exported function as the entry point for that scenarios. If all scenarios do that - k6 will not need a `default` function.

src/data/markdown/translated-guides/en/02 Using k6/06 Test life cycle.md
```diff
@@ -41,33 +62,41 @@ export default function () {
...
+
+Code inside `default` is called *VU code*.
+VU code runs over and over through the test duration.
+Besides setup and teardown, code outside of `default` is  *init code*.
+It runs only once per VU.
+
+VU code can make HTTP requests, emit metrics, and generally do everything you'd expect a load test to do.
+There are a few important exceptions. VU code:
+* *Does not* load files from your local filesystem,
+* *Does not* import any other modules.
```
MStoykov, line 78: I think that this explains *what* you can do and where, but not *why* which arguably is what the question above asks. 

The short answer is that "we need the script to be executed in order to know what k6 should do". That includes:
- getting hte configuration
- knowing which functions to be called on `setup`/`teardown` and iterations
- all the things above that you have mentioned above are also separated in order to 
  - on one side not generate metrics whenever we have to run the "init code" and to be able to do it fairly repeatable
  - getting an `archive` with everything needed to run the code somewhere else (like in the cloud)  

And that last two points is where a lot of the restrictions come from. But ultimately we just *use* the fact that we need to be able to run "init code" to get the `options` and the definitions to do that.

Also, ultimately running 
```javascript
let b = 5;
```
twice in the same javascript context will give you an error as `b` has already been defined. Putting it in the function and calling *the function* is what makes a new context and let you do 

```javascript
function f () {
  let res = http.request()
  // something with res
}
f();
f();
```
So there is also kind of another technical reason for this.

MattDodsonEnglish: Thanks for these explanations! I think you are getting at something very essential to keep in mind when explaining these internals.

 We always must ask: _what does this information mean to the user? What bad thing will happen if they don't know this information?_

 It's tricky sometimes to separate what information matters for implementers from what matters for end users.

ppcano:
**How does k6 execute scripts?** great. 
 
 **Why k6 works this way** is very advanced and too detailed. I think it should not be shown upfront. I suggest using either an internal page or the "collapsible component".

- `Functions can be reused`: unsure about this one.

MattDodsonEnglish:
> Why k6 works this way is very advanced and too detailed. I think it should not be shown upfront. I suggest using either an internal page or the "collapsible component".

This comment made me think that the whole article could be restructured and simplified. Basically, give the very TL,DR info in the intro. Then, use the table to go into a bit more detail. After that, explain each stage. Finally, provide the reasons why this works this way at the end.

This would give the whole document a structure of [progressive disclosure](https://en.wikipedia.org/wiki/Progressive_disclosure). Generally it's good to leave the most technical details at the end.


```
# The Test lifecycle
<Intro: same content as "How does k6 execute scripts>
<code stub>
## Essential information about each test stage
<Table> 
## The init context
<Information about what init does. I think this is the only content that's missing>
## VU code runs in the default function
### The default function lifecycle
## Setup and teardown code
## Why k6 works this way
```

src/data/markdown/translated-guides/en/02 Using k6/06 Test life cycle.md
```diff
@@ -160,11 +158,8 @@ You might have noticed the function signatures of the `default` and `teardown` f
 take an argument, referred to here as `data`.
 
 This `data` is whatever the `setup` function returns.
-Having a mechanism for passing data from the setup stage to the subsequent VU and teardown stages is compatible with our goal of supporting local, cloud, and clustered execution modes without requiring script changes when switching between them.
-In cloud and clustered modes, the node that runs the setup code might not be the node that runs the teardown code.
 
-To support these modes, you can pass only data (i.e. JSON) between `setup()` and the
-other stages.
+You can pass only data (i.e. JSON) between `setup()` and the other stages.
```
MStoykov, line 158: Maybe a note that it's a copy of what `setup` returns. So that it:
1. uses more memory if it's big
2. you can't actually change it in the VU code and see the changes in the teardown. Although any VU will see it's changes to the `data`. This though is likely to change as we might freeze the whole object just to make it less obvious that it won't work

src/data/markdown/translated-guides/en/02 Using k6/06 Test life cycle.md
```diff
@@ -1,10 +1,18 @@
...
 ---
 
-The four distinct life cycle stages in a k6 test are "init", "setup", "VU" and "teardown"
-Throughout the documentation, you will also see us referring to it as "init code", "VU code" etc.
+A k6 test has four distinct stages, which always run in the same order:
+
+1. Code in the `init` context prepares the script. During this preparation, the `init` context
+     - Imports modules
+     - Loads files from the local filesystem
+     - initializes all VUs, configuring their `options`
```
MStoykov, line 11:
```suggestion
     - configures the whole k6 through `options`
```

it doesn't "initialize" VUs.

src/data/markdown/translated-guides/en/02 Using k6/06 Test life cycle.md
MattDodsonEnglish: > Although any VU will see it's changes to the data. This though is likely to change as we might freeze the whole object just to make it less obvious that it won't work

@MStoykov 

Sorry, I'm a bit confused here. So, VU code can manipulate data from setup, but it can't pass that data to teardown. Is that correct?

And, in the table, I note that setup code can be used to " share data among VUs". How is this data shared. Is it from one VU to the next VU? Or does each VU grab "clean" data from the `setup` code?

However it works, it seems like, if VU code can manipulate data that doesn't end up in the teardown stage, then the data is not really "passed" from Setup to VU to Default. 

Rather, setup passes data in two places. Once, directly to `teardown`. And once (or many times??) directly to `default`.

I've made a small diagram to illustrate how I conceive this. Excuse me, I'm sure said some inaccurate things here, but I thought it may be easier to correct than to explain :-) .

image

MStoykov: more or less -yes to all of those :). 

In practice implementing `data` being modified and all that modifications being visible by :
1. all VUs
2. teardown
Will be ... extremely complicated and (computationally) expensive especially in distributed execution. Which is why we haven't even discussed it ever. But long long time ago I made fixes so that they have separate copies - previous to that this was sometimes panicking as multiple VUs were writing to the same not thread safe structure.

MattDodsonEnglish: Haha, this sounds pretty complicated. I think I'm just going to just suggest that it's a "copy," and leave it at that. Especially since it sounds like this data may get "frozen" in the VU stage in some later k6 version anyway.  

 For the people who are motivated enough to actually use this implementation in some advanced way, well they can look at the source code (and find this discussion). :slightly_smiling_face:

MattDodsonEnglish:
@ppcano @MStoykov 

With 534db8c80818284b0a9578e61e28c06d597214be , this page is almost at a complete rewrite now. 

I tried to add progressively more detail. Hopefully, the page gives readers a solid mental model of how a k6 script runs, without adding too much technical details or suggesting complex/impossible operations (e.g passing and manipulating data from setup to VU to teardown).

I'd like the diagram to be more beautiful, and I'd prefer not having the "description list" in pure HTML, but fixing those two issues would require some time investment.

If you all think it looks alright, I'd like to merge this.

ppcano (approved):

PR #630, opened 2022-04-22
feat: update list styles

### LINKED ISSUE #626: Cannot embed bullets in a numbered list
It seems impossible to use markdown to embed an unordered list within an ordered list. In Github markdown, it's possible to do by indenting the nested items with two spaces:

1. Item 1 contains unordered items
   * thing 1
   * thing 2
2. Item 2
3. Item 3

On the k6-docs site, this doesn't seem possible.
I've tried a few variations. In each attempt, the site generator presents the list in a surprising and incorrect way.

A) Using two-space indention.

When the unordered items are indented with two spaces, the generator assumes that the next ordered item begins a new ordered list.

See how it jumps from 1 back to 1:

image

b) Indenting with more spaces

Indenting with more than two spaces creates an embedded list that continues the numbering. Even if we wanted to embed a numbered list like this, we almost certainly want the embedded numbers to have their own numbering.

image

SilencerWeb:
This PR brings an update of list styles in order to resolve #626

### DIFF
--- a/src/data/markdown/translated-guides/en/02 Using k6/06 Test life cycle.md
+++ b/src/data/markdown/translated-guides/en/02 Using k6/06 Test life cycle.md
@@ -6,6 +6,16 @@ excerpt: 'The four distinct life cycle stages in a k6 test are "init", "setup", 
 The four distinct life cycle stages in a k6 test are "init", "setup", "VU" and "teardown"
 Throughout the documentation, you will also see us referring to it as "init code", "VU code" etc.
 
+Delete before merging:
+
+1. Item 1 contains unordered items
+   * thing 1
+   * thing 2
+2. Item 2
+   1. thing 1
+   2. thing 2
+3. Item 3
+
 <CodeGroup labels={["The four life cycle stages"]} lineNumbers={[true]}>
 
 ```javascript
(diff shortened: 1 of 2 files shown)


MattDodsonEnglish:
I've pulled this branch, and tested it locally.

Visually, I think it looks great, but it seems the numbered list still restarts after the unordered list. Even with the `ol` `start` attribute having a value of 2.

In the screenshots, you can see the markdown source and the generated page and HTML source.

image

image


Let me know if I've missed something.

MattDodsonEnglish:
I've pulled this branch, and tested it locally.

Visually, I think it looks great, but it seems the numbered list still restarts after the unordered list. Even with the `ol` `start` attribute having a value of 2.

In the screenshots, you can see the markdown source and the generated page and HTML source.

image

image


Let me know if I've missed something.

SilencerWeb:
@MattDodsonEnglish are you sure you are testing changes on a branch [nested-lists](https://github.com/grafana/k6-docs/tree/nested-lists)?

I've tried to reproduce the behavior but everything works fine for me:

image

I've pushed an [example](https://github.com/grafana/k6-docs/pull/630/files#diff-451907ca6b236224983fb0b1deb7a249f0c2faff29b29b0373dc0dec97d4a99a) for you that needs to be deleted before merging PR

Example can be checked out live [here](https://mdr-ci.staging.k6.io/docs/refs/pull/630/merge/using-k6/test-life-cycle/)

MattDodsonEnglish:
Awesome, thanks for adding the example. Not sure what I did wrong on my end.

LGTM.

PR #625, opened 2022-04-19
chore: replace let with const

Maksimall89:

PR #637, opened 2022-04-25
chore: update linters setup

SilencerWeb:
This PR brings an update for linter and code formatter: commands, commit flow, and CI/CD. Also, it fixes all current errors and warnings that linter and code formatter showed before.

We have 7 commands for linting and formatting now:
1. `format` - checks code formatting with Prettier (works for `.js`, `.jsx`, `.css`, and `.scss` extensions)
2. `format:fix` - same as `format` but fixes automatically whatever is possible to fix
3. `lint` - lints code with ESLint using root `.eslintrc` config (works for `.js`, and `.jsx` extensions)
4. `lint:fix` — same as `lint` but fixes automatically whatever is possible to fix
5. `lint:examples` — lints code examples in `src/data/markdown` using `src/data/markdown/.eslintrc.js` config
6. `lint:examples:fix` — same as `lint:examples` but fixes automatically whatever is possible to fix
7. `lint:prose` — lints prose, works the same as before

We have 3 commands that are being run on a commit:
1. `"*.{js,jsx,css,scss}": "prettier --write"` — works the same as `format:fix`
2. `"*.{js,jsx}": "eslint --fix"` — works the same as `lint:fix`
3. `"*.{md,mdx}": "eslint -c ./src/data/markdown/.eslintrc.js --fix"` — works the same as `lint:examples:fix`

This way, code that doesn't meet our standards:
1. Will be automatically fixed when making a commit
2. If it can't be automatically fixed, then developers will see errors that they need to handle in order to successfully make a commit

To make 100% sure that code meets our standards, we run commands `format`, `lint`, `lint:examples` and `lint:prose` in CI/CD

Some suggestions:
1. Run Prettier on all files possible (YAML, JSON, dotfiles, etc)
2. Run ESLint on all markdown files, not only on files in `src/data/markdown`
3. Think of using [Markdownlint](https://github.com/DavidAnson/markdownlint) for linting Markdown files
4. Upgrade all dependencies related to linting and formatting to the latest versions
5. Think if there are some files that need to be excluded from linting and formatting

After we finish with this PR, I'll duplicate changes to other repositories

Update:
- Removed dependency [`babel-eslint`](https://www.npmjs.com/package/babel-eslint) since it's no longer supported. At first, replaced it with [`@babel/eslint-parser`](https://www.npmjs.com/package/@babel/eslint-parser) but seems like we do not need it, at all. Note from docs:

  > You only need to use @babel/eslint-parser if you are using Babel to transform your code. If this is not the case, please use the relevant parser for your chosen flavor of ECMAScript (note that the default parser supports all non-experimental syntax as well as JSX).
- Upgraded ESLint and Prettier to the latest versions
- Fixed ESLint errors after upgrading to the latest versions
- Turned off 2 new ESLint rules: `react/function-component-definition` and `react/no-unstable-nested-components`. The second one seems to be pretty useful though but we'll need to refactor some components and it's better to do it in another PR
- Upgraded Node version to v16 since ESLint v8 is not supporting node v15 anymore, [read more](https://eslint.org/docs/8.0.0/user-guide/migrating-to-8.0.0#-nodejs-10-13-and-15-are-no-longer-supported)
- Removed everything related to MDX from the root ESLint config since we are using it only for JavaScript files
- Tested changes for Github Workflows in https://github.com/grafana/k6-docs/pull/641


ppcano:
@SilencerWeb 

Please, check why this PR does not detect the error mentioned at https://github.com/grafana/k6-docs/pull/625/commits/05ede4600de2ae371ba66340eb308d30048413a7

To avoid bringing more complexity, I think we can now upgrade dependencies and postpone for later (upon request) the other suggestions

- Run Prettier on all files possible (YAML, JSON, dotfiles, etc)
- Run ESLint on all markdown files, not only on files in src/data/markdown
- Think of using [Markdownlint](https://github.com/DavidAnson/markdownlint) for linting Markdown files.
- Think if there are some files that need to be excluded from linting and formatting

SilencerWeb:
Hey @ppcano!

This PR does not detect the error mentioned at https://github.com/grafana/k6-docs/pull/625/commits/05ede4600de2ae371ba66340eb308d30048413a7 because changing to `const` will make the code invalid since `options` variable is being used later in the code

Full code:

Error when changing `let` to `const`:
image

bolotskydev (approved):

PR #652, opened 2022-04-29
fix: fix ID generation for TOC

SilencerWeb:
This PR returns back condition `id || anchorify(innerText)` when generating links for TOC because sometimes TOC renders before headings get IDs assigned

Also, this PR brings a small fix for anchor selector in headings


SilencerWeb:
cc @ppcano

PR #657, opened 2022-05-04
feat: add xk6-output-error

Maksimall89:


Maksimall89:
@javaducky thanks for the notice about a change to the `state` package. I will  check and fix all my extensions.

PR #663, opened 2022-05-09
Fix the output intro URL

ankur22:
The link in https://k6.io/docs/extensions/guides/what-are-k6-extensions/#:~:text=for%20many%20popular-,output%20backends,-%2C%20this%20list%20will goes to https://k6.io/docs/docs/getting-started/results-output/, but this leads to a blank page. What I believe this needs to go to is https://k6.io/docs/getting-started/results-output/ i.e. remove the extra `docs` in the url path.


ankur22:
I've confirmed that the new url works

inancgumus (approved):

PR #665, opened 2022-05-10
In not-spanish we don't use dots to detonate thousands

mstoykov:


mstoykov:
This seemed to be stolen from the Spanish version where all numbers have `.` to denote each 1000 or 1 000 000 ... This seems to be [okay for Spanish](https://docs.oracle.com/cd/E19455-01/806-0169/overview-9/index.html), but in all other language notation this makes it seem like a decimal place - so k6 can handle 30 to 40 VUs, which is clearly not the case.

The Spanish version uses `.` everywhere which seems to be correct for ... Spanish.

WDYT @ppcano @dgzlopes @MattDodsonEnglish

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -8,7 +8,7 @@ This document explains how to launch a large-scale k6 test on a single machine w
 The common misconception of many load testers is that [distributed execution](#distributed-execution) (ability to launch a load test on multiple machines) is required to generate large load. This is not the case with k6.
 
 k6 is different from many other load testing tools in the way it handles hardware resources. A single k6 process will efficiently use all CPU cores on a load generator machine.
-A single instance of k6 is often enough to generate load of 30.000-40.000 simultaneous users (VUs). This amount of VUs can generate upwards of 300,000 requests per second (RPS).
+A single instance of k6 is often enough to generate load of 30,000-40,000 simultaneous users (VUs). This amount of VUs can generate upwards of 300,000 requests per second (RPS).
```
MattDodsonEnglish, line 11:
```suggestion
A single instance of k6 is often enough to generate load of 30,000-40,000 simultaneous users (VUs). This number of VUs can generate upwards of 300,000 requests per second (RPS).
```

MattDodsonEnglish (changes requested):
I agree, especially since we use a comma as a separator of thousands in the next sentence! This also gave me the chance to add my own nitpick fix about countable nouns.

PR #668, opened 2022-05-12
Replace k6 module mention with @types/k6

oleiade:
As discussed internally, this PR replaces the mention of the k6 module in the **using k6** section, with a mention of the `@types/k6` typescript package. 

ref https://github.com/grafana/k6/issues/2516


olegbespalov (approved):

src/data/markdown/translated-guides/en/02 Using k6/07 Modules.md
```diff
@@ -77,12 +77,17 @@ These libraries can either be downloaded and included with the test project or l
...
+  - [Built-in modules](#built-in-modules)
+  - [Local filesystem modules](#local-filesystem-modules)
+  - [Remote HTTP(S) modules](#remote-https-modules)
+  - [The JSLib repository](#the-jslib-repository)
+- [Bundling node modules](#bundling-node-modules)
+  - [Setting up the bundler](#setting-up-the-bundler)
+  - [Running the bundling](#running-the-bundling)
+  - [Running the tests](#running-the-tests)
+- [Using local modules with Docker](#using-local-modules-with-docker)
+- [See also](#see-also)
```
mstoykov, line 90: This seems wrong - this is index of the whole page but : 
1. is in the middle of the page
2. from the line just about is only about how to Bundle node modules  

so I feel like this should just be reverted. 

Was there a particular reason you decided to change this?

oleiade: I can't remember touching this 😞 I guess my mardown formatter might be the issue. Will fix it.

src/data/markdown/translated-guides/en/02 Using k6/07 Modules.md
```diff
@@ -170,10 +175,10 @@ $ npm install --save-dev \
 </CodeGroup>
 
 | Package                                                                                   | Usage                                                                                                                                                                                                                                                                 |
-| ----------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| :---------------------------------------------------------------------------------------- | :-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
```
mstoykov, line 178: I also don't know why you added `:` ? Seems to work so apperantly it's fine with the makrdown processor we are using, I just never seen `:` in markdown tables before :thinking:

oleiade: I use a mardown table formatter. The `|:` is actually correct markdown, it indicates that the table cell should be left aligned 😺 https://www.markdownguide.org/extended-syntax/#alignment

From what I can read here and there, this format is indeed supported by Gatsby.

mstoykov (approved):
LGTM, although this doesnt' really close the reference issue IMO - we still need to do something with the package

PR #670, opened 2022-05-13
Rewrites end-of-test summary doc

MattDodsonEnglish:
This is almost a complete rewrite, so I'd appreciate a review. I am especially still unsure about the sections of the `handleSummary` part:

* Is the information about the keys and values correct?
* Is that example of requests to remote good? Does it include too much?

Besides adding examples, I've restructured quite a bit, trying to make the doc easier to process. I also moved options to its own h2, since some options are shared by both `handlesummary` and the default report.

Likely, the docs on output need more reworking, but this is a start.

### DIFF
--- a/src/data/markdown/translated-guides/en/04 Results visualization/00 End-of-test Summary.md
+++ b/src/data/markdown/translated-guides/en/04 Results visualization/00 End-of-test Summary.md
@@ -61,13 +61,8 @@ Ramp_Up ✓ [======================================] 00/20 VUs  30s
 
 ERRO[0044] some thresholds have failed  
 ```
+
 </CodeGroup>
-
-### Summary options
-
-A few options affect how this report behaves:
-- The [`--summary-trend-stats` option](/using-k6/options#summary-trend-stats) defines which [Trend metric](/javascript-api/k6-metrics/trend) statistics to calculate and show.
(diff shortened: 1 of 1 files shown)


src/data/markdown/translated-guides/en/04 Results visualization/00 End-of-test Summary.md
```diff
@@ -3,102 +3,120 @@ title: 'End-of-test summary'
...
+     vus............................: 1      min=1      max=19
+     vus_max........................: 20     min=20     max=20
+
+ERRO[0044] some thresholds have failed  
 ```
 
 </CodeGroup>
 
-## Summary options
+<Collapsible title="Summary export to a JSON file (Discouraged)">
```
ppcano, line 67: I suggest moving the content of the Collapsible component to the Summary Options section. Something alongs the line:

- The [--summary-export:](/using-k6/options#summary-export) exports a summary report with a predefined JSON format to a file. This option is Discouraged. Use the `handleSummary` callback instead.


Note that the same Collapsible component is shown twice. The next one is in the `Custom output examples` section.  This is a duplication and should be removed.

ppcano (changes requested):
Very good improvements. Ready to go from me, except the requested change.

PR #671, opened 2022-05-13
Updated the threshold key in the example

SimonPerryman:
Update the threshold key to be in line with the name of the counter.

Using the current example produces the error:

ERRO[0000] invalid threshold defined on requests; reason: no metric name "requests" found


MattDodsonEnglish:
Thank you! Just to double check,  I ran the new example on my machine with v0.38.2 . Looks good to me.

PR #650, opened 2022-04-29
feat: add Tooltip component for Markdown content

SilencerWeb:
This PR adds 3 components in order to resolve #648:
1. `Tooltip` — just a general tooltip that can be used across docs
2. `BNIT` — xk6-browser "Not Implemented" tooltip
3. `BWIPT` — xk6-browser "Work in Progress" tooltip

Usage example:
```jsx
// BNIT *without* link to the issue
<BNIT/>

// BNIT *with* link to the issue
<BNIT id="103"/>


// BWIPT *without* link to the issue
<BWIPT/>

// BNIT *with* link to the issue
<BWIPT id="103"/>
```

TODO:
- [x] Remove examples before merging PR

PR #674, opened 2022-05-17
feat: add component TableWithNestedRows

### LINKED ISSUE #656: Better design for documenting object properties
While working on [xk6-browser JS API docs](https://github.com/grafana/k6-docs/issues/582) (WIP branch [`docs/582-xk6-browser`](https://github.com/grafana/k6-docs/tree/docs/582-xk6-browser)) we have to document many methods that receive objects, and it's important to document each property as well.

We're currently experimenting with two approaches for this:

- [Document properties in a separate table and link to it from the object's Description column.](https://github.com/grafana/k6-docs/blob/3416a29a957f2bc3346ddcf28fd985c1c4b13548/src/data/markdown/docs/30%20xk6-browser/01%20xk6-browser/01-browser/05-newcontext--options--.md)
  The problem with this is that it clutters the page with many tables, and forces the user to navigate from the original table to view the properties and then back again, which is not great UX.
  This is mostly in line with how the [k6 docs](https://k6.io/docs/javascript-api/k6-http/batch/) look, but the xk6-browser JS API is much larger, so the approach doesn't scale well.
  It also wouldn't work well with nested objects, as the page would look too cluttered and confusing.

- [Document properties in the same table, right below the object itself.](https://github.com/grafana/k6-docs/commit/3416a29a957f2bc3346ddcf28fd985c1c4b13548#diff-ce3514099e143339d07ca744e875bd6a288d9315af4a26a76ddfd08622e7f740R8-R13)
  I think this is an improvement, and is more in line with how [Playwright](https://playwright.dev/docs/api/class-page#page-pdf) and [Puppeteer](https://pptr.dev/#?product=Puppeteer&version=v13.7.0&show=api-pagepdfoptions) docs look, although they use unordered lists, which can be easily nested.
  This initial version simply prefixes the properties with a hyphen, but we'd like to make it more visually obvious that these are properties of the object above, and not method arguments.

So this issue is to discuss design ideas and ultimately an implementation to improve this.

A good visual cue would be to indent properties and "link them" somehow to the object. E.g.:

```
geolocation
├─ latitude
├─ longitude
└─ accuracy
```

Or for nested props:

```
toplevel
└─ prop
   └─ nested
```

Alternatively, we could shift right the left table border for properties to have something like:

```
| toplevel arg | ... |
  | prop       | ... |
    | nested   | ... |
| another arg  | ... |
```

@inancgumus mentioned having sub-tables for properties and making them collapsible. Sub-tables could work, but I'm not sure if we should add the collapsible feature, as properties should be visible by default. The problem is not so much the amount of options/properties, but having them scattered around the page.

Let us know if you have any ideas to accomplish this. Thanks!

SilencerWeb:
This PR adds component `TableWithNestedRows`. This component will automatically parse nested properties and collapse them by default.

> ⚠️ Note: Make sure that object name is always the same since it's very important in order to get expected result

In order to use this component wrap a table in `TableWithNestedRows`, e.g.
```mdx
<TableWithNestedRows>

| Parameter             | Type   | Description                                         |
|-----------------------|--------|-----------------------------------------------------|
| geolocation           | object |                                                     |
| geolocation.latitude  | number | Latitude between -90 and 90                         |
| geolocation.longitude | number | Latitude between -180 and 180                       |
| geolocation.accuracy  | number | Optional non-negative accuracy value. Defaults to 0 |

</TableWithNestedRows>
```
> ⚠️ Note: Make sure that there are empty lines around a table, otherwise component won't work!

Collapsed example:

image

Expanded example:

image

Closes #656 

TODO:
- [x] Remove example before merging PR


SilencerWeb:
Hey guys @inancgumus @imiric @ppcano!

The component is ready, you can check it out [here](https://mdr-ci.staging.k6.io/docs/refs/pull/674/merge/javascript-api/k6-http/response/). Waiting for you feedback!

Also, could you please give me more real examples for this component so I could test it out with real content?

inancgumus:
Hey @SilencerWeb, nice work, thanks!

### 1

Can we make some style changes if @ppcano agrees too? IDK how, but is there a way to stand out somehow that the expanded table belongs to the main object? Maybe, with another background color or using some indicators as follows:

```
geolocation
├─ latitude
├─ longitude
└─ accuracy
```

(suggested by @imiric in #656)

Or, with right padding the sub-tables as follows (as I suggested before in #656):

166695692-b096f4f4-6d32-4d46-9b8a-421958e85e86

As I said, I don't have the best idea in UX wise, so these are just examples.

### 2

Would it be better if we remove the main object name from the sub-properties as well (while displaying them)?

Instead of the following:

```
geolocation
geolocation.latitude
```

To the following:

```
geolocation
latitude
...
```

ppcano:
@inancgumus, this issue is stagnated. What is pending?

inancgumus:
Hey @ppcano, we should decide on the layout format. Let's also ask other xk6 browser team members. @imiric @ankur22

ankur22:
Looks like a good start. I agree with what has been said, and this is basically a repeat:

1. Can we do some UX wizardry to make it clearer that the sub-table is part of the initial unexpanded table. I also think it might be visually clearer if the original table didn't change shape horizontally, but only expanded vertically under the current row so that a hidden sub table appeared just like the example that @inancgumus is displaying in his comment.

2. The dot notation is very clear (i.e. `geolocation.latitude`) as to what is expected from the user, but if we're going to have many nested tables within each other than that notation will not be maintainable and become visually unappealing. So I would drop it (i.e. `latitude`).

imiric:
Hi @SilencerWeb, can you let us know about the status of this, and how feasible it would be to implement the UI changes we mention above and in #656?

This is not blocking us from continuing with the documentation, but we would prefer settling on a solution before the first release (sometime within the next 2 weeks).

bolotskydev:
Hey there guys, sorry for the delay, I'm taking over this task as @SilencerWeb is no longer available. Currently getting the context and heading for a UX-wizardry to our creative dep 🚀

inancgumus:
@bolotskydev, thanks, I think it looks great!

ankur22:
@bolotskydev looks good to me too 👍

bolotskydev:
@imiric @inancgumus alright guys, feel free to check this [out on staging](https://mdr-ci.staging.k6.io/docs/refs/pull/674/merge/javascript-api/k6-http/response/)! 

**A couple of notes:**
1. Current version is almost ready to be published, I am still working on little things, but decided to show you anyway because this is very close to how the final result is going to look and feel
2. The very first table still has to be removed, it is there purely for dev purposes
3. The second (original) got wrapped inside `TableWithNestedRows` to feel what is like to work with the real content

**Let's talk about dev contract as well 😄 Inside the real content I've spotted following problems:**
1. `Parameter` column values sometimes represented as `inlineCode` elements. This adds some complexity to the table logic as we have to modify the inner content by stripping down some '.' pieces getting down to nested parts. So whilst it is absolutely possible to keep those (I made it backwards compatible), it is being rendered as a plain string. ⚠️ You still can use links in `Parameter` column, but that's just about it. Please, tell me if you need to keep the ability to use `inlineCode` in `Parameter` column.
2. If you'd [explore the changes to content file](https://github.com/grafana/k6-docs/pull/674/files#diff-6dcf855f37267dbbd0f81090e93e641ec32c2b1825085565ebd3eb81b5da2ff1R31), you'd see I've added some additional rows. Due to how currently table works, one can not omit any intermediate entries. Let me give an example:

```yaml
// Parameter column values, will work just fine
prop
prop.nest
prop.nest.more
```
```yaml
// will throw an error due to skipped 'prop.nest'
prop
prop.nest.more
```

If you absolutely sure this is necessary, I can think of workaround, but this, again, will contribute to overall logic complexity quite significantly.

3. I haven't changed the contributing file format just yet, but you bet I will once we agreed the task is completed and the dev contract is satisfying.

inancgumus:
Thanks, it looks great! 🎉 

> 1. `Parameter` column values sometimes represented as `inlineCode` elements.

I think that shouldn't be a problem 👍 

> 2. ... will contribute to overall logic complexity quite significantly.

Agreed 👍  It's OK as is.

> 3. One of the `Parameter` values goes like this: [`Response.timings.looking_up` (currently unavailable)]. The second part in brackets is somewhat problematic as parser operates on the very first string (usually dotted property path).

Do you mean "the second part is in _paragraphs_"? If so, we don't add parentheses like that, at least, not in the xk6-browser docs. We add an icon instead as in #650.

> 4. Only 3 level nesting is supported.

AFAIK, that should be enough 👍

bolotskydev:
> Do you mean "the second part is in _parentheses_"? If so, we don't add parentheses like that, at least, not in the xk6-browser docs. We add an icon instead

Yep, the parentheses! 😅 

Cool beans then, let me make sure the parser works with tooltips, remove content changes, and make a new section for the nested table component in contributing file format before we can merge this PR 🙆‍♂️

bolotskydev:
Latest changes:
- added tooltip support (place it anywhere in Parameter column, but prefer right side because of toggler icon)
- changed down font size of tooltip emoji's - 20x20 way too much, made it the same as the mentioned toggler, looks nice!
- added another sample table to poor `Response` page 
- Parameter column is now almost bulletproof, it is able to handle inline code, links, tooltips, plain text, goes easy on the order of these things and if something bad happened instead of throw renders `Invalid row markup` in place

inancgumus:
LGTM, thanks, @bolotskydev 🎉

inancgumus:
LGTM, thanks, @bolotskydev 🎉

bolotskydev:
Alright, I believe I'm done here. Please, pay attention to the changes in `61 Response` content file where I've removed all dev tables and made use of new component, contributing format file and, once again, [staging](https://mdr-ci.staging.k6.io/docs/refs/pull/674/merge/javascript-api/k6-http/response/). Feel free to merge afterwards!

ppcano:
@ankur22, this is ready for merging. Feel free to merge it and test it on your incoming changes.

imiric (approved):
Awesome work @bolotskydev, thanks! :clap:

This looks like a very neat solution from a UI perspective, even if the code is a bit too magic for my taste :sweat_smile:

Do you want to resolve the current merge conflict or should we?

bolotskydev:
@imiric Yep, you're completely right about magic code! This nested table built in a robust, but hacky way, which is inevitable as we need to represent completely flat table from MDX perspective as 2-levels nested in UI and make sure nothing breaks when using other custom components and markup elements. I was testing the native `table` nesting to fight the render logic abundance, but then the copywriting experience suffered and I still had to introduce severe CSS crutches to make it look sane, not much of a win. 

The conflict lies in the content area, so I didn't touch it, because I am not really sure whether we want to merge the whole thing with `Response` file using the new component, or do it later and just introduce the component to codebase. If you give me a hint on this, I will resolve the conflict. 👌

imiric:
@bolotskydev We can do it later once this is merged, so you can remove it from here. Thanks!

bolotskydev:
@imiric done!

ankur22 (approved):
I've had a play around with this and it looks great! Good work!

PR #684, opened 2022-05-24
Added example of parametrized exec function with multiple scenarios

paloliska:


src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/02 Advanced Examples.md
```diff
@@ -175,3 +175,79 @@ export function apitest() {
 ```
 
 </CodeGroup>
+
+## Different flow for each scenario with parameter
+
+Using multiple separate exec functions for each scenario cannot leverage usage of passing data form setup() function called before test begin.
```
mstoykov, line 181: Hi @paloliska, 

There seems to be some confusion because this script 
```
export const options = {
  scenarios: {
    two: {
      executor: "shared-iterations",
      exec: "two",
    },
    some: {
      executor: "shared-iterations",
      exec: "some",
    }
  }
}

export function setup() { return { "something ehre": "value" } }
export function some(data) { console.log(JSON.stringify(data)) }
export function two(data) { console.log(JSON.stringify(data)) }
```
Definitely seems to have access to the data returned by setup. 

Can you please explain what you tried and what didn't work as I seem to be misunderstanding what isn't working or perhaps the experiment you did had some other issue.

paloliska: What I experieced this will not run as it is missing default function:
`export function setup() { return { "something ehre": "value" } }
export function some(data) { console.log(JSON.stringify(data)) }
export function two(data) { console.log(JSON.stringify(data)) }`

And with following, data are not passed to function 'two':
`export function setup() { return { "something ehre": "value" } }
export default function some(data) { console.log(JSON.stringify(data)) }
export function two(data) { console.log(JSON.stringify(data)) }`

paloliska: > And with following, data are not passed to function 'two':
running k6 with `docker run --rm -igrafana/k6 run - <load-test.js`

mstoykov: Hi sorry for the really slow reply, 

But for all of those it *does* work for me.

Can you please open an issue on the [k6 repo](https://github.com/grafana/k6) if you are still experiencing this.

mstoykov:
I will close this as the information documented here is either false or a bug that I can not reproduce. And if it is reproducible it should be fixed not documented

PR #686, opened 2022-05-30
Restructures options page

MattDodsonEnglish:
This restructuring aims to make the "Options" docs easier to read, scan, and find desired information.

The old options reference was quite long, and it mixed a few different topics. There was information about the order of precendence, and examples of different executions, and the list of all options. It didn't say anything about *Why* a user may prefer to set options in one place or another.

This PR breaks the options docs into multiple pages. A brief, top-level overview points to two new pages, one page about how to use options in practice, and another page with the reference information.

Splitting the options reference into such topics should also make it easier to add more content about options in the future. Now that the page is split into broad topic types, [^1] the writer should have to think less about where to add new information about options.  This also might help create a template for further work. Long term, each aspect of the k6 documentation should have a similar structure (not necessarily this structure, but something that readers could recognize when looking across different pages).

There are still some issues with the how-to document: really, the section "Where to set options" is more geared towards understanding. It would probably be better suited in a separate page called something like "About setting options." Furthermore, the examples of how to set options have little task orientation. Honestly, it's still more of a reference. It'd be better to add a little context about what the reader would be doing if they followed these examples.  

Nevertheless, this should be an improvement upon the original, and offers an experiment to look at for further restructuring.

[^1]: https://www.writethedocs.org/videos/eu/2017/the-four-kinds-of-documentation-and-why-you-need-to-understand-what-they-are-daniele-procida/


MattDodsonEnglish:
Hey @ppcano I have looked over the PR in staging, and I think it looks respectable. I'm worried about the redirects, though. Shouldn't this URL now redirect to /using-k6/k6-options/reference? Maybe I have misconfigured the redirect.

https://mdr-ci.staging.k6.io/docs/refs/pull/686/merge/using-k6/options

src/data/markdown/translated-guides/en/02 Using k6/05 k6 Options/01 How to.md
```diff
@@ -0,0 +1,166 @@
...
+  http.get('http://test.k6.io/');
+}
+```
+
+</CodeGroup>
+
+### Options in a config file
+
+<div id="config-json-example">
+You can also define the same options through a config file:
```
sniku, line 79: Creating `config.json` file is not enough. We must show how this config is read in the script.

```
export const options = open('config.json')
```

src/data/markdown/translated-guides/en/02 Using k6/05 k6 Options/01 How to.md
```diff
@@ -0,0 +1,166 @@
...
+```
+
+```bash
+PS C:\k6> $env:K6_NO_CONNECTION_REUSE=true; $env:K6_USER_AGENT="MyK6UserAgentString/1.0"; k6 run script.js
+
+PS C:\k6> k6 run --no-connection-reuse --user-agent "MyK6UserAgentString/1.0" script.js
+```
+
+</CodeGroup>
+
```
sniku, line 138: Another popular way of getting options is to set them on CLI as k6 variables. (not the same as env variables).

```
k6 run script.js -e MY_USER_AGENT="hello"
```

and then in the script: 

```
import http from 'k6/http';

export const options = {
    userAgent: __ENV.MY_USER_AGENT
};

export default function () {
  http.get('http://test.k6.io/');
}

```

This is very flexible as it allows users to add more logic in how they configure the execution

MattDodsonEnglish:
@sniku and @ppcano ,

Now I'm thinking of moving everything about setting options through `--config` in "hidden" documentation. I've explained my reasoning in issue https://github.com/grafana/k6-docs/issues/688.

However, I think that can happen after this PR is merged. Thoughts?

na-- (changes requested):
:+1: for splitting the explanation and the reference into separate pages, :-1: for removing documentation about the config files. I don't think it adds a lot of clutter or complexity, especially when the explanation is now a separate page, see my arguments in https://github.com/grafana/k6-docs/issues/688#issuecomment-1148386683

Removing the explanation for the config file from the docs when users can see its existence when they can type `k6 --help` or see this new section on multiple cloud accounts just adds confusion and will lead to questions on the community forum, not less... :disappointed:

MattDodsonEnglish:
> 👍 for splitting the explanation and the reference into separate pages, 👎 for removing documentation about the config files. I don't think it adds a lot of clutter or complexity, especially when the explanation is now a separate page, see my arguments in [#688 (comment)](https://github.com/grafana/k6-docs/issues/688#issuecomment-1148386683)
> 
> Removing the explanation for the config file from the docs when users can see its existence when they can type `k6 --help` or see this new section on multiple cloud accounts just adds confusion and will lead to questions on the community forum, not less... 😞

@na-- As a compromise, how about just putting the "Set options with --config" section in a <Collapsible> in the new how-to page?

And using this for the diagram.
image

I'm still in favor of simply removing reference to it, but I'm more in favor of just getting something up, so we can continue to finess these Options docs, which are among the most important of all k6 docs.

na--:
The dotted line border and the out-of-box text probably draw more attention to the config file box, not away from it :sweat_smile: Besides, I said in the issue, I am not sure we should actually discourage usage of the config files :thinking:

MattDodsonEnglish:
@na-- I brought it back and fixed the diagram.

https://github.com/grafana/k6-docs/pull/686/commits/acf3769c5bef49b33729bab11561cfd2c4cef263

gatsby-node.js
```diff
@@ -1291,6 +1291,12 @@ const createRedirects = ({ actions }) => {
     isPermanent: true,
   });
 
+  createRedirect({
```
na--, line 1294: hmm this redirect doesn't seem to be working :confused: See this for example: https://mdr-ci.staging.k6.io/docs/refs/pull/686/merge/using-k6/options/#console-output (with current docs: https://k6.io/docs/using-k6/options/#console-output)

MattDodsonEnglish: As far as we can tell, the redirects only work in production and on staging.k6.io.

Mihail investigated with a different redirect:

> https://mdr-ci.staging.k6.io/docs/refs/pull/686/merge/using-k6/options/#dns this doesn't work
https://k6.io/docs/cloud/project-and-team-management/team-members/#managing-members - works in production (it replaces team-members with `members
https://mdr-ci.staging.k6.io/docs/refs/pull/686/mergecloud/project-and-team-management/team-members/#managing-members also doesn't work

MattDodsonEnglish:
Let's see if the redirects work in staging. 👀 🤞

imiric:
Hey, I only took a cursory glance at the changes here and the discussion in #688, but did we intentionally skip mentioning the config file in the [Order of precedence](https://k6.io/docs/using-k6/k6-options/how-to/#order-of-precedence) section?

2022-06-08-113914_635x190_scrot

It goes from Defaults (1) to script options (2), but the graph shows `--config` as second step.

Just wondering if this was agreed or forgotten, as it currently looks inconsistent.

MattDodsonEnglish:
Oh yes, thanks. I'd removed that for reasons documented in issue https://github.com/grafana/k6-docs/issues/688#issuecomment-1148386683, then I forgot to put it back in.

Fixing now...

PR #693, opened 2022-06-09
fix: link HAR converter

Maksimall89:


javaducky (approved):
LGTM! Thank you for the correction!

Issue #694, opened 2022-06-10
Add markdown and CSS support for description lists


MattDodsonEnglish:
The `<dl>` HTML element&mdash;a.k.a description/definition list&mdash;can be a surprisingly useful tool in technical documentation. It provides an accessible way to make term lists, and it's a generally good way to add structure to a text when a writer needs more than bullets and less than headings.

 In the k6 docs, I can think of three places where it already would come in handy:

- In the [Question and Answer section](https://k6.io/blog/private-load-zones/#faqs-about-private-load-zones) of the private load zones blog.
- In the "[Where to set options](https://k6.io/docs/using-k6/k6-options/how-to#where-to-set-options)" section of the Options doc
- In the [Glossary](k6.io/docs/misc/glossary)

Currently, our site generator doesn't support description lists. We can write them in Raw HTML, but it's tedious and doesn't allow in-line markdown. Also, there's no CSS styling, so the writer must add further tags (like `<strong>`).

![image](https://user-images.githubusercontent.com/47385188/173024564-6b565ce4-c049-47f0-9445-e0f38c619675.png)

I propose two fixes:

- [ ] Add markdown syntax for DLs

  [There's already a semi-official syntax in extended markdown.](https://www.markdownguide.org/extended-syntax/#definition-lists). 
  I'd prefer using what is already adopted.
   ```markdown
   term
   : definition line one
   : definition line two
   ```

- [ ] Add a little CSS support for the <dl>

  This doesn't need to be fancy at all. Maybe, bold text for the term, then indentation for the description.
  
  I slightly modified the [default example from the MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) to have multiple paragraphs. This seems more or less good enough?
  
  
![image](https://user-images.githubusercontent.com/47385188/173028204-c5922da6-978f-44b3-a193-13c405eb3431.png)

MattDodsonEnglish:
The list of features for the script builder is another good place for a `<dl>`, as is, perhaps, the top-level page of the cloud docs.
https://k6.io/docs/cloud/creating-and-running-a-test/script-editor/


I noticed that the MDN docs frequently use definition lists to good effect. Here's one example: https://developer.mozilla.org/en-US/docs/Web/HTTP.

PR #698, opened 2022-06-13
WIP: programmatically builds Glossary w/ JS script and JSON data

MattDodsonEnglish:
This is more of a proof of concept over anything else at this point.

Writing a glossary by hand in markdown has some problems:
- It's boring and repetitive to write
- So much repetition makes it easy to format improperly
- It doesn't let us use a description list, probably the ideal HTML element for a glossary
  - Writing a description list by hand is even more error-prone
- If we want to reuse the glossary in multiple places, we have to copy and paste
   - Possible areas for Reuse would be at the top of a reference (print the definition) or in a style guide (with notes on usage).

To solve these issues, this branch experiments with programmatically creating the glossary page:
- it writes definitions in `src/data/glossary/gloss.json`
- a script, vocab.json, parses the JSON file into definition list and writes the list to a file.

However, there are many, many issues to fix before this is a usable solution:

- [ ] The definitions themselves need to be proofread
- [ ] The definitions need to be vetted and approved by experts, and cross-referenced with Grafana's technical vocabulary
- [ ] I need to figure out how to add a "Read more" field only if the `see_also:` property in the JSON file is non-empty
- [ ] The code is likely really, really bad. I would be shocked if it wasn't. I would put on socks, rub them on a carpet, and touch a doorknob, just to reduce the shock.
- [ ] The markdown links in the definitions are broken in in-line HTML
  -  This is probably worth waiting for better extended-markdown support: https://github.com/grafana/k6-docs/issues/694
- [ ] There needs to be an introductory list, using the `<Glossary>` function.
- [ ] It would be better to use YAML instead of JSON, for increased writer comfort
- [ ] Reading the datafile needs to be incorporated into the build automation:
   - either with the CI
   - Or more likely, using [Gatsby's source from YAML feature](https://www.gatsbyjs.com/docs/how-to/sourcing-data/sourcing-from-json-or-yaml/)

### DIFF
--- a/src/data/glossary/gloss.json
+++ b/src/data/glossary/gloss.json
@@ -1,0 +1,247 @@
+[
+  {
+    "term": "Concurrency",
+    "definition": "When multiple computations happen at the same time.\nIn the context of k6, virtual users can make concurrent requests as a test runs. In k6 cloud, you can also run multiple tests concurrently."
+  },
+  {
+    "term": "Correlation",
+    "definition": "The process of capturing dynamic data received from the system under test and reusing the data in subsequent requests. A common use case for correlation is retrieving and reusing a session id, or token, throughout the whole lifetime of a [virtual user](#virtual-user).",
+    "see_also": [
+      "[Correlation and dynamic data example](https://k6.io/docs/examples/correlation-and-dynamic-data/)",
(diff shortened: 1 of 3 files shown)


MattDodsonEnglish:
Irrelevant now that https://github.com/grafana/k6-docs/pull/700 exists

PR #708, opened 2022-06-27
Update example for Page in docs

mnayak06:
This PR updates the example mentioned in 'page' docs: https://k6.io/docs/javascript-api/xk6-browser/page/

In the example, the login credentials are set without waiting if login page is loaded or not.
The updated example waits till the login page loads and then tries to login with credentials.

### DIFF
--- a/src/data/markdown/docs/30 xk6-browser/01 xk6-browser/09-page.md
+++ b/src/data/markdown/docs/30 xk6-browser/01 xk6-browser/09-page.md
@@ -121,7 +121,6 @@ import launcher from 'k6/x/browser';
 export default function () {
   const browser = launcher.launch('chromium', {
     headless: false,
-    slowMo: '500ms', // slow down by 500ms
   });
   const context = browser.newContext();
   const page = context.newPage();
@@ -140,7 +139,7 @@ export default function () {
   page.$('input[type="submit"]').click();
 
   // Wait for next page to load
-  page.waitForLoadState('networkdidle');
+  page.waitForNavigation();
 
   page.close();
   browser.close();
--- a/src/data/markdown/docs/30 xk6-browser/01 xk6-browser/09-page.md
+++ b/src/data/markdown/docs/30 xk6-browser/01 xk6-browser/09-page.md
@@ -131,6 +131,9 @@ export default function () {
   const elem = page.$('a[href="/my_messages.php"]');
   elem.click();
 
+  // Wait for login page to load
+  page.waitForLoadState();
+
   // Enter login credentials and login
   page.$('input[name="login"]').type('admin');
   page.$('input[name="password"]').type('123');


imiric (changes requested):
Hey, thanks for this!

It does make sense to `waitForLoadState()` at that point, even though there was likely sufficient time for the `click()` to run and navigation to happen for the `page.$()` below it to work properly. We shouldn't rely on that to always be true, so it makes sense to wait at that point.

I'm curious though, did you get any failures without it?

And while we're changing this example, I think we should fix the `page.waitForLoadState('networkidle')` on line 143. That _does_ fail for me consistently, and in [the same example in the xk6-browser repo](https://github.com/grafana/xk6-browser/blob/eb136726e37282fffad0c642062a04a6680a6adc/examples/fillform.js) we changed it to `page.waitForNavigation()`, which works better in this case.

BTW, if you don't mind, feel free to propose this same change in the xk6-browser repo.

ankur22 (approved):

src/data/markdown/docs/30 xk6-browser/01 xk6-browser/09-page.md
```diff
@@ -131,6 +131,9 @@ export default function () {
   const elem = page.$('a[href="/my_messages.php"]');
   elem.click();
 
+  // Wait for login page to load
+  page.waitForLoadState();
+
```
inancgumus, line 136: Hi, thanks for sending the PR. I'd prefer using `page.waitForNavigation()` instead of `page.waitForLoadState()` as the first one will wait until the next page is navigated and loaded.

inancgumus (changes requested):

src/data/markdown/docs/30 xk6-browser/01 xk6-browser/09-page.md
imiric: I tried that, but it won't actually work at this point and `waitForNavigation()` will timeout, probably because of the race condition between `click()` and `waitForNavigation()`, since we're not handling this with [`Promise.all()` like Playwright](https://playwright.dev/docs/api/class-page#page-wait-for-navigation).

`waitForLoadState()` seems to be harmless here though, but I'd like to confirm with @mnayak06 whether they experienced issues without it.

inancgumus: Ah, OK. It works on my side if I remove `slowMo`.

imiric: Yeah, this is definitely racy. :disappointed: For me it doesn't work with or without `slowMo`.

Though we should also remove it from this example, since it's not needed.

@mnayak06 Can you do that change as well?

To summarize:
- replace `page.waitForLoadState('networkidle')` on line 143 with `page.waitForNavigation()`.
- remove `slowMo` from the `launcher.launch()` options.

Essentially make it the same as [the example in the xk6-browser repo](https://github.com/grafana/xk6-browser/blob/eb136726e37282fffad0c642062a04a6680a6adc/examples/fillform.js), while also adding your `page.waitForLoadState()` change.

mnayak06: I have committed the changes in 31696ee 
Also, the error which I got before adding `page.waitForLoadState()`
```
$ ./xk6-browser run browser-test.js

          /\      |‾‾| /‾‾/   /‾‾/
     /\  /  \     |  |/  /   /  /
    /  \/    \    |     (   /   ‾‾\
   /          \   |  |\  \ |  (‾)  |
  / __________ \  |__| \__\ \_____/ .io

  execution: local
     script: /browser-test.js
     output: -

  scenarios: (100.00%) 1 scenario, 1 max VUs, 10m30s max duration (incl. graceful stop):
           * default: 1 iterations for each of 1 VUs (maxDuration: 10m0s, gracefulStop: 30s)

ERRO[0004] cannot query element for selector ("input[name=\"login\"]"): element handle cannot evaluate: cannot call function on expression ("\n\t\t(node, injected, selector) => {\n\t\t\treturn injected.querySelector(selector, node || document, false);\n\t\t}\n\t\n//# sourceURL=__xk6_browser_evaluation_script__\n") in execution context (3) in frame (CE304BA63B5A935D34E9C40A8CAD2411) with session (1C8842F98F863E84295E74E6EEA5CD2C): Cannot find context with specified id (-32000)
        at reflect.methodValueCall (native)
        at file://browser-test.js:14:3(44)
        at native  executor=per-vu-iterations scenario=default source=stacktrace

running (00m03.6s), 0/1 VUs, 1 complete and 0 interrupted iterations
default ✓ [======================================] 1 VUs  00m03.6s/10m0s  1/1 iters, 1 per VU
ERRO[0004] err:websocket: close 1006 (abnormal closure): unexpected EOF  category="Connection:handleIOError" elapsed="0 ms" goroutine=54

     browser_dom_content_loaded.......: avg=171.48ms min=50µs     med=171.48ms max=342.92ms p(90)=308.63ms p(95)=325.77ms
     browser_first_contentful_paint...: avg=332.06ms min=332.06ms med=332.06ms max=332.06ms p(90)=332.06ms p(95)=332.06ms
     browser_first_paint..............: avg=332.06ms min=332.06ms med=332.06ms max=332.06ms p(90)=332.06ms p(95)=332.06ms
     browser_loaded...................: avg=112.52ms min=332µs    med=365µs    max=336.89ms p(90)=269.58ms p(95)=303.24ms
     data_received....................: 27 kB 7.3 kB/s
     data_sent........................: 985 B 273 B/s
     http_req_connecting..............: avg=635ms    min=0s       med=0s       max=1.9s     p(90)=1.52s    p(95)=1.71s
     http_req_duration................: avg=909.8ms  min=227.62ms med=313.77ms max=2.18s    p(90)=1.81s    p(95)=2s
     http_req_receiving...............: avg=228ms    min=224ms    med=230ms    max=230ms    p(90)=230ms    p(95)=230ms
     http_req_sending.................: avg=0s       min=0s       med=0s       max=0s       p(90)=0s       p(95)=0s
     http_req_tls_handshaking.........: avg=560.66ms min=0s       med=0s       max=1.68s    p(90)=1.34s    p(95)=1.51s
     http_reqs........................: 3     0.831159/s
     iteration_duration...............: avg=3.6s     min=3.6s     med=3.6s     max=3.6s     p(90)=3.6s     p(95)=3.6s
     iterations.......................: 1     0.277053/s
     vus..............................: 1     min=1      max=1
     vus_max..........................: 1     min=1      max=1
```

imiric (approved):
Thanks again!

inancgumus (approved):
LGTM. It works here. Thanks!

PR #715, opened 2022-07-01
Paul/rework extensions

javaducky:


ppcano:
@javaducky, a quick general review:

1 - `Build a k6 binary with extensions` and `Build bundle` articles refer to the topic. Both articles should not be separated into different sections but grouped somehow. 

2 - `Overview` is the new "extension" main page and the grid with the list of extensions is now unavailable. 

I think the grid with the extensions should be included on the main page. All the users navigating to extensions will quickly see the list and get an overview of the different technologies. Otherwise, they'll have to navigate to `Explore` and probably, many new visitors won't do it.

MattDodsonEnglish:
Thanks for all this work! I think it'll be much, much better.

Later in the week, I will go through each line for style and copy edits. For now, I just have structural proposals: 
- I agree that "Explore" could be on the top-level page. Perhaps, you could precede it with the section that is currently `#what-are-k6-extensions`. Then, instead of "Explore," I'd make an `h2` with a more descriptive title---something like "Explore all k6 extensions". A more descriptive heading will help both readers who scan the doc and robots that crawl the site.
- On the top page, there's no mention of `xk6` naming. How about mentioning this convention? A lot of people may see xk6 and wonder.
- With the real estate you save by moving "Explore" to the top level, you could add a section called "About xk6". Here, you could move all the text "What is k6" and "use cases". You could also use this page to _completely nerd out_ at the end, writing about custom binaries or internal runtime particularities, as much useful detail to your heart's content.
- I agree that the "Build a Binary section" could precede the "Builder" page. By the way, how about calling that page something like "Interactive Builder"? This way, readers would know what they get. I'm not necessarily saying "Build a Binary" and "Interactive Builder" should be on one page; I also think it would work to have one page after the other, or make a little dropdown.
- "Interactive builder" should also have at least one example to start with, just so readers can see the syntax.
- Are these "Create" initial tutorials repeatable? I will try to follow them at the end of the week. If they are repeatable, great! Our docs need this!
  - The only thing I would say about this is that perhaps the first sections could be split. The initial instructions require no background; anyone who could use the CLI could follow them. I would call these "Tutorials." The subsequent sections get into more technical detail. 
 Maybe you could make those sections the first guides?
 
 To wit: My proposal would have the sidebar look something like this.

ALL CAPS: sidebar title
A `-`: sidebar article
`--` A nested article
`#.+*` A section within an article.
 
 ```
 GETTING STARTED
 - Overview
   ## What are k6 extensions?
    <Mention xk6>
    ## Explore the xk6 extensions
- About k6 extension
  ## What is xk6?
  ## Use cases
  ## JS vs GO extensions
  ## Seriously heavy tech stuff [Maybe some other title]
- Build an extension
  ## Pre-requisites
  ## How to build a pre-existing extension
  -- Interactive builder  [Could be an h2 instead of a nested page]
- Tutorial: Build an extension
-- Build a JS extension
   ## What we're going to do
   ## Prerequisites
   ## Instructions
   ## What we just did
   ## See more
-- Build an output extension
  [Same structure as JS tutorial]
GUIDES
 - Advanced guide to building JS extensions
   ## The Go-to-JS bridge
   ## Advanced API module
   ## Accessing run-time state
 - Advanced guide to building Output extensions
  [Maybe there's not enough info yet]
```

Excuse me for this rather detailed review. There's not much new writing, but a lot of copy and pasting to new places. I am excited about being very clear with the structure, because the xk6 documentation set is relatively small. So, we have a chance to organize it using a known and recognized doc framework, like Diataxis. We could use these docs to try to be "perfect" with its structure. Maybe it could then serve as the model for future docs and rewrites. I also think such a structure would be a very sturdy structure for future xk6 documentation, even if it becomes quite lengthy.

Of course, I'll help with any of this as much as I can. There are also some possible sentence-level edits to make. I'll get to those later in the week.

**Comment by MattDodsonEnglish** [comment]
By the way, I like the presentation of the extensions and the interactivity. Whatever you've done in all those React files looks nice!

**Comment by javaducky** [comment]
> 1 - `Build a k6 binary with extensions` and `Build bundle` articles refer to the topic. Both articles should not be separated into different sections but grouped somehow.

This is pretty much unchanged from current documentation. My only change in this PR is to include a bit of context in the "frozen" portion of the page when scrolling through the list of extensions. My plan _was_ to modify the `Build a k6 binary with extensions` guide to be a much more detailed walkthrough with deep-dive on what xk6 does under the covers.

> 2 - `Overview` is the new "extension" main page and the grid with the list of extensions is now unavailable.
> 
> I think the grid with the extensions should be included on the main page. All the users navigating to extensions will quickly see the list and get an overview of the different technologies. Otherwise, they'll have to navigate to `Explore` and probably, many new visitors won't do it.

I think the grid, as currently implemented, is overwhelming unless we redesign that listing into a better-curated listing. I feel that landing on a "What are extensions?" with a bold callout to the listing is more user-friendly, IMO.

**Comment by javaducky** [comment]
> * how about calling that page something like "Interactive Builder"?

I would prefer that as well, but I know that we've shown very explicitly the "Build Bundle" name/links in videos. I was a bit hesitant to change that to avoid potential questions for folks trying to find that; may not really be an issue though.

**Comment by javaducky** [comment]
@ppcano, @MattDodsonEnglish as I noted, this is pretty much just the start. The amount of changes is rather large at the moment, that's why I thought I'd open this PR and then iterate with additional articles and updates.

In general, my thoughts were to have the main pages be somewhat of TLDRs, with Guides then being the deep dive.

**Comment by MattDodsonEnglish** [comment]
Sorry for the feedback onslaught :-( . But on the bright side, xk6 docs could really be leader to help us with more information restructuring.

> In general, my thoughts were to have the main pages be somewhat of TLDRs, with Guides then being the deep dive.

I think this is a good plan. In my opinion, there are two "secrets" to docs:
- Having a sensible, progressive disclosure of information, accomodating multiple levels of detail and reader motivation,
- Understanding what people look for a in technical documentation, and making the information architecture accomadate these needs. Basically, all people look to docs for one of these reasons:
  - To learn by copying from examples (tutorial)
  - To do some background reading (explanation)
  - To find procedures that they can follow in the course of their work (how to guides)
  - To just check values and parameters as quickly as possible (reference)
  
> I think the grid, as currently implemented, is overwhelming unless we redesign that listing into a better-curated listing. I feel that landing on a "What are extensions?" with a bold callout to the listing is more user-friendly

Now that you wrote this, I think I agree that it makes more sense to separate the extension list to its own page.
It relates to the concept of progressive disclosure. If the information arch supports clear wayfinding, I don't think any users will give up because they have to make one more click. And research by NN group has shown that the [3-click rule is false.](https://www.nngroup.com/articles/3-click-rule/).

Yes, the list should also be better chunked down, into at least supported/community, but that's another topic...

File: src/components/pages/doc-extensions/what-is-xk6/what-is-xk6.view.js [hunk 1]
```diff
@@ -0,0 +1,34 @@
...
+      What is xk6?
+    </Heading>
+
+    <p className={styles.description}>
+      <Link to="https://github.com/grafana/xk6/" className="link">
+        xk6
+      </Link>{' '}
+      is the command-line tool and framework inspired by{' '}
+      <Link to="https://github.com/caddyserver/xcaddy/" className="link">
+        xcaddy
```
ppcano, line 19: This article is the intro/homepage. I suggest removing the mention to the xcaddy project.

src/data/markdown/docs/07 extensions/01 Guides/02 bundling.md
```diff
@@ -3,63 +3,112 @@ title: 'Build a k6 binary with extensions'
...
+- With each `--with`, we specified a full GitHub URI for the extension repository. If not specifying 
+  a version, the default is `latest` once again. Check your extension repository for stable
+  release versions, if available, to lock in your version as we've done with `xk6-sql@v0.0.1`.
+- We did not use the `--output` option; therefore, our new binary is `k6` within the current directory. 
+  Had we used `--output k6-extended`, our binary would be named `k6-extended` within the current 
+  directory. If a directory is specified, then the new binary would be `k6` within 
+  _that_ directory. If a path to a non-existent file, e.g. `/tmp/k6-extended`, this will be the
+  path and filename for the binary.
+
+> Suppose now you've cloned the `xk6-sql` repository and want the local version included in your
```
ppcano, line 87: This is a frequent case when building or debugging extensions.

I'd create an H3 or H2 section so it is highlighted on the Table Of Content (right sidebar).  The section could be something like `Building an extension from a local repository`

ppcano (changes requested):
The PR changes a few URLs. It needs to set up the redirections.

https://github.com/grafana/k6-docs/blob/main/gatsby-node.js#L1294

Note that redirections only work when deployed to `staging` or `production.

javaducky:
> The PR changes a few URLs. It needs to set up the redirections.

@ppcano I went through the `gatsby-node.js` and it appears all the redirects will be correct from what I could tell. I had previously added some redirects, but not sure if you're seeing something broken that I'd missed.

MattDodsonEnglish:
I've made a PR to this branch with some language fixes: https://github.com/grafana/k6-docs/pull/727

In internal convo, a few snags in the tutorial instructions were discovered. Before publishing, we need to make sure no errors or missed steps happen so that all users who follow will succeed.

Beyond that, the only major thing I see to fix would be the JS lint error I created and any language issues I introduce :grimacing: .

-----

However, I think there will still be work to do to make the docs the best they can be. Here are some things I noted as I edited:

- The bundle builder and the list are almost the same content. I wonder if the pages could be combined.
- there's no information about the actual build syntax:

   ```sh
   xk6 <version> --with /path/to/external --with /path/to/local
   ```
- This admonition seems strange:
   > You should be committed to maintaining a public repository that keeps up to date
  > with the latest k6 APIs if you intend to share with the community.

  Why should I? Doesn't all open source software come without warranty? I'd just delete this. Or explain it better, like "We might remove links to unmaintained repos."
- The   "Go-to-Js bridge" explanation should not be in the tutorial. Nor should any material that follows that section, like "advanced API"
- Similarly, the tutorial intro says you should know about the go-to-JS bridge, but doesn't link.


The last bullets point to a need for more conceptual information. Since the doc set is quite light, I think a little sidebar called "Explanations" or "Concepts" would not take up too much real estate.  Some ideas:
- The Go-to-JS Bridge
- Tradeoffs of making an extension, in terms of performance and maintenance
  - Maybe the section I renamed "Avoid unneeded work" could go there too.
- Deep dives into how the application works, e.g. how does it detect that an extension name is unique, taking advantage of the Go API 


Again, these comments are just things to consider for future iterations. I can extract them to their own issue after.

ppcano:
> I went through the gatsby-node.js and it appears all the redirects will be correct from what I could tell. I had previously added some redirects, but not sure if you're seeing something broken that I'd missed.

@javaducky 

Two redirections are missing:

- /extensions/guides/create-an-extension/ -> /extensions/getting-started/create/
- /extensions/guides/what-are-k6-extensions/ -> /extensions/

Also:

- The `getting started` category is not consistent. Some articles are under `/getting-started` and others not.

- `*/bundle-builder` URL could be changed to `*/bundle` and create a new redirect

javaducky:
> Two redirections are missing:
> 
> * /extensions/guides/create-an-extension/ -> /extensions/getting-started/create/
> * /extensions/guides/what-are-k6-extensions/ -> /extensions/

Good catch! I'll get those added.

> * The `getting started` category is not consistent. Some articles are under `/getting-started` and others not.

Yeah...that was me trying not to break existing links. I'll make them consistent and add redirects for the old paths.

> * `*/bundle-builder` URL could be changed to `*/bundle` and create a new redirect

👍 

@MattDodsonEnglish has some changes coming to the verbiage, so I'll wait on these changes until those are merged in to avoid potential merge conflicts.

javaducky:
Squashed commits to make fixing/merging of #727 a bit easier. 😅

PR #734, opened 2022-07-20
Defines thresholds and checks more precisely

MattDodsonEnglish:


javaducky (approved):
LGTM!

Just wanted to note that we do have a broader inconsistency with how the JavaScript language is noted. Some places have "Javascript" and others have "JavaScript." Looking now, it _seems_ that the latter is the more correct.

Leaving this as a task for another issue to resolve.

src/data/markdown/translated-guides/en/02 Using k6/03 Checks.md
```diff
@@ -3,16 +3,23 @@ title: 'Checks'
 excerpt: 'Checks are like asserts but differ in that they do not halt the execution, instead, they just store the result of the check, pass or fail, and let the script execution continue.'
 ---
 
-[*Checks*](/javascript-api/k6/check/) are like assertions, but they don't halt execution.
-Instead, they store the result of the check, pass or fail, and let the script continue.
-If you want to halt execution, take a look at [thresholds](/using-k6/thresholds) (which can include checks).
+Checks are true/false conditions that evaluate the content of some value in the Javascript runtime.
```
MattDodsonEnglish, line 6:
```suggestion
Checks are true/false conditions that evaluate the content of some value in the JavaScript runtime.
```

src/data/markdown/translated-guides/en/07 Misc/09 Glossary.md
```diff
@@ -61,8 +61,7 @@ The number of simultaneous VU requests in a test run.
 
 ### Checks
 
-Conditions that validate the correctness of a service.
-In k6 scripts, a check validates any true/false condition in the test.
+Checks are true/false conditions that evaluate the content of some value in the Javascript runtime.
```
MattDodsonEnglish, line 64:
```suggestion
Checks are true/false conditions that evaluate the content of some value in the JavaScript runtime.
```

src/data/markdown/translated-guides/en/02 Using k6/03 Checks.md
```diff
@@ -3,16 +3,23 @@ title: 'Checks'
 excerpt: 'Checks are like asserts but differ in that they do not halt the execution, instead, they just store the result of the check, pass or fail, and let the script execution continue.'
 ---
 
-[*Checks*](/javascript-api/k6/check/) are like assertions, but they don't halt execution.
```
oleiade, line 6: Having worked quite a bit around checks and thresholds in the last months, I ALWAYS have to check which is halting the execution, and which is not.

From my perspective, this information is major, and should remain in the first sentence, maybe with a formulation that mirrors the one you rephrased about thresholds? 🙇🏻

MattDodsonEnglish:
@oleiade Interesting. I suppose continuing to run goes against the expectations many programmers will bring with them. As I am coming to everything with a clean slate, halting execution seems like an important technical detail, but not _the_ detail.

Putting key information up front is definitely good document design.

How about these openings:

> ## Checks
> Checks are true/false criteria for your test runtime values.
> If a check fails, k6 stores the result and the test continues to run.


> ## Thresholds
> Thresholds are pass/fail criteria for your test metrics.
> If a test metric does not meet the expectation you defined, the threshold fails.

What do you think?

I think that writers can expect _most_ web readers to read the first two sentences of a doc (if the sentences are short). After that, we can only _hope_ that they'll read the first sentences of the subsequent paragraphs. I'm basing this on research in the [F-shaped pattern of reading](https://www.nngroup.com/articles/f-shaped-pattern-reading-web-content/).

oleiade:
Perfect 👍🏻 

I'm guilty of reading in "diagonal" as we call it in french. A good technique to help guide the eye through documentation is to **bold** the key aspects. For instance, in a long and technical document I would write, I often **bold** picked sentences, that, brought together can be read as a TL;DR (<- those I'm not a fan of, and they're often put at the end of documents for some reason) 😸

PR #735, opened 2022-07-20
Add code and default value for websockets

mkerollos:
Status code currently missing from websockets close documentation.


mstoykov (approved):
Thanks for the contribution :bow:

PR #738, opened 2022-07-21
Fix gRPC Params object to account for headers deprecation in v0.37.0

ndrpnt:
`Params.headers` was deprecated in v0.37.0 and replaced by `Params.metadata` (see grafana/k6#2370).


codebien:
Hi @ndrpnt, thanks for your contribution :bow: 

@na-- after two releases I think we can consider removing `headers` entirely from the documentation. WDYT?

na-- (approved):
@codebien, if we had remembered to document it properly at the time, yes, that would make sense... :disappointed: Unfortunately, we seem to have forgotten, so let's have it like this for a version or two and then remove it. 

@ndrpnt, thanks! :bow:

codebien (approved):

PR #742, opened 2022-07-26
Update broken links to Options

saikdasari:


MattDodsonEnglish:
Thank you very much for getting rid of those pesky old links!

PR #752, opened 2022-08-03
Revert new layout/format changes in the xk6-browser docs

ankur22:
Closes: https://github.com/grafana/xk6-browser/issues/479

Some of the class documents have been changed to use subheadings for an experimental layout. We want to be able to merge `docs/582-xk6-browser` into `main` asap to allow out users to quickly see what has recently changed and also to correctly document the issues we have found. We've not had time to go through all the classes and document them as we would like. In those cases we need to revert back to the table format as it is in `main`.


src/data/markdown/docs/30 xk6-browser/01 xk6-browser/03-element-handle.md
```diff
@@ -3,7 +3,48 @@ title: "ElementHandle"
 excerpt: "xk6-browser: ElementHandle Class"
 ---
 
-<BrowserCompatibility/>
+## Supported APIs
```
imiric, line 6: Should we add a warning notice to all of these pages that they're a WIP?

src/data/markdown/docs/30 xk6-browser/01 xk6-browser/03-element-handle.md
```diff
@@ -34,84 +75,46 @@ export default function () {
...
-## elementHandle.waitForElementState(state[, options])
-## elementHandle.waitForSelector(selector[, options])
+<CodeGroup labels={["Check element state"]} >
+
+```javascript
+import launcher from 'k6/x/browser';
+import { check } from 'k6';
+
+export default function () {
+  const browser = launcher.launch('chromium', {
```
imiric, line 87: Reminder to update the examples here in #753 once this PR is merged.

imiric (approved):
:+1:

src/data/markdown/docs/30 xk6-browser/01 xk6-browser/03-element-handle.md
ankur22: I could add the `<BrowserCompatibility/>` back in but that references PW, so maybe something like:

```
<Blockquote mod="warning">

This is still work in progress, and some of the functionality that is listed below may not behave as expected.

</Blockquote>
```

imiric: Yeah, that text sounds good to me. Just with a slight change: "This documentation is a work in progress, and some ...".

I think we can delete that `BrowserCompatibility` component, we won't use it as is. Or if you prefer to avoid repeating that `Blockquote` on each page, rename the component to something like `BrowserDocsWIP` and use it instead.

ankur22: Resolved in: e075b018e915ff48e97b48d5d8567ee0f0e012cd & c01d309faa84683cc8275f5a2b677b3d4cdf073f

PR #753, opened 2022-08-03
Change docs to use chromium instead of launcher

ankur22:
Closes: https://github.com/grafana/xk6-browser/issues/465

The latest version of `xk6-browser` now exports a `chromium` type when it comes to launching and testing against browser. This is still the only supported browser. This api also matches PW. This PR changes the examples and docs to match the new way of launching the browser.

## Previous

```javascript
import launcher from 'k6/x/browser';

export default function () {
  const browser = launcher.launch('chromium', {
```

## New

```javascript
import { chromium } from 'k6/x/browser';

export default function () {
  const browser = chromium.launch({
```

### DIFF
--- a/src/data/markdown/docs/30 xk6-browser/01 xk6-browser/01 Browser.md
+++ b/src/data/markdown/docs/30 xk6-browser/01 xk6-browser/01 Browser.md
@@ -7,7 +7,7 @@ The `Browser` class is the entry point for all your tests, and it is what intera
 - [BrowserContext](/javascript-api/xk6-browser/browsercontext/)s which is where you can set a variety of attributes to control the behavior of pages;
 - and [Page](/javascript-api/xk6-browser/page/)s which is where your rendered site is displayed.
 
-A new Browser instance (hence a new browser process) can be created using the `launch()` function of the `'k6/x/browser'` module.
+A new Browser instance (hence a new browser process) can be created using the `launch()` method of the `chromium` module from `'k6/x/browser'`.
 
 | Method                                                                                    | Description                                                                                                                                           |
 |-------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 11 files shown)


src/data/markdown/docs/30 xk6-browser/01 xk6-browser/01 Browser.md
```diff
@@ -7,7 +7,7 @@ The `Browser` class is the entry point for all your tests, and it is what intera
 - [BrowserContext](/javascript-api/xk6-browser/browsercontext/)s which is where you can set a variety of attributes to control the behavior of pages;
 - and [Page](/javascript-api/xk6-browser/page/)s which is where your rendered site is displayed.
 
-A new Browser instance (hence a new browser process) can be created using the `launch()` function of the `'k6/x/browser'` module.
+A new Browser instance (hence a new browser process) can be created using the `launch()` function of the `chroimum` exported type from the `'k6/x/browser'` module.
```
imiric, line 10:
```suggestion
A new Browser instance (hence a new browser process) can be created using the `launch()` function of the `chromium` exported type from the `'k6/x/browser'` module.
```

imiric (changes requested):
:+1: Just a typo.

Also, let's merge #752, rebase this branch, and update the examples on those pages as well.

src/data/markdown/docs/30 xk6-browser/30 xk6-browser.md
```diff
@@ -19,13 +19,14 @@ If you're more adventurous or want to get the latest changes of the xk6-browser
...
-|--------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| launch(browserName, [options]) | The first parameter is required, and the only accepted `browserName` is currently `chromium`. The `options` parameter is optional. You can see a list of all possible `options` in [the options heading](#options). |
+| Method                     | Description                                                                      |
+|----------------------------|----------------------------------------------------------------------------------|
+| chromium                   | Import module from `'k6/x/browser'`                                              |
+| chromium.launch([options]) | You can see a list of all possible `options` in [the options heading](#options). |
```
imiric, line 29: It's a bit strange listing `chromium` under "Method". Technically, `import { chromium } from 'k6/x/browser';` should import the Chromium `BrowserType` implementation/object directly, and we should remove [our `launch()` wrapper that currently exists at the module level](https://github.com/grafana/xk6-browser/blob/aadaffec87c8ffba66db89f417bb68e3aacbbe08/main.go#L71-L87).

So we should probably add a new page for `BrowserType`, where we document `launch()` and all of its options. And on this page we should document the module itself, which has global attributes like `devices` and `version`.

I'll create issues for both of these things, and hopefully we'll have time to work on them in this cycle. They shouldn't take long.

imiric (approved):

Issue #745, opened 2022-07-29
Destructure cloud FAQ


MattDodsonEnglish:
The FAQ has a venerable role in the history of information on the internet.

However, as we've learned more about internet reading habits and hypertextual document writing, many web writers now disdain the FAQ. For example, Gov.uk (one of the best examples of great technical content for a variety of audiences) does not allow them.[^1] There are two central criticisms:

- FAQs offload the burden of content structure and organization onto the reader.
  
  FAQs put disparate topics in a single long list, rather than placing support questions where they are most relevant in the overall information environment.
- FAQs are hard to scan:

  Readers of English web documents scan their eyes down the left side of the document. FAQ headings almost always start with a relative noun and either an auxiliary verb to be or the verb to be(for example, "How does..." or "Why does...). So, the reader's eye lands on a phrase that conveys the grammatical time and mood of the sentence but bears almost no semantic content.
  
I'm not totally anti-FAQ: if someone wants to document something but doesn't have much time to focus on written content, FAQs are an easy, accessible medium. But, I think these criticisms apply to the k6 Cloud FAQ.

The only thing that gives me pause about destructuring the FAQ is that it is a popular page (relative to the cloud docs). However, analytics show that 72% of readers to the Cloud FAQ come from Google. For comparison, the core k6 documentation has only 50% of its traffic from Google.

## Supplementary reading:

For the curious, I'm copying this list of anti-FAQ articles from an article on the Canadian Government site[^2]:

- [No More FAQs: Create Purposeful Information for a More Effective User Experience](https://alistapart.com/article/no-more-faqs-create-purposeful-information-for-a-more-effective-user-experi/)
- [Why FAQ Pages Are Almost Always a Bad Idea (And What to Do About It)](https://thegood.com/insights/faq-pages/)
- [FAQs: why we don’t have them (Gov UK)](https://gds.blog.gov.uk/2013/07/25/faqs-why-we-dont-have-them/)
- [18F content guide - Don’t use FAQs (US government)](https://content-guide.18f.gov/structure-the-content/#dont-use-faqs)
- [Why You Shouldn’t Use a FAQs Page](https://turbofuture.com/internet/Why-You-Shouldnt-Use-a-FAQs-Page)
- [FAQs are the dinosaurs of web navigation - Gerry McGovern](https://gerrymcgovern.com/faqs-are-the-dinosaurs-of-web-navigation/) 

### Counterpoint

[...]

PR #764, opened 2022-08-10
Destructures cloud FAQ

MattDodsonEnglish:
Fixes: https://github.com/grafana/k6-docs/issues/745

- Moves reference information to Cloud reference (along with some other miscellaneous topics)
- PLZs and Troubleshooting to running tests
- Some language fixes (not all)

This moves useful content higher up the doc hierarchy, and splits the content to the places where it's most relevant.
As a side bonus, this extra reference makes it easier for future content additions and restructuring.

**One breaking change**: links to https://k6.io/docs/cloud/cloud-faq/general-questions/ will break. This page got split across three docs, so a redirect for each anchor would be too tricky. 

Before merging, I must:
- [x] Run a link checker on changed pages
- [x] Review images
- [ ] Proofread work
- [x] Check Cloud UI text for FAQ links, fix if needed.

@ppcano I think I can handle this independently, but I'm tagging you just so you see what is incoming. @Vmihajlovic , these changes may move some info that you look at and link to often to other places.


ppcano:
Some small inputs:

- The `Cloud references` category is similar to `Misc` category on the OSS guides. I suggest testing whether both docs could be consistent in the naming.

- Instead of having a 404 page for the old `cloud-faq/general-question` page, I think it is better to have a redirect to a general page; for example, either the Cloud Docs main page or the `Cloud references` category page.

Vmihajlovic:
Just a few notes on my end as well:

1. I know that we are linking to specific documentation articles directly in the web app, i.e., Performance Insights links
2. not sure if we want to highlight which features are not available by default (similar to Private Load Zones). Namely PDF export and Usage Reports; they are available only in Annual Pro and above. Naturally, this is the current state of things, and updating these references as we change plans/pricing/models in the future might prove to be cumbersome. Perhaps a stop-gap solution would be to link to the pricing page.

MattDodsonEnglish:
> 1. I know that we are linking to specific documentation articles directly in the web app, i.e., Performance Insights links. We should make sure they continue working after we merge these changes.

Great point, Viktor, thanks! I'm going to grep around the cloud repo, and I've added that to the publication checklist. Even if I don't find any breaks this time, more changes are likely to come. 

> Instead of having a 404 page for the old cloud-faq/general-question page, I think it is better to have a redirect to a general page;

Definitely. Should have thought of that. Fixed.

> The Cloud references category is similar to Misc category on the OSS guides. I suggest testing whether both docs could be consistent in the naming.

Great question. Not sure if it's for the scope of this PR, but I support making the OSS docs have a distinct "Reference" section too. However the current "Misc" has a lot more miscellany than just reference.

For the sake of terminology, let's say that reference material is material that readers look at only to check certain values and see examples. A reference just tries to help a reader find information; assumes the reader will know what to do with it.

Let's review:

Misc
- [Usage collection](https://k6.io/docs/misc/usage-collection/) :heavy_check_mark:  Pure reference
[IntelliSense](https://k6.io/docs/misc/intellisense/) :question:  I don't even know what this is! Should this really be on the main docs page? Seems more for integrations.
[Fine tuning OS](https://k6.io/docs/misc/fine-tuning-os/) :x: No, definitely a deep how-to guide or explanation. One of the deepest dives we have. Great article, not reference.
[Archive Command](https://k6.io/docs/misc/archive-command/) :question:  Kind of a mixed reference, how-to, explanation. Not really a reference, because readers will go there to learn, not just to look.
[TypeScript template](https://github.com/k6io/template-typescript) :question:  I don't even know what this is! Should this really be on the main docs page? Seems more for integrations.
[Release notes](https://github.com/grafana/k6/releases) :heavy_check_mark: Pure reference
[Glossary](https://k6.io/docs/misc/glossary/) :heavy_check_mark: Pure reference

MattDodsonEnglish:
I made such a mess here that I'm opting for a clean restart:
https://github.com/grafana/k6-docs/pull/783

PR #770, opened 2022-08-12
chore: deleted dependabot config

bolotskydev:
**Describe the changes this PR brings**
This PR removes the dependabot config. It proved to be a very inefficient approach to handle dependency updates and creates too much noise.

⚠️  Also I've closed all pending dependabot PRs.


bolotskydev:
cc @ppcano , part of the task regarding DU system.

Issue #771, opened 2022-08-12
Slug creates extra slash in og:url metadata 


MattDodsonEnglish:
While running a link checker, I discovered a little bug in our socials metadata.
If a page frontmatter declares a slug that starts with a `/`, the URL for the twitter `og:url` ends up with two slashes between `https://k6.io/docs` and whatever the path of the slug is.

Proof:
1. Head to view-source:https://k6.io/docs/cloud/analyzing-results/checks/
2. `CTRL+f "og:url"`
3. find:
  > property="og:url" content="https://k6.io/docs//cloud/analyzing-results/checks/

I am guessing that it could be fixed by removing the slash in this line of `seo-view.js`:
https://github.com/grafana/k6-docs/blob/f00d3a0d4c15e0f89bd9727260099a4f95e081fb/src/components/shared/seo/seo.view.js#L48
https://github.com/grafana/k6-docs/blob/f00d3a0d4c15e0f89bd9727260099a4f95e081fb/src/components/shared/seo/seo.view.js#L102

## Counterpoint: What if people don't put a slash in a slug?

It's true: some people might do this.
But at the time of writing, every slug that has been added starts with a slash.
At least, that's what I determined with:

```sh
grep -r "^slug:" src/data/markdown
```

So I think that adding the slash is the intuive choice for writers in the majority of cases.

PR #772, opened 2022-08-12
Gatsby; removes inserted slash from og:url

MattDodsonEnglish:
Fixes: https://github.com/grafana/k6-docs/issues/771

I tested locally, as far as I can tell it fixes the double slash issue:

```html
<meta property="og:url" content="http://localhost:8000/cloud/analyzing-results/checks/" data-react-helmet="true">
```

The only issue is that it now will create invalid URLs if the slug doesn't start with a slash. I guess we could fix this by blocking invalid slugs in the CI, but every author so far has independently decided to start slug values at the site root, so it doesn't seem like a big issue.

### DIFF
--- a/src/components/shared/seo/seo.view.js
+++ b/src/components/shared/seo/seo.view.js
@@ -44,8 +44,8 @@ export const SEO = ({
     // eslint-disable-next-line no-nested-ternary
     slug && slug !== '*'
       ? slug.startsWith('jslib/')
-        ? `${docs}/javascript-api/${slug}`
-        : `${docs}${slug}`
+        ? `${docs}/javascript-api${slug}`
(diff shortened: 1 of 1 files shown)


bolotskydev:
Hey @MattDodsonEnglish , thank you for dealing with it! I see it is really important to resolve. As you rightly mentioned before, the solution is valid so far as every author puts the slash at the beginning of custom slug in frontmatter. I think we should enforce it behavior under the hood, making sure the `slug` we are working with always has a `/` in front. Let me extend a bit your solution and test it.

MattDodsonEnglish:
Sure, there's no rush at all. It's not a big deal in the scheme of things. The only reason I made a PR so quickly is so that I could pretend to be a JS developer :grin:

PR #778, opened 2022-08-25
Correction for integrating K6 to Datadog

sonali9696:
In cloud APM, we should select Datadog instead of Azure monitor as that's the service we are trying to integrate it with


MattDodsonEnglish:
Nice fix! Yes, that instruction was completely inaccurate. Thanks for making the docs more coherent :-).

Issue #678, opened 2022-05-20
Remove external content from guides sidebar


MattDodsonEnglish:
The majority of the sidebar links in the docs link to a doc. It is surprising and disorienting to select a link and go to an external, or non-docs, page. Current non-docs links are “Community”, “Release notes,” and “Typescript extension”.

Of course, these are useful links, but they seem out of place in the sidebar. Two proposed solutions:

a) Remove them from the sidebar, place them in the fat footer or in other doc pages
b) Remove them from the sidebar, place them on a page like “Guide to these docs”, which provides useful external links, and describes what the examples, jslib, cloud, etc pages contain.

I lean towards option B for now, but seems like an information architecture kludge. Longer term, it would probably be better to have this top-level map at some higher place in the site.

PR #779, opened 2022-08-29
Structure; replaces external links with infomap

MattDodsonEnglish:
Fixes https://github.com/grafana/k6-docs/issues/678

Some things to consider before merging:
- Is this information structure comprehensive, logical, and maintainable?
- Should we really excise "Release notes" from the sidebar?
  
  It went from the very bottom of `/docs` but on the sidebar, to a page at the very top of `/docs` but off the sidebar.
  I don't feel strongly about this either way.
- Is "k6 info map" really the best name? It doesn't thrill me, but I couldn't think of anything better:
  - "Site map" would be inaccurate, as many links are not to this site.
  - "Resources" seems okay, but some links on the page are called "resources" too. "Resources to resources" is ugly.
  - Finally, it is more or less a map of the k6 information environment. So it seems accurate.
  
This page could also use some aesthetic improvements. Right now it's bare and brutalistic. Besides some visuals, it also could perhaps present the information in a more digestible, interesting, or intuitive way. I experimented with tables and with the `<Glossary>` and `<DescriptionList>` components, but I didn't like how they looked. Maybe the best presentation would involve cards, like on the [Explore extensions](https://k6.io/docs/extensions/getting-started/explore/) page.

However, it might be worth saving visual improvements for a separate PR sometime in the future.


MattDodsonEnglish:
Alternative names: "Resource index", "Information index".

Issue #303, opened 2021-05-20
Show k6 extension type on Ecosystem page


imiric:
Since we now have more than one extension type it would be useful to show whether an extension is a `javascript` or an `output` type on the [Ecosystem](https://k6.io/docs/ecosystem/) page.

We'd need to add a new `type` property to [src/data/ecosystem/extensions.js](https://github.com/k6io/docs/blob/9e44322afddc3f1064f2300c0e2516d3ccf8ea58/src/data/ecosystem/extensions.js) for each extension. This could be rendered as simply "Type: JavaScript" in a small font under the extension name, but it would be nicer if we'd show small icons instead next to the name (e.g. `[JS]`, `[Out]`) to avoid cluttering each element.

In the future it would be good if extensions were filterable by type and category, but that can be done separately.

simskij:
Categories are already there, just not enabled yet. But yeah, having type would be great as well. @vi-nastya can you assist in implementing this? @lnikell can probably assist with what icons to use. I'm thinking something along the lines of `<>` and `->`, wdyt? I'll take care of assigning types to all the existing extensions once we've got things up and rocking.

Thanks 👍🏽

bolotskydev:
@imiric I see this is still relevant despite how old the feature request is. I'll mention you in PR with the necessary changes to fill up the extension types 👍

PR #802, opened 2022-09-12
feat: add categories & type for Extension card

Den-Bratchikov:
**Describe** what changes this pull request brings:
This pull request adds categories & type for Extension card for Docs.

**Steps to test**
1. Open the [page](https://mdr-ci.staging.k6.io/docs/refs/pull/802/merge/extensions/getting-started/explore/)
2. Confirm that everything looks and works as expected

<details>
<summary>Screenshots</summary>

Screenshot 2022-09-12 at 12 06 30

</details>

TODO: please remove test types in `extensions.json`

**References**
[Design](https://www.figma.com/file/qOPbJuR5wWjOablZeFCQtu/K6-Website-Design?node-id=10348%3A31996)

Closes: #303

### DIFF
--- a/src/data/doc-extensions/extensions.json
+++ b/src/data/doc-extensions/extensions.json
@@ -180,7 +180,7 @@
       },
       "official": false,
       "categories": ["Protocol"],
-      "type": "JS"
+      "type": false
     },
     {
       "name": "xk6-mqtt",
@@ -192,8 +192,7 @@
         "url": "https://github.com/pmalhaire"
       },
       "official": false,
-      "categories": ["Messaging"],
-      "type": "Out"
+      "categories": ["Messaging"]
     },
     {
       "name": "xk6-kv",
@@ -205,8 +204,7 @@
         "url": "https://github.com/dgzlopes"
       },
       "official": false,
-      "categories": ["Data"],
-      "type": "JS"
+      "categories": ["Data"]
     },
     {
       "name": "xk6-url",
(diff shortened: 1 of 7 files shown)


src/data/doc-extensions/extensions.json
```diff
@@ -217,7 +218,8 @@
         "url": "https://github.com/dgzlopes"
       },
       "official": false,
-      "categories": ["Misc"]
+      "categories": ["Misc"],
+      "type": "Out"
```
javaducky, line 222: This is not actually an output extension. For purposes of this PR, we can remove all uses of the "type" attributes for now given we see it is available.

javaducky (changes requested):
Design looks good to me!

My only ask is that we remove use of the `type` attribute in `extensions.json` for the moment. I will create a follow-up PR to validate the categories and types prior to publishing the docs.

MattDodsonEnglish:
I think it looks good! For future development, it'd be cool if there were also a way to filter by categories. 

Another thing, there's really two separate concepts, right?
- Extension _type_ is whether it's JS or Output
- Extension _category_ is for the kind of work the extension does.

Also, it'd be nice if the "misc" items had more descriptive category,

javaducky:
> For future development, it'd be cool if there were also a way to filter by categories.

💯 to that! Figured we'd write up an issue after this is completed.

> Also, it'd be nice if the "misc" items had more descriptive category,

I'll probably be doing this myself in a followup now that we'll have this.

Also for future, would love to see those items with the `official` attribute being `true` to have a star or something to show it is a Grafana-owned extension.

src/data/doc-extensions/extensions.json
Den-Bratchikov: Yes, i will remove all uses of the `type` attributes in `extensions.json`.

javaducky (approved):

PR #804, opened 2022-09-12
Fix xk6-browser class menu ordering

ankur22:
I noticed that the ordering of the classes in the docs were incorrect (not alphabetical `locator` was above `keyboard`) so i've just fixed that.


imiric (approved):
:+1:

Could we drop the numbers from the directory and file names, and just let the default (alphabetical?) order take place? Though we might need the numbers for some reason :thinking: :facepalm:

inancgumus (approved):

ankur22:
> 👍
> 
> Could we drop the numbers from the directory and file names, and just let the default (alphabetical?) order take place? Though we might need the numbers for some reason 🤔 🤦

I removed the numbering and it does work without them, but it looks like all the docs are using this numbered approach to force an order, so i'll leave it as is for consistency.

PR #814, opened 2022-09-23
Clarify xk6-browser running instructions

imiric:
We've had [reports on the forum](https://community.k6.io/t/issues-with-installing-xk6-browser-extension/4681) about confusions with running xk6-browser. The main issue was that the "Build from source" instructions build a `k6` binary, and the example mentions running `xk6-browser`. The `InstallationInstructions` element is not flexible enough to specify the output binary, so I chose to repeat the instructions inline instead.  We could also parametrize this via an `output` argument to the element, but I feel like we shouldn't complicate it too much, as there will often be slight changes of the instructions. It doesn't make much sense as a reusable element for me, anyway, and the need to use JSX/HTML instead of Markdown is cumbersome (this last point could probably be resolved with yet another Gatsby plugin).

The other change here is to clarify the `./` and `.\` prefix required on some systems, as that is another frequent source of confusion for users.

### DIFF
--- a/src/data/markdown/docs/30 xk6-browser/30 xk6-browser.md
+++ b/src/data/markdown/docs/30 xk6-browser/30 xk6-browser.md
@@ -75,9 +75,13 @@ If you have downloaded the pre-built binary you will find the binary named `xk6-
 ./xk6-browser run browser_test.js
 ```
 
-> Note the `./` prefix, which tells your shell to run the binary located in the current working directory. This is required on macOS and Linux, but not on Windows if you're using the `cmd.exe` shell. If you're using PowerShell, you should specify `.\xk6-browser` instead.
->
-> If you installed xk6-browser via a system package, or you've placed the binary in a directory that's part of your `$PATH` environment variable, then you can omit the `./` or `.\` prefix.
+<Blockquote mod="note" title="">
+
+Note the `./` prefix, which tells your shell to run the binary located in the current working directory. This is required on macOS and Linux, but not on Windows if you're using the `cmd.exe` shell. If you're using PowerShell, you should specify `.\xk6-browser` instead.
+
+If you installed xk6-browser via a system package, or you've placed the binary in a directory that's part of your `$PATH` environment variable, then you can omit the `./` or `.\` prefix.
+
+</Blockquote>
 
 ## Module Properties
 
--- a/src/data/markdown/docs/30 xk6-browser/30 xk6-browser.md
+++ b/src/data/markdown/docs/30 xk6-browser/30 xk6-browser.md
@@ -13,11 +13,28 @@ xk6-browser is currently being developed as a [k6 extension](/extensions). You h
 
 The quickest way to get started is to [download a release binary from GitHub](https://github.com/grafana/xk6-browser/releases).
 
+We suggest this option if you don't want to build your own binary with [xk6](https://github.com/grafana/xk6), which can be challenging in some cases.
+
 ### Build from source
 
-If you're more adventurous or want to get the latest changes of the xk6-browser extension, you can also build from source. 
+If you're more adventurous or want to get the latest changes of the xk6-browser extension, you can also build a binary from source.
 
-<InstallationInstructions extensionUrl="github.com/grafana/xk6-browser"/>
+To build a binary with the extension, first, ensure you have installed [Go](https://golang.org/doc/install) and [Git](https://git-scm.com/). Then run the following commands in a terminal:
+
+```bash
+# Install xk6
+go install go.k6.io/xk6/cmd/xk6@latest
+
+# Build the xk6-browser binary
+xk6 build --output xk6-browser --with github.com/grafana/xk6-browser
+
+... [INFO] Build environment ready
+... [INFO] Building k6
+... [INFO] Build complete: xk6-browser
+```
+
+xk6 will create the `xk6-browser` binary in the current working directory.
+
 
 ## Your First Test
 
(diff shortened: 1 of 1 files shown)


src/data/markdown/docs/30 xk6-browser/30 xk6-browser.md
```diff
@@ -55,9 +72,13 @@ export default function () {
 If you have downloaded the pre-built binary you will find the binary named `xk6-browser`. Using the [example](#example), create a new file (such as `browser_test.js`) in the same directory as the pre-built binary, and paste the example in that file. Now run:
 
 ```
-xk6-browser run browser_test.js
+./xk6-browser run browser_test.js
 ```
 
+> Note the `./` prefix, which tells your shell to run the binary located in the current working directory. This is required on macOS and Linux, but not on Windows if you're using the `cmd.exe` shell. If you're using PowerShell, you should specify `.\xk6-browser` instead.
+>
+> If you installed xk6-browser via a system package, or you've placed the binary in a directory that's part of your `$PATH` environment variable, then you can omit the `./` or `.\` prefix.
```
MattDodsonEnglish, line 80: Is this necessary information? My guess is that most people who want to browser test know this, especially if they know about $PATH variables.

But, I trust that the author knows the reader best.

I'm more open to leaving a note just for windows users, especially if `cmd.exe` and Powershell will fail with this syntax.

If we leave a note in, I suggest using the `<Blockquote>` Gatsby component to keep things consistent.

```
<Blockquote mod="note" title="A note for Windows users">

On Windows shells, the `./` file prefix will raise an exception.
Instead, try these:
- For `command.exe`, use no file prefix.
- In Powershell, use `.\xk6-browser`.

</Blockquote>
```

Will render as:

image

src/data/markdown/docs/30 xk6-browser/30 xk6-browser.md
```diff
@@ -13,11 +13,28 @@ xk6-browser is currently being developed as a [k6 extension](/extensions). You h
...
 The quickest way to get started is to [download a release binary from GitHub](https://github.com/grafana/xk6-browser/releases).
 
+We suggest this option if you don't want to build your own binary with [xk6](https://github.com/grafana/xk6), which can be challenging in some cases.
+
 ### Build from source
 
-If you're more adventurous or want to get the latest changes of the xk6-browser extension, you can also build from source. 
+If you're more adventurous or want to get the latest changes of the xk6-browser extension, you can also build a binary from source.
+
+To build a binary with the extension, first, ensure you have installed [Go](https://golang.org/doc/install) and [Git](https://git-scm.com/). Then run the following commands in a terminal:
```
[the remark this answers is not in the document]
MattDodsonEnglish: Putting instructions in ordered lists highlights that this is a procedure with a set sequence.

MattDodsonEnglish (changes requested):
LGTM, except my nitpicks.

src/data/markdown/docs/30 xk6-browser/30 xk6-browser.md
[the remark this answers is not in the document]
imiric: In general, I'd agree, but in this case I think it's better if we keep it consistent with the `InstallationInstructions` element.

Besides, it's a sequence of just 2 steps, and the first one is more of a pre-requisite we mention in passing. If we make it a standalone step, we might want to expand on how to confirm that Go and Git are installed properly, minimum versions, and things like that. Since this is rather off-topic for this document (and most of our installation instructions), I think it's fine if we leave this inline, as it was.

inancgumus (approved):

src/data/markdown/docs/30 xk6-browser/30 xk6-browser.md
imiric: Resolved by 41c7e883.

src/data/markdown/docs/30 xk6-browser/30 xk6-browser.md
```diff
@@ -55,9 +72,17 @@ export default function () {
...
 ```
-xk6-browser run browser_test.js
+./xk6-browser run browser_test.js
 ```
 
+<Blockquote mod="note" title="">
+
+Note the `./` prefix, which tells your shell to run the binary located in the current working directory. This is required on macOS and Linux, but not on Windows if you're using the `cmd.exe` shell. If you're using PowerShell, you should specify `.\xk6-browser` instead.
+
+If you installed xk6-browser via a system package, or you've placed the binary in a directory that's part of your `$PATH` environment variable, then you can omit the `./` or `.\` prefix.
```
MattDodsonEnglish, line 82:
```suggestion
The `./` prefix tells your shell to run the binary located in the current working directory. This is required on macOS and Linux, but not on the Windows `cmd.exe` shell. If you're using PowerShell, specify `.\xk6-browser` instead.

If you installed xk6-browser with a system package, or placed the binary in a directory that's part of your `$PATH` environment variable, you can omit the `./` or `.\` prefixes.
```

MattDodsonEnglish: Small language trimming. I don't think we need to add the word "Note" anymore, since the new `blockquote` component  expresses that already, even for screenreaders.

MattDodsonEnglish:
One other thing I noted while looking at the broader extension docs.

In those install docs, the procedures have an init command. For example:

> `go mod init xk6-compare`

We added this after I tried to reproduce the instructions and failed. Does the xk6-browser install need this init operation?

src/data/markdown/docs/30 xk6-browser/30 xk6-browser.md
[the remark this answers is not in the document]
imiric: I disagree, not all instructions have to be formatted as a list, particularly not 1 or 2 step ones like here. I find the Go and Git list in the "Before you start" section on the "Output Extensions" page overly verbose, for no practical gain, IMO.

But this is a style preference, and you've been working on the broader docs more than me, so I'll defer to your opinion.

Updating the `InstallationInstructions` element is out of scope for this PR, but I'll accept your suggestion here.

imiric:
> Does the xk6-browser install need this init operation?

No, this is only needed when creating a k6 extension project from scratch. In this case, we're only documenting how to build/run the extension.

MattDodsonEnglish (approved):

PR #815, opened 2022-09-26
Results output: minor suggestions

ppcano:


MattDodsonEnglish:
We had a minor merge conflict to resolve.

 For now, I'm leaving the example command with `summary-trend-stats` in. The idea is that it gives readers something to quickly copy and paste without having to jump to the options reference and figure out the syntax.
 
 But I don't think this is a crucial part of the doc, so I'm still open to deleting it.

PR #816, opened 2022-09-26
Fold result visualization docs in sidebar

MattDodsonEnglish:
All the folded items can be grouped under a single topic, "streaming".Besides emphasizing the logical organization, this also makes the sidebar section more balanced, since the current section has "End of test summary" as the lonely "outcast" of the group. 

This also frees up sidebar space, and provides a top-level page for users to look at different streaming options. The current docs only have this view at the bottom of the result-output doc, under a completely different sidebar heading.

TODO
- [x] Program in all redirects
- [ ] Test redirects on staging


MattDodsonEnglish:
**Do not merge** until we add redirects. Do not deploy until testing redirects in staging.

ppcano:
@MattDodsonEnglish, the current proposal is:

```
Results output

│
└───End of test summary
│   
│   
└───Real-time metrics
│   │  
│   └───Amazon Cloud Watch
│   │  
│   └───CSV
│   │  
│   └───JSON
│   │  
│   └───...
```

If we use the term `End of test` vs `Real-time`, I interpret real-time metrics as metrics that are accessible in real-time. But it is not true for the CSV and JSON options. Because of this, I disagree with this proposal.

Considering the `End of test/Real-time` categorization, I think an accurate proposal would be:

```
Results output

│
└───End of test
│   │  
│   └───Summary
│   │
│   └───Metrics
│             │  
│             └───CSV
│             │  
│             └───JSON
│   
└───Real-time metrics
│   │  
│   └───Amazon Cloud Watch
│   │  
│   └───....
```

I'd try to simplify with:




```
Results output

│
└───Summary
│   
└───Metrics
│   │  
│   └───Amazon Cloud Watch
│   │  
│   └───CSV (docs states that is not accessible in real-time)
│   │  
│   └───JSON (docs states that is not accessible in real-time)
│   │  
│   └───....
```

and continue with:

```
Results output

│
└───Summary
│   │  
│   └───Customize with `handleSummary`
│   
└───Metrics
│   │  
│   └───Amazon Cloud Watch
│   │  
│   └───CSV (docs states that is not accessible in real-time)
│   │  
│   └───JSON (docs states that is not accessible in real-time)
│   │  
│   └───....
```

if we want to include other type of results, 


```
Results output

│
└───Summary
│   │  
│   └───Customize with `handleSummary`
│   
└───Metrics
│   │  
│   └───Amazon Cloud Watch
│   │  
│   └───CSV (docs states that is not accessible in real-time)
│   │  
│   └───JSON (docs states that is not accessible in real-time)
│   │  
│   └───....
│   
└───Logs
```

I followed the Slack thread, and I expect other folks disagree with this proposal. Still, I think this proposal brings a good balance between simplicity and correctness.

MattDodsonEnglish:
@ppcano , really I agree with your logic completely. How about:

```
Result output
- End of test >
  - Summary
  - Custom summary
  - Json
  - CSV
- Real-time >
  ...
- Logs
```

Or, to go flat:

```
Result output
- Summary >
    - Customize with handlesummary
- JSON
- CSV
- Real time >
  ---
- Logs
```

MattDodsonEnglish:
I also like this
```
Results output
└───Summary
│   │  
│   └───Customize with `handleSummary`
│   
└───Metrics  
│   └───Amazon Cloud Watch 
│   └───CSV (docs states that is not accessible in real-time)
│   └───JSON (docs states that is not accessible in real-time)
│   └───.... 
└───Logs
```

The only problem is that we have another page already called "Metrics" under "using".

ppcano:
I don't have the data but I think exporting metrics in `JSON` and `CSV` has minimal adoption, which might be more of an advanced case.  Because of this, I think the JSON and CSV options should not be very prominent. 

I am for having the more common cases at the top level:

- Summary
- Metrics
- * Logs

> The only problem is that we have another page already called "Metrics" under "using".

Agree. I also find the distinction between Summary and Metrics could be inaccurate.  Still, I find it the more intuitive option -  but, it might require some tweaks on the articles.

MattDodsonEnglish:
Ok, great. And really the summary is metrics―aggregated metrics―so we could just add this:

```
Result output
End-of-test metrics >
Real-time metrics >
Logs >
```

I think of logs as a distinct entity, but we don't have any docs to put there yet, so it's not a real problem :grimacing: .

Maybe we can break down the `using-k6/metrics` page. Something like for options:

```
Metrics >
  - Custom metrics
  - Make Metric types
  - Metric reference
  - Overview page (links to Results output)
```

MattDodsonEnglish:
Hey @ppcano , I think it should be ready now. I've reduced the entry points for each section, and made the titles more sensible.

src/data/markdown/translated-guides/en/01 Getting started/04 Results Output.md
```diff
@@ -3,35 +3,30 @@ title: 'Results output'
...
-
-k6 generates many metrics about the load that your test generates and how the system under test (SUT) responds.
+As k6 generates load for your test, it also makes _metrics_ that measure the performance of the system.
 Broadly, you can analyze metrics in two ways:
-- As summary statistics, in an _end-of-test-summary_ report.
-- As _time-series data_, which you can write to a file, or stream to external services such as Prometheus or InfluxDB.
-
-![A diagram of the two broad ways to handle results: aggregated and granular](./images/k6-results-diagram.png)
+- As summary statistics, in an _end-of-test_ summary report.
+- In granular detail, with measurements for every point of the test (and timestamps)
```
MattDodsonEnglish, line 9:
```suggestion
- In granular detail, with measurements for every data point across test (and timestamps)
```

src/data/markdown/translated-guides/en/03 Results output/200 Real-time.md
```diff
@@ -0,0 +1,32 @@
+---
+title: Real time
+excerpt: Send your time-series k6 metrics to multiple file formats and services
+---
+
+Besides writing metric data points to [JSON](../end-of-test/json) and [CSV](../end-of-test/json), you can also stream them in real-time to:
```
MattDodsonEnglish, line 6:
```suggestion
Besides writing metric data points to [JSON](../end-of-test/json) and [CSV](../end-of-test/json), you can also stream them in real-time to the following services:
```

PR #825, opened 2022-10-13
Resources; Add link to sample test scripts

MattDodsonEnglish:


MattDodsonEnglish:
@ppcano I requested a review just in case you have a better idea about which section to put this link. "Learn" seems appropriate to me. I'm confident that directory could be useful to a lot of people, and it's definitely good to boost the discoverability.

ppcano:
Good catch! Merging it. 

I think we never linked because we wanted to migrate these examples somewhere else. Anyway, we can do it later and update this link.

PR #827, opened 2022-10-13
Move CSV and JSON to real-time metrics

MattDodsonEnglish:
@Codebien made a point that CSV and JSON _can_ be real-time metrics, since the user can follow the file with `tail -f``.

Because of this, I've moved those files to the Real time section.

However, since these files can also looked at the end of test, I've left the links to their docs in the /end-of-test table. This solution is the closest I can think of towards creating a [polyhierarchy](https://www.nngroup.com/articles/polyhierarchy/).

Along with moving the files and fixing paths, I cleaned up the CSV and JSON docs just a little bit. I also added the `tail -f` command, so users know why these outputs can go in the real-time section.

Don't worry: `tail -f` is POSIX complaint[^1]
[^1]: https://www.unix.com/man-page/POSIX/1posix/tail/


ppcano:
> However, since these files can also looked at the end of test, I've left the links to their docs in the /end-of-test table. This solution is the closest I can think of towards creating a [polyhierarchy](https://www.nngroup.com/articles/polyhierarchy/).

Got the point. But why are not the other real-time options included in this table? I guess we can argue that they behave in the same way - stream outputs can also be looked at the end of test.

2 - The redirects are missing. We could: 
- update old ones to point to the new location (old-path/csv -> real-time/csv)
- create new ones for the new relocation (end-of-test/csv -> real-time/csv)

MattDodsonEnglish:
Can't believe I forgot the redirects, thanks! I just put in new ones, on the off chance the paths have already been linked.


> Got the point. But why are not the other real-time options included in this table? I guess we can argue that they behave in the same way - stream outputs can also be looked at the end of test.

This is true, but doesn't a seem file to likely be at the end of a test? It seems so likely that we let it slip through the first time :grimacing: .

On the other hand a lot of the other services are part of monitoring and observability stacks. Monitoring basically implies "real-time," doesn't it?

But most importantly, adding all these services to "end-of-test" would just open up the whole wormhole again. I'd prefer to remove JSON and CSV from the table instead. But I think that could go against some reader expectations. Now that I've typed this, I don't have a strong opinion.

ppcano (approved):

codebien:
The table in https://k6.io/docs/results-output/end-of-test sounds not consistent anymore.

Issue #829, opened 2022-10-16
Update REST API docs and embed in site


MattDodsonEnglish:
The current REST API docs are on a separate page. This degrades UX in a few ways:
- it disorients the reader.
- It makes the reader have to navigate what are essentially two different docs sites.

For this issue, the task is to move the content on the current page to inside the Gatsby site. In the process, we can clean the docs up and make sure things are correct.

For another issue, it'd be good to keep an underlying openAPI source. However, to turn this source into coherent styled documentation, we probably have to write a custom parser to turn the YAML into markdown. For now, losing the OpenAPI as a source of truth is not a giant loss because:
- The API is stable and unchanging. No updates have happened for years.
- The API is small.

If either of these facts change, I'd say that it'd be worth figuring out how to maintain an OpenAPI spec. This is not only to make automation easier, but also to take advantage of the great tooling that support the OpenAPI ecosystem.

MattDodsonEnglish:
Closed with https://github.com/grafana/k6-docs/pull/833

PR #830, opened 2022-10-16
DRAFT: Cloud REST, clean intro docs

MattDodsonEnglish:
Fixes: https://github.com/grafana/k6-docs/issues/829

Some outstanding information to confirm before merging. I'm adding these in the order I edit, so they may get answered by the doc itself.

## Authentication

- Why does it say "Bearer" and "custom token"? It looks like the example CURL syntax declares an auth-scheme as `token`.

  ```sh
  curl -H "Authorization: token <YOUR_API_TOKEN_HERE>"  #replace with your token
  ```
  This seems to violate a MUST of the RFC 6750, which describes the OAuth2 bearer-token usage:
   >  All challenges defined by this specification MUST use the auth-scheme
   >  value "Bearer".  This scheme MUST be followed by one or more
   > auth-param values. [^1]
 
   Or is this a different use of "Bearer?"
 
 ## Errors
 
 - These are errors returned by the k6 API, right? The current text is a little vague.
 - If the errors come from the API, it seems weird to say that fields are required, since that implies that the user must do something. Wouln't it make more sense to say "The API always returns the following fields:".
 - What's an example of an erroneous data field? (maybe that will be answered when I read the request docs)
 

 
 
 [^1]: https://datatracker.ietf.org/doc/html/rfc6750#section-3 

## Versioning

- I've removed all paragraphs about "asking for new endpoint docs", since k6 cloud hasn't finalized what it'd like to expose.
- Out of scope of doc, but having different version paths in the same API reference feels like confusing devX. It'd be nice to regularize this.

## Test runs

- This page has two endpoints that are not on the current doc: "Read doc run" and "List test run". Should these be removed, or are the endpoints stable enough to add?

- While "Start test run" goes through the `/tests` path, these endpoints go through the `/run` path. Putting separate paths on the same page seems like confusing DevX.  Edit: moved to the "Test" page

## Metrics

This page has a number of endpoints that are currently undocumented:
- Two calls to `v2/thresholds`
- POST `/loadtests/v2/runs/{test_run_id}/export`
- `/loadtests/v2/run-overviews?test_run_id={test_run_id}`


src/data/markdown/docs/06 Cloud REST API/01 Cloud REST API/01 Introduction.md
```diff
@@ -4,10 +4,8 @@ excerpt: 'Introduction to Cloud REST API'
 draft: 'true'
 ---
 
-Welcome to the k6 Cloud API reference documentation. 
-
-This documentation should be of interest to anyone who wishes to use or build upon the k6 Cloud platform. This API gives you programmatic access to the same set of features used to power the service at [app.k6.io](https://app.k6.io). 
+Welcome to the k6 Cloud REST API reference.
```
MattDodsonEnglish, line 7:
```suggestion
Welcome to reference for the k6 Cloud REST API.
```

src/data/markdown/docs/06 Cloud REST API/01 Cloud REST API/01 Introduction.md
```diff
@@ -4,10 +4,8 @@ excerpt: 'Introduction to Cloud REST API'
...
+The k6 Cloud API gives you programmatic access to the same set of features used to power the service at [app.k6.io](https://app.k6.io).
```
MattDodsonEnglish, line 8:
```suggestion
This API gives you programmatic access to the same set of features used to power the service at [app.k6.io](https://app.k6.io). 
```

src/data/markdown/docs/06 Cloud REST API/01 Cloud REST API/02 Authentication.md
```diff
@@ -6,9 +6,12 @@ draft: 'true'
 
 Authentication is based on a custom token scheme.
 
-Your Auth Token lets you to interact with the k6 Cloud through k6 CLI and REST API. To get the token, please visit [this page](https://app.k6.io/account/api-token).
+The Auth Token lets you to interact with the k6 Cloud through the k6 CLI and REST API.
+If you have adequate permissions, you can get the token in your account settings in the k6 Cloud app.
+The k6 Cloud [Token docs](https://k6.io/docs/cloud/integrations/token) describe the different access levels a token can have and the procedure to generate a token.
```
MattDodsonEnglish, line 11:
```suggestion
The k6 Cloud [Token docs](https://k6.io/docs/cloud/integrations/token) describe the procedure to generate a token and the different access levels that a token can have.
```

src/data/markdown/docs/06 Cloud REST API/01 Cloud REST API/04 Tests.md
```diff
@@ -6,18 +6,22 @@ draft: 'true'
 
 ## List tests
 
-Returns all tests within a specified project (if project is not specified then default project will be used). You may sort or query the results with the specified options.
+Returns all tests within a specified project.
+If no project is specified, k6 returns the default project.
```
MattDodsonEnglish, line 10:
```suggestion
If you don't specify a project, k6 returns the default project.
```

MattDodsonEnglish:
Hey @Griatch do you mind giving a quick review of the information in this doc? It has almost full information parity with the current docs. The only thing that needs to be added is a table reference of the fields and their data types. With that, the doc would be equal to current and a little more stylish.

In the future, it could be futher improved with some examples and better descriptions. But before any of this work, it'd be nice to confirm that the information is correct and presents what we want to present.

Mainly, can you confirm whether we should delete or keep in the Endpoints in the Test run and Metrics pages (they are not in the original).

For accuracy, it'd also be nice to know whether the error reference looks correct, and to clarify whether the authentication section should indeed mention a bearer token (read the "Authentication" question at the top of this PR).

ppcano:
Closing it in favour of #833

PR #833, opened 2022-10-19
Cloud REST API

ppcano:
Supercedes #830 
Screenshot 2022-10-19 at 06 23 32

See at https://mdr-ci.staging.k6.io/docs/refs/pull/833/merge/cloud/cloud-reference/cloud-rest-api/

The content to edit is available at: 

- https://github.com/grafana/k6-docs/blob/content/rest-api2/src/data/markdown/docs/03%20cloud/07%20Cloud%20reference/800%20Cloud%20REST%20API.md

- https://github.com/grafana/k6-docs/tree/content/rest-api2/src/data/markdown/docs/03%20cloud/07%20Cloud%20reference/800%20Cloud%20REST%20API


MattDodsonEnglish:
(Summarizing from other PR)

This doc has almost all the information of the current REST doc. The only thing missing is the tables of request fields and their data types. Before I add that, I'd like to confirm that the information is what should be presented.

- This doc (which I reused from a draft from 2020) has some endpoints that aren't in the current doc:
  - All the `/runs` endpoints. Should we just delete the entire "Test Runs" page?
  - The thresholds and half the other endpoints in this metrics Page. Should we delete them.
  
 - Authentication scheme seems confusing to me: Why does it say "Bearer" and "custom token"? It looks like the example CURL syntax declares an auth-scheme as token.

    curl -H "Authorization: token <YOUR_API_TOKEN_HERE>"  #replace with your token

    This seems to violate a MUST of the RFC 6750, which describes the OAuth2 bearer-token usage:

      > All challenges defined by this specification MUST use the auth-scheme
      > value "Bearer". This scheme MUST be followed by one or more
      > auth-param values. [1](https://github.com/grafana/k6-docs/pull/830#user-content-fn-1-87e85f2cb314297f7fb1e29444a76d5f)

     Or is this a different use of "Bearer?"
     
- I know it's out of the scope of a documentation review, but I would like to raise the inconvenience of having a single API doc that goes to multiple version paths. It'd be nice to have just one path (putting in some server-side redirects to keep it backwards compatible).


@mpandurovic @Griatch

MattDodsonEnglish:
- [ ] After this merges, we need to redirect developers.k6.io to these docs. @ppcano

MattDodsonEnglish:
One more thing: is it possible to add a "Stop test run" endpoint? It seems kind of weird that we can allow a user to programmatically start a test, but not stop one.

Griatch (changes requested):
@MattDodsonEnglish 
- The `/v2/runs` endpoints are just the old aliases for `/loadtests/v2/runs` endpoints - so they should be renamed, not be removed. 
- The first part of `Token <token>` could technically be named whatever, including "Bearer". But we have historically always used "Token" as an example here and should probably keep doing so. So this is a sort-of "bearer token" scheme and we used that because it was a term people likely recognized more. 
- Good point about the stop-test endpoint. One already exists, but it sits under `/loadtests/v2/runs` (rather than `under /loadtests/v2/tests`) and was apparently not added to the docs: 
```
POST /loadtests/v2/runs/<test_run_id>/stop
Manually stop a test-run.
```

src/data/markdown/docs/03 cloud/07 Cloud reference/800 Cloud REST API.md
```diff
@@ -1,4 +1,45 @@
 ---
 title: 'Cloud REST API'
-redirect: 'https://developers.k6.io/'
+excerpt: 'Introduction to Cloud REST API'
 ---
+
+Welcome to the k6 Cloud REST API docs!
+
+The k6 Cloud API gives you programmatic access to the same set of features used to power the service at [app.k6.io](https://app.k6.io).
```
MattDodsonEnglish, line 8:
```suggestion

```

src/data/markdown/docs/03 cloud/07 Cloud reference/800 Cloud REST API/04 Tests.md
```diff
@@ -1,23 +1,27 @@
 ---
 title: 'Tests'
 excerpt: ''
-draft: 'true'
 ---
 
+To read, update, and start tests, use the `/tests` path.
```
MattDodsonEnglish, line 6:
```suggestion
To read, update, and start tests, use the `/loadtests/v2/tests/` path.
```

src/data/markdown/docs/03 cloud/07 Cloud reference/800 Cloud REST API.md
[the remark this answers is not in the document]
MattDodsonEnglish: Removed.

src/data/markdown/docs/03 cloud/07 Cloud reference/800 Cloud REST API/04 Tests.md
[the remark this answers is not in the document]
MattDodsonEnglish: Sure. changed

src/data/markdown/docs/03 cloud/07 Cloud reference/800 Cloud REST API/04 Tests.md
```diff
@@ -1,23 +1,27 @@
 ---
 title: 'Tests'
 excerpt: ''
-draft: 'true'
 ---
 
+To read, update, and start tests, use the `/loadtests/v2/tests/` path.
+
 ## List tests
```
MattDodsonEnglish, line 8:
```suggestion
## List load tests
```

src/data/markdown/docs/03 cloud/07 Cloud reference/800 Cloud REST API/05 Test Runs.md
```diff
@@ -190,3 +198,10 @@ Returns test runs for a particular test.
 ```
 
 </CodeGroup>
+
+
+## Stop test runs
+
+Manually stops a test run.
+
+**POST** `/loadtests/v2/runs/<test_run_id>/stop``
```
MattDodsonEnglish, line 207:
```suggestion
**POST** `/loadtests/v2/runs/<test_run_id>/stop`
```

MattDodsonEnglish:
> Good point about the stop-test endpoint. One already exists, but it sits under /loadtests/v2/runs (rather than under /loadtests/v2/tests) and was apparently not added to the docs:

With https://github.com/grafana/k6-docs/pull/833/commits/4669b57f60a77684c917aa5ba451d20bb1faf2da , I added the stop test endpoint. I used what I think is the most minimal text possible. From trying to reverse engineer the request in the console, it looks like there's no body in a stop-test request.

With this change: I also moved "start test" to runs, even thought it goes to /tests. No matter what, something will be a little incoherent. Either the start request will be on "Runs" even though it goes through /tests, or Start test and Stop test will be split across pages.

Keeping "Start" and "stop" on the same page seems most  logical here.

I think that's all we can do! Adding more endpoints would require some digging in the code or network console, and the decision is more for product. Feel free to commit on top, though. If it ooks good, feel free to merge. Thanks for the review :+1: .

Griatch (approved):
Thanks for the updates! I made a few more updates to this, to fix some links and to clarify some texts. 
I didn't double-check endpoints (sounds like you already did). 
Since this also involves re-pointing the old doc domain to this, I'll defer to @ppcano on how to merge this.

MattDodsonEnglish:
I've tested the endpoints. (Some I needed to confirm by replicating in app, then using the `copy-as-curl` from the developer tools.) Putting in a redirect is in progress. But we need to deploy first (to have something to redirect to :slightly_smiling_face: ).

PR #838, opened 2022-10-20
[fix] Remove duplicate Manage teams header

EdvinasDaugirdas:
Teams doc page has duplicate **Managing teams** header. This PR removes that.

Screenshot 2022-10-20 at 09 30 58


MattDodsonEnglish:
Thanks! Will deploy right away.

Issue #679, opened 2022-05-20
Doc needed: How to customize the end-of-summary report


MattDodsonEnglish:
Or, how to use the `handlesummary` function. The docs mention `handlesummary` often, and discourage using `summary-export` often, but there is no real example that shows how to use `handlesummary` to customize a report. There are a few user-contributed external articles, much appreciated, but they exist in their own context and solve their particular problems.

The docs would benefit from a how-to guide that:
1. Comes up with an example custom-report specification
2. Uses handlesummary to implement the spec , describing each step
3. Shows how the reader can verify the report says what we want it to say

MattDodsonEnglish:
The documentation is also missing a reference for handle summary data. If someone wants to create a custom reports, they might prefer to look at the general structure of the data in a single web page.

To be specific, I mean what the user would find in `summary-data.json` in this snippet:

```js
export function handleSummary(data) {
  return {
    "summary-data.json": JSON.stringify(data), // JSON with all the details...
  };
}
```
While the data is going to differ between users depending on their metrics and configured trends, the basic structure is the same.

MattDodsonEnglish:
Community request: "[Can I use handleSummary and print to console at the same time?](https://community.k6.io/t/how-can-i-use-handlesummary-and-print-the-summary-on-console-at-the-same-time/)"

The link contains the solution.

PR #842, opened 2022-10-21
Rewrites HandleSummary() doc

MattDodsonEnglish:
Besides sentence editing and content restructuring, this:
- Adds a small explanatory section
- Makes a progressive series of examples to show how it works
- Splits the reference into a separate section and describes the data model a bit more

Fixes: https://github.com/grafana/k6-docs/issues/679

### DIFF
--- a/src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
+++ b/src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
@@ -19,7 +19,7 @@ However, we plan to support the feature for k6 Cloud tests, too.
 
 ## About the `handleSummary()` callback
 
-After your VU code runs, k6 aggregates your metrics into a JavaScript object.
+After your test runs, k6 aggregates your metrics into a JavaScript object.
 The `handleSummary()` function takes this object as an argument (called `data` in all examples here).
 
 You can use `handleSummary()` to create a custom summary or return the default summary object.
--- a/src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
+++ b/src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
@@ -48,7 +48,7 @@ export function handleSummary(data) {
 
 Fundamentally, `handleSummary()` is just a function that can access a data object.
 As such, you can transform the summary data into any text format: JSON, HTML, console, XML, and so on.
-You can pipe your custom summary to standard out or standard error, write it to a file, or send it to a remote server.
+You can pipe your custom summary to [standard output or standard error](https://en.wikipedia.org/wiki/Standard_streams), write it to a file, or send it to a remote server.
 
 ## Use handleSummary()
 
@@ -62,14 +62,14 @@ k6 expects `handleSummary()` to return a `{key1: value1, key2: value2, ...}` map
 
 The keys must be strings.
 They determine where k6 displays or saves the content:
-- `stdout` for [standard output](https://en.wikipedia.org/wiki/Standard_streams)
+- `stdout` for standard output
 - `stderr` for standard error,
 - any relative or absolute path to a file on the system (this operation overwrites existing files)
 
 The value of a key can have a type of either `string` or [`ArrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer).
 
 You can return mutiple summary outputs in a script.
-As an example, this `return` statement sends a report to standard out and writes the `data` object to a JSON file.
+As an example, this `return` statement sends a report to standard output and writes the `data` object to a JSON file.
 
 <CodeGroup labels={["example keys for handleSummary output"]}>
 
@@ -84,7 +84,7 @@ As an example, this `return` statement sends a report to standard out and writes
 
 ### Extract data properties
 
-This minimal `handleSummary()` extracts the `median` value for the `iteration_duration` metric and prints it on standard out:
+This minimal `handleSummary()` extracts the `median` value for the `iteration_duration` metric and prints it to standard output:
 
 <CodeGroup labels={["Print metric value"]} lineNumbers={[]} showCopyButton={[true]}>
 
--- a/src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
+++ b/src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
@@ -1,9 +1,9 @@
 ---
 title: Custom summary
-excerpt: With the handlesummary() function, you can customize every part of your report. Change the content, redirect output, and more.
+excerpt: With handlesummary(), you can customize every part of your report. Change the content, redirect output, and more.
 ---
 
-With the `handleSummary()` callback, you can completely customize your end-of-test summary.
+With `handleSummary()`, you can completely customize your end-of-test summary.
 In this document, you can read about:
 - How `handleSummary()` works
 - How to customize the content and output location of your summary
@@ -19,11 +19,11 @@ However, we plan to support the feature for k6 Cloud tests, too.
 
 ## About the `handleSummary()` callback
 
-After your VU code runs, k6 aggregates your metrics into an object.
+After your VU code runs, k6 aggregates your metrics into a JavaScript object.
 The `handleSummary()` function takes this object as an argument (called `data` in all examples here).
 k6 calls `handleSummary()` at the end of the test run, even after [`teardown()`](/using-k6/test-life-cycle).
 
-You can use `handleSummary()` to manipulate this object as you want.
+You can use `handleSummary()` to create a custom summary or return the default summary object.
 To get an idea of what the data looks like,
 run this script and open the output file, `summary.json`.
 
@@ -82,7 +82,7 @@ As an example, this `return` statement sends a report to standard output and wri
 
 </CodeGroup>
 
-### Extract data properties
+### Example: extract data properties
 
 This minimal `handleSummary()` extracts the `median` value for the `iteration_duration` metric and prints it to standard output:
 
@@ -108,7 +108,7 @@ export function handleSummary(data) {
 </CodeGroup>
 
 
-### Modify default output
+### Example: modify default output
 
 If `handleSummary()` is exported, k6 _does not_ print the default summary.
 However, if you want to keep the default output, you could import `textSummary` from the [K6 JS utilities library](https://jslib.k6.io/).
@@ -211,7 +211,7 @@ For compactness, these outputs were limited with the `summaryTrendStats` option.
 </Collapsible>
 
 
-### Custom reports to a remote server
+### Example: Send to remote servers and make custom reports
 
 You can also send the generated reports to a remote server (over any protocol that k6 supports).
 This script:
--- a/src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
+++ b/src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
@@ -4,7 +4,7 @@ excerpt: With handlesummary(), you can customize every part of your report. Chan
 ---
 
 With `handleSummary()`, you can completely customize your end-of-test summary.
-In this document, you can read about:
+In this document, read about:
 - How `handleSummary()` works
 - How to customize the content and output location of your summary
 - The data structure of the summary object
@@ -21,7 +21,6 @@ However, we plan to support the feature for k6 Cloud tests, too.
 
 After your VU code runs, k6 aggregates your metrics into a JavaScript object.
 The `handleSummary()` function takes this object as an argument (called `data` in all examples here).
-k6 calls `handleSummary()` at the end of the test run, even after [`teardown()`](/using-k6/test-life-cycle).
 
 You can use `handleSummary()` to create a custom summary or return the default summary object.
 To get an idea of what the data looks like,
@@ -49,6 +48,9 @@ export function handleSummary(data) {
 Fundamentally, `handleSummary()` is just a function that can access a data object.
 As such, you can transform the summary data into any text format: JSON, HTML, console, XML, and so on.
 You can pipe your custom summary to [standard output or standard error](https://en.wikipedia.org/wiki/Standard_streams), write it to a file, or send it to a remote server.
+
+
+k6 calls `handleSummary()` at the end of the [test life cycle](/using-k6/test-life-cycle).
 
 ## Use handleSummary()
 
@@ -162,9 +164,10 @@ export function handleSummary(data) {
 In the collapsible, you can use the tabs to compare default and modified reports.
 
 
-<Collapsible title="Compare default and modified reports" isOpen="" tag="">
+<Collapsible title="Compare the default and modified reports" isOpen="" tag="">
 
-Select `Modified` to see the output of the preceding script.
+To see the output of the preceding script,
+select **Modified**.
 For compactness, these outputs were limited with the `summaryTrendStats` option.
 
 <CodeGroup labels={["Default report", "Modified"]} lineNumbers={[]} showCopyButton={[true]}>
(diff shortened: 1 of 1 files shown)


MattDodsonEnglish:
By the way, I intentionally left some of the summarytrend options in the `textSummary` function undocumented. These options would get overwritten by script options, (I think), and it seems like using them would just cause unnecessary confusion.

src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
```diff
@@ -1,35 +1,76 @@
...
+  };
+}
+```
+
+</CodeGroup>
+
+
+Fundamentally, `handleSummary()` is just a function that can access a data object.
+As such, you can transform the summary data into any text format: JSON, HTML, console, XML, and so on.
+You can pipe your custom summary to standard out or standard error, write it to a file, or send it to a remote server.
```
MattDodsonEnglish, line 51:
```suggestion
You can pipe your custom summary to [standard output or standard error](https://en.wikipedia.org/wiki/Standard_streams), write it to a file, or send it to a remote server.
```

src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
```diff
@@ -41,7 +82,227 @@ The keys determine where k6 displays or saves the content:
 
 </CodeGroup>
 
-To get an idea how `data` would look in your specific test run, just add `return { 'raw-data.json': JSON.stringify(data)};` in your `handleSummary()` function and inspect the resulting `raw-data.json` file. Here's a very abridged example of how it might look:
+### Extract data properties
+
+This minimal `handleSummary()` extracts the `median` value for the `iteration_duration` metric and prints it on standard out:
```
MattDodsonEnglish, line 87:
```suggestion
This minimal `handleSummary()` extracts the `median` value for the `iteration_duration` metric and prints it to standard output:
```

src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
```diff
@@ -1,35 +1,76 @@
...
-k6 expects `handleSummary()` to return a `{key1: value1, key2: value2, ...}` map that represents the summary-report content.
-While the values can have a type of either `string` or `ArrayBuffer`, the keys must be strings.
+k6 expects `handleSummary()` to return a `{key1: value1, key2: value2, ...}` map that represents the summary metrics.
 
-The keys determine where k6 displays or saves the content:
+The keys must be strings.
+They determine where k6 displays or saves the content:
 - `stdout` for [standard output](https://en.wikipedia.org/wiki/Standard_streams)
 - `stderr` for standard error,
 - any relative or absolute path to a file on the system (this operation overwrites existing files)
```
MattDodsonEnglish, line 67:
```suggestion
- `stdout` for standard output
- `stderr` for standard error,
- any relative or absolute path to a file on the system (this operation overwrites existing files)
```

src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
```diff
@@ -1,35 +1,76 @@
...
 
+The value of a key can have a type of either `string` or [`ArrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer).
+
+You can return mutiple summary outputs in a script.
+As an example, this `return` statement sends a report to standard out and writes the `data` object to a JSON file.
```
MattDodsonEnglish, line 72:
```suggestion
As an example, this `return` statement sends a report to standard output and writes the `data` object to a JSON file.
```

src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
```diff
@@ -1,35 +1,76 @@
...
 [Track progress in this issue](https://github.com/grafana/k6-cloud-feature-requests/issues/24).
 
 </Blockquote>
 
-## Data format returned by `handleSummary()`
+## About the `handleSummary()` callback
+
+After your VU code runs, k6 aggregates your metrics into an object.
+The `handleSummary()` function takes this object as an argument (called `data` in all examples here).
+k6 calls `handleSummary()` at the end of the test run, even after [`teardown()`](/using-k6/test-life-cycle).
```
ppcano, line 24: > k6 calls `handleSummary()` at the end of the test run, even after [`teardown()`](/using-k6/test-life-cycle).

I think this could be explained later.

src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
```diff
@@ -1,35 +1,76 @@
...
-However we plan to support the feature for k6 Cloud tests, too.
+However, we plan to support the feature for k6 Cloud tests, too.
 [Track progress in this issue](https://github.com/grafana/k6-cloud-feature-requests/issues/24).
 
 </Blockquote>
 
-## Data format returned by `handleSummary()`
+## About the `handleSummary()` callback
+
+After your VU code runs, k6 aggregates your metrics into an object.
```
ppcano, line 22: We mostly talk about test results and end-of-test summary. The word `object` does not sound familiar at this moment. 

One attempt:

> After k6 executes the load test, it aggregates the test results into a summary. The handleSummary() function receives this summary as an argument (called data in all examples here). 

> You can use handleSummary() to manipulate the test result summary as you want. To get an idea of what the data looks like, run this script and open the output file, summary.json.

src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
```diff
@@ -41,7 +82,227 @@ The keys determine where k6 displays or saves the content:
 
 </CodeGroup>
 
-To get an idea how `data` would look in your specific test run, just add `return { 'raw-data.json': JSON.stringify(data)};` in your `handleSummary()` function and inspect the resulting `raw-data.json` file. Here's a very abridged example of how it might look:
+### Extract data properties
```
ppcano, line 85: Suggest making it explicit that this is a (dummy) example.

### Example: extract data properties

??

src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
```diff
@@ -41,7 +82,227 @@ The keys determine where k6 displays or saves the content:
...
+  return {
+    stdout: latency_message,
+  };
+}
+```
+
+</CodeGroup>
+
+
+### Modify default output
```
ppcano, line 111: Same here:

Example: modify default output

src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
```diff
@@ -41,7 +82,227 @@ The keys determine where k6 displays or saves the content:
...
+
+</CodeGroup>
+
+
+</Collapsible>
+
+
+### Custom reports to a remote server
+
+You can also send the generated reports to a remote server (over any protocol that k6 supports).
```
ppcano, line 216: 1 - Suggest the same:

- ### Example: custom reports to a remote server

2 - The example says `custom reports to a remote server` but it export to 3 outputs with different formats.

I suggest changing this example to showcase only the output to the remote server. 

If not, the title should be different.

ppcano (approved):
This PR significantly improves the article.  💯

I have suggested a few changes.

src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
```diff
@@ -1,35 +1,76 @@
...
 ---
 
-Use the `handleSummary()` function to completely customize the end-of-test summary report.
-
-k6 calls`handleSummary()` at the end of the test run, even after [`teardown()`](/using-k6/test-life-cycle).
-If `handleSummary()` is exported, k6 does _not_ print the default summary.
-
-Besides customizing the CLI summary, you can also transform the summary data into machine- or human-readable formats.
-This lets you create JS-helper functions that generate JSON, CSV, XML (JUnit/xUnit/etc.), HTML, etc. files from the summary data.
+With the `handleSummary()` callback, you can completely customize your end-of-test summary.
```
MattDodsonEnglish, line 6:
```suggestion
With the `handleSummary()` callback, you can completely customize your end-of-test summary.
```

Is there any reason to specify that it's a "callback" function? Or is this exposing too much internals? What function does it call back to?

If it's necessary to mention that it is a callback, maybe we could add an explanatory sentence at the end of the "About handleSummary()) section.

ppcano: I am for Less is More. So, in this context, it could be unnecessary. 

> With `handleSummary()`, you can 


[Read more about callbacks](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Introducing#callbacks)

mstoykov: I would argue it shouldn't be called callback function.

From the [MDN article](https://developer.mozilla.org/en-US/docs/Glossary/Callback_function) the definition is 
> a function passed into another function as an argument

We are missing the whole "passed into another function" so IMO we can drop it.

If we don't I would argue we should call `setup`, `teardown` and `default` callback functions as well ;)

src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
mstoykov: I am also against the wording of

> VU code runs

IMO this has always been confusing, but IMO in this case can just use "the load test has been executed" or some other variant of the "test was executed" instead of specyfing some code that has run and name it something,

src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
```diff
@@ -1,35 +1,76 @@
...
-## Data format returned by `handleSummary()`
+## About the `handleSummary()` callback
+
+After your VU code runs, k6 aggregates your metrics into an object.
+The `handleSummary()` function takes this object as an argument (called `data` in all examples here).
+k6 calls `handleSummary()` at the end of the test run, even after [`teardown()`](/using-k6/test-life-cycle).
+
+You can use `handleSummary()` to manipulate this object as you want.
+To get an idea of what the data looks like,
+run this script and open the output file, `summary.json`.
```
mstoykov, line 28: I do remember there were some differences betweeen this and the file you get from the [summary export](https://k6.io/docs/using-k6/k6-options/reference/#summary-export). 

But I guess we should say that somewhere and also specify you can skip this whole thing by using it. :thinking:

src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
ppcano: I'd separate this convo in another thread. Here my thoughts:


1 - I think `default` would be different --> the default scenario function?

2 - But I agree `setup`, `teardown`, and `handleSummary` follow the same pattern.



3.1 - I have seen using `callback` for this case. But it is confusing.
3.2 - Another common term in JS land is [hook](https://mochajs.org/#hooks)
3.3 - To keep it simple - I am leaning toward using the term `global functions`?

mstoykov (approved):
LGTM :clap:

src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
MattDodsonEnglish: Maybe I'm wrong, but to me this sentence is a key detail:

> Fundamentally, `handleSummary()` is just a function that can access a data object.

So, I'm mildly against removing the reference to `object` in other places because:
- It causes the doc to lose coherence (unless we remove all mentions of the data object).
- It's just less specific, calling it a "summary" doesn't describe the data as well as calling it a "summary object". And working with objects is a fundamental aspect of JavaScript programming, so I don't think it adds new vocabulary or unnecessary complication. Am i making an incorrect assumption here?


On the other hand, I agree with removing the reference to VU code and to callbacks, because these internal details don't really inform how a developer would code a `handleSummary()`.

mstoykov: > So, I'm mildly against removing the reference to object in other places because:

I am also against not calling it `object` or `data` :thinking:. Object is bit vague in general ... so maybe `JavaScript object` just to make certain people understand that in this case we do mean a simpla javascript object.

> You can use handleSummary() to manipulate the test result summary as you want. 

IMO "manimulate" is the wrong word. You return the summary you want, you don't manipulate anything. In practice you can return w/e you want even something completely unrelated to the summary object provided.

src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
MattDodsonEnglish: Yes, this is beyond the scope of this PR, but I have two questions:

These functions are call backs to what other function? So it seems like these functions  run within a "super-function" that k6 automatically sets up when the script runs?

Is this detailed information about runtime useful to any users? If it is, we could probably add it to the Test life cycle document.

mstoykov: > these functions run within a "super-function" that k6 automatically sets up when the script runs?

This won't be how I will explain it. K6 just happens to be calling this functions in a particular pattern that can be explained in the Test life cycle.

> I am leaning toward using the term global functions?

I am for "exported function"  as it (and the rest) need to be *exported* and a *function* so it seems like the best choice to me.

MattDodsonEnglish: Sorry, just for my understanding, isn't there some kind of "superfunction" under the hood? I mean, _to what_ are these functions exported?

I know this is way too technical, and I beg forgiveness for my poor vocabulary, but it has confused me for a while. I think it's related to how I got so confused by "where" the data in `handlesummary(data)` came from.

src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
```diff
@@ -1,35 +1,76 @@
...
 </Blockquote>
 
-## Data format returned by `handleSummary()`
+## About the `handleSummary()` callback
+
+After your VU code runs, k6 aggregates your metrics into an object.
+The `handleSummary()` function takes this object as an argument (called `data` in all examples here).
+k6 calls `handleSummary()` at the end of the test run, even after [`teardown()`](/using-k6/test-life-cycle).
+
+You can use `handleSummary()` to manipulate this object as you want.
```
MattDodsonEnglish, line 26:
```suggestion
You can use `handleSummary()` to create a custom summary or return the default summary object.
```

src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
ppcano: I think the term "exported function" lacks some more context as any k6/JS module can export a function. 

Moving the convo to https://github.com/grafana/k6-docs/issues/845

MattDodsonEnglish:
@ppcano and @mstoykov 
I thank you both for the great reviews, as always! :bow:  

I've removed all references to callbacks and VU code. I also removed the word manipulate.

I'm keeping object in, but I did qualify it as a _JavaScript object_.

With [db6b20d](https://github.com/grafana/k6-docs/pull/842/commits/db6b20dc70c5a63f1f16518ece7dc166bc9df4d4), I split up the examples and hopefully made them less trivial.

If it looks okay to you, Pepe, I think I'm happy to merge it.

ppcano:
@MattDodsonEnglish, I don't have further feedback. Thanks to you for these great improvements!

src/data/markdown/translated-guides/en/03 Results output/100 End-of-test/150-custom-summary.md
mstoykov: > Sorry, just for my understanding, isn't there some kind of "superfunction" under the hood? I mean, to what are these functions exported?

They are exported as [defined](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export) by the ecmascript specification. Which means that other JS modules can use them. 

k6 is not JS code it just uses the same mechanic to call them. 

I guess you can call it a superfunction (there is no such term, and it is not a [higher-order function](https://en.wikipedia.org/wiki/Higher-order_function)) it is just that k6 calls those functions with given parameters in a given ... pattern. They are [called](https://github.com/grafana/k6/blob/8a74171c8b43f48fb53078e0bcfa7851828124bc/js/runner.go#L321) from things that *are* functions within the go language :shrug:. 

But I am not really certain how this is relevant to anyone :shrug: 


>  got so confused by "where" the data in handlesummary(data) came from.

It comes from k6 providing it. 

In practice *almost* everything in computers comes from another program. When you run your program your OS provides it with a bunch of stuff in order to do it. In the same way we provide the info to the script when we execute it. It is just that OSes have some "standards" on how they will execute a program and we have some "standards" on how we will execute the script. And for the most parts most of the users do not care about *most* of those. 

In the cases of the functions we are talking about, you need to know that the first argument will be the data returned from setup or in the case of handleSummary the "summary data". Both of those are just things that when we designed the system decided will be useful. 

To be honest if we make this today - there is a good chance that some of those will just be things `k6/execution` will be able to return to you instead. As documenting that seems a lot easier and is definitely less confusing. Or at least we would have a discussion about it :shrug: 

Hope this helps you!

PR #854, opened 2022-10-27
Drop some extremely experimental extensions by mstoykov

mstoykov:


MattDodsonEnglish:
If the author himself wants to drop the extensions, we must let him. I looked at the diff, seems very safe and uncontroversial. Please merge at will, @mstoykov .

Issue #261, opened 2021-04-08
Fix UUID example


na--:
This page is old: https://k6.io/docs/examples/generating-uuids/

And it's causing problems, as was reported in https://community.k6.io/t/getting-error-while-generating-uuid-in-k6/1607/

Also, as @tom-miseur showed in the reply, we now have a much easier way for people to generate UUIDs

na--:
This seems a bit outdated: https://k6.io/docs/examples/http-authentication/#aws-signature-v4-authentication

We have an alternative in the works (https://github.com/k6io/jslib.k6.io/pull/11), which I see has recent active work from an external contributor (comments in that PR and also https://github.com/k6io/jslib.k6.io/pull/29), so we might get a better alternative soon, especially if someone with JS knowhow and time to test it on AWS chips in.

MattDodsonEnglish:
For the AWS example, would it work now with the [aws jslib](http://localhost:8000/javascript-api/jslib/aws/) that came in https://github.com/grafana/jslib.k6.io/pull/46? I'm not sure how to test that myself.

But if the jslib works, we should update. After that and with the resolution of PR #869, we can close this issue.

PR #869, opened 2022-11-04
Updates UUID example

MattDodsonEnglish:
Partially solves Issue #261.

I tested the example, but before merging there are some things to consider.

- Is it necessary to document an example of a v1 uuids? If so, we still need to include the browserify example. Not sure how needed v1 UUIDs are, though.
- Is this POST request example needed? Or is the first `console.log` enough? If that's enough and we don't need the v1 example, can just we route this page to the documentation for the function?


MattDodsonEnglish:
After 5 minutes of reflection, I feel the second v4 example is totally unnecessary, but I'm not sure about how to handle the code-snippet duplication in the example and in the function documentation.


Maybe we can just duplicate and not worry about it :shrug:

PR #874, opened 2022-11-08
Fix links

olegbespalov:
# What?

:facepalm: I missed the fact that the blockquote isn't a markdown

# Why?

image


MattDodsonEnglish:
Hmm I missed that too. And I think this is unexpected behavior 🤔 . But I will investigate later so we can get the fix out now.

PR #898, opened 2022-11-16
fix: linkinator CI config

sashasushko:
Until I figure out what the problem is with the config file for the linkinator action - https://github.com/JustinBeckwith/linkinator-action/issues/136 - I passed the config directly to workflow.

### DIFF
--- a/.github/workflows/staging.yml
+++ b/.github/workflows/staging.yml
@@ -120,4 +120,10 @@ jobs:
       - uses: JustinBeckwith/linkinator-action@v1
         with:
           paths: ${{ env.GATSBY_DEFAULT_DOC_URL }}
-          config: "linkinator.config.json"
+          concurrency: 50
+          recurse: true
+          retry: true
+          skip: ".*(\\.js|\\.css)$"
+          linksToSkip: "linkedin.com, googleapis.com"
+          verbosity: "ERROR"
+          markdown: false
--- a/.github/workflows/staging.yml
+++ b/.github/workflows/staging.yml
@@ -16,10 +16,10 @@ env:
   GATSBY_ALGOLIA_APP_ID: ${{ secrets.STAGING_CI_GATSBY_ALGOLIA_APP_ID }}
   GATSBY_ALGOLIA_INDEX_NAME: ${{ secrets.STAGING_CI_GATSBY_ALGOLIA_INDEX_NAME }}
   GATSBY_ALGOLIA_SEARCH_ONLY_KEY: ${{ secrets.STAGING_CI_GATSBY_ALGOLIA_SEARCH_ONLY_KEY }}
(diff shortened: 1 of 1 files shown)


sashasushko:
By the way @MattDodsonEnglish because we use GitHub `${{ serets }}` for storing stage server URL, it will be hidden in output:
Снимок экрана 2022-11-16 в 12 10 16

This asterics:
```text
Error: [404] https://***/docs/results-output/custom-summary
                     ^^^
```

Also, because of `${{ serets }}` I can't set up the next workflow job to pretify output:
Снимок экрана 2022-11-16 в 12 09 27

So, the question is - can we move the stage server URL from `${{ serets }}` to public `${{ env }}`?

MattDodsonEnglish:
Thanks for this, and thanks for making the upstream issue. The biggest project that I know used this is bootstrap. Their config is also directly in the action:

 https://github.com/twbs/bootstrap/blob/1a33235adce370dcf36c96d9509e7ce64beaf660/.github/workflows/docs.yml#L41


Can't imagine a problem with moving the staging variable from secrets; that seems better anyway. But I will double check to be sure.

sashasushko:
> Can't imagine a problem with moving the staging variable from secrets; that seems better anyway. But I will double check to be sure.

Cool 👍 I'll be waiting news from you

ppcano:
>  can we move the stage server URL from ${{ serets }} to public ${{ env }}?

@sashasushko I don't see any problem moving this key as a environment variable. Go ahead, thanks!

MattDodsonEnglish (approved):

sashasushko:
@MattDodsonEnglish @ppcano cool 💪 Thanks. One more thing – can someone of you provide to me value of this variable `STAGING_CI_MAIN_URL`?

sashasushko:
I guess it's `staging.k6.io`, but who knows? :-)

ppcano:
@sashasushko staging env is under https://staging.k6.io/

Not sure the real value atm, but you can guess by how the app is building the final URL. Let me know if you need the confirmation.

MattDodsonEnglish:
Thanks for this fix, Sasha. The UX improvement is much better than I thought.

The logs are much lighter, and it's really nice being able to select the links from the logs and go right to the page where the break is:
https://github.com/grafana/k6-docs/actions/runs/3479124301/jobs/5817526912

sashasushko:
You're welcome 🤝 

And... We can do it a little bit better 😄 For example, if we'll format output as `JSON`:
Снимок экрана 2022-11-16 в 18 35 10
_it's my website links, just for testing_

But in that case, we have to use [GitHub Action Artifacts](https://docs.github.com/en/enterprise-cloud@latest/rest/actions/artifacts). That means the report will be attached for each workflow run:
Снимок экрана 2022-11-16 в 18 38 11

PR #904, opened 2022-11-18
chore: fix workflow config syntax

sashasushko:
You can skip quotes


MattDodsonEnglish:
LMAO I hope this makes all links correctly skip. Thanks for saving me more of my monkey work.

MattDodsonEnglish:
I do not understand why the linkedin URL will not skip:
https://github.com/grafana/k6-docs/actions/runs/3498152277/jobs/5858333599#step:3:39. Maybe we can just add it to the regex.

On the bright side. It also caught the four links that I just broke. :-)

PR #945, opened 2022-12-12
Update 14 Scenarios.md

David-Kristek:
I think there is a mistake in docs with naming scenarios executor.

### DIFF
--- a/src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios.md
+++ b/src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios.md
@@ -90,7 +90,7 @@ export const options = {
       startTime: '0s'
     },
     another_scenario: {
-      executor: 'shared-iterations',
+      executor: 'per-vu-iterations',
       startTime: '5s'
     },
   },
--- a/src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios.md
+++ b/src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios.md
@@ -76,32 +76,39 @@ For the list of the executors, refer to the [Executor guide](/using-k6/scenarios
 
 ## Scenario example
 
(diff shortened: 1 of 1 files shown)


MattDodsonEnglish:
Thanks for this!

@David-Kristek, do you mind explaining why you think it should change?  The script works as is, but if it seems confusing or weird to one person, I'm sure it's weird to many others, too.

My idea was to show, in the minimum amount of code, that it's possible to combine scenarios in one script. But, if it looks like a bug, maybe it's too minimal :-). 

Maybe, instead of your suggestion, we could expand the script and add a little context. Something like this:

> This script combines two scenarios, with sequencing:
>
> - The `shared_iter_scenario` starts immediately. Ten VUs try to use 100 iterations as quickly as possible (some VUs may use more iterations than others).
> - The `per_vu_scenario` starts after 10s. In this case, ten VUs each run ten iterations.
>
> Which scenario takes longer?
> You can run to discover.
> You can also add a `maxDuration` property to one or both scenarios.
>
> 
> ```javascript
> import http from "k6/http";
> 
> export const options = {
>   scenarios: {
>     shared_iter_scenario: {
>       executor: "shared-iterations",
>       vus: 10,
>       iterations: 100,
>       startTime: "0s",
>     },
>     per_vu_scenario: {
>       executor: "per-vu-iterations",
>       vus: 10,
>       iterations: 10,
>       startTime: "10s",
>     },
>   },
> };
> 
> export default function () {
>   http.get("https://test.k6.io/");
> }
> ```
> 
> 

@David-Kristek , what do you think?

David-Kristek:
Yes this looks better to me. My point was that the output of  `scenario-example.js` didn't match with  theoutput two lines below: `scenario example output`. But now as you wrote it looks perfect.

MattDodsonEnglish:
Thanks for your help! The docs are definitely improved :heart: .

PR #949, opened 2022-12-15
Add How it works section to xk6-disruptor

pablochacin:
Signed-off-by: Pablo Chacin <pablochacin@gmail.com>


MattDodsonEnglish (changes requested):
Looks great! I made some minor fixes for language.

Besides that, I have a few comments:

- I don't exactly understand how the part about the extension and the agent relates to the diagrams about the disrupters. These seem like separate things: one is the architecture of the system, the other is the implementations of some specific APIs.
   - Would it make sense to split the page in two? Something like "Architecture" and "Implementations"
   - Maybe these concepts are so connected that they should be on one page. If so, I think we can improve the flow of information a bit.
- The requirements page links to this https://github.com/grafana/xk6-disruptor/blob/main/docs/01-development/02-architecture.md
   - Would it make sense to move that link to these explanations instead?

MattDodsonEnglish:
Also, I think I'd prefer if "Explanations" came before API. It seems more useful to have the conceptual information before using in this case. What do you think?

image

pablochacin:
@MattDodsonEnglish  
> I don't exactly understand how the part about the extension and the agent relates to the diagrams about the disrupters.

Ok, maybe I should explain the extension provide the disruptors. Would that work?

> Would it make sense to split the page in two? Something like "Architecture" and "Implementations"

I prefer not. I think that will over-complicate the description. 

> The requirements page links to this (repository)

Yeah, good catch I will update this.

>  I think I'd prefer if "Explanations" came before API. 

I would prefer to keep the current order because I think the more common order is how to get started, then try the API and then dig into details.

MattDodsonEnglish:
> Ok, maybe I should explain the extension provide the disruptors. Would that work?

It flows better now.

The one thing that still trips me is this sentence:

> With this mode of operation, the xk6-disruptor can work without requiring you to install additional components.

Without this sentence, the intro flows really well into the following sections. But this sentence breaks my coherence. I'm also not sure about what "this mode of operation" means. Does it mean that the extension installs the agent? Can we just delete this sentence?

This might be a bit beyond the scope of this PR, but I'm still generally confused about how the install works.

As I understand:
1. Build the custom binary.
2. Install the docker container agent in the system under test.
3. Run commands?

Or, does the binary somehow install the docker container? This new page suggests that.

MattDodsonEnglish:
> I would prefer to keep the current order because I think the more common order is how to get started, then try the API and then dig into details.

Here is my thinking:
- Will this high-level information help people use the API, or do people need hands-on experience with the API to appreciate this conceptual information?
- Why not put a minimal hands-on example in the get started section?
- How big will this API get? If it's stays this size, readers can still find all pages in one screen. If the API gets as long as the xk6-browser or some of the main API modules (e.g. redis), then the list will absolutely dominate the page and many readers will likely never see the examples.

pablochacin:
> This might be a bit beyond the scope of this PR, but I'm still generally confused about how the install works.
> As I understand:
>
> Build the custom binary.
> Install the docker container agent in the system under test.
> Run commands?

I see the problem. There is not actual explanation of the whole process in the get started guide that explains this process. But I would prefer to do this as a different PR.

By the way, the docker container agent is installed automatically by the extension, this should be also be clearly stated because is a major advantage of the disruptor.

> Why not put a minimal hands-on example in the get started section?
Yes totally, but this is unrelated to the description of how it works. It is related to the point above.

> Will this high-level information help people use the API, or do people need hands-on experience with the API to appreciate this conceptual information?

No, it is not really needed, that is why I wanted to put it as an optional content and not make it seem more relevant than the API, that is what is needed for using the extension. But if you think it SHOULD be before the API, let's move. 

> The one thing that still trips me is this sentence:
>> With this mode of operation, the xk6-disruptor can work without requiring you to install additional components.

I see the problem. I also struggled with this. I just wanted to highlight that the user doesn't need to install any other component besides the binary.

MattDodsonEnglish (approved):
Looks great. I agree the stuff about clarifying the contain is better for another PR.

MattDodsonEnglish:
If you are ready, feel free to merge.

PR #951, opened 2022-12-16
Update Grafana Cloud guide

codebien:
Update the Grafana Cloud guide using the new k6 builtin output.

### DIFF
--- a/src/data/markdown/translated-guides/en/03 Results output/200 Real-time.md
+++ b/src/data/markdown/translated-guides/en/03 Results output/200 Real-time.md
@@ -34,7 +34,7 @@ You can also stream real-time metrics to the following services:
 - [InfluxDB](/results-output/real-time/influxdb-grafana)
 - [Netdata](/results-output/real-time/netdata)
 - [New Relic](/results-output/real-time/new-relic)
-- [Prometheus](/results-output/real-time/prometheus)
+- [Prometheus remote write](/results-output/real-time/prometheus-remote-write)
 - [TimescaleDB](/results-output/real-time/timescaledb)
 - [StatsD](/results-output/real-time/statsd)
 
--- a/src/data/markdown/translated-guides/en/03 Results output/200 Real-time.md
+++ b/src/data/markdown/translated-guides/en/03 Results output/200 Real-time.md
@@ -34,7 +34,7 @@ You can also stream real-time metrics to the following services:
 - [InfluxDB](/results-output/real-time/influxdb-grafana)
 - [Netdata](/results-output/real-time/netdata)
 - [New Relic](/results-output/real-time/new-relic)
-- [Prometheus remote write](/results-output/real-time/prometheus-remote-write)
+- [Prometheus remote write](/results-output/real-time/prometheus)
 - [TimescaleDB](/results-output/real-time/timescaledb)
 - [StatsD](/results-output/real-time/statsd)
 
--- a/src/data/markdown/translated-guides/en/03 Results output/200 Real-time/00 Grafana Cloud.md
+++ b/src/data/markdown/translated-guides/en/03 Results output/200 Real-time/00 Grafana Cloud.md
@@ -5,7 +5,8 @@ excerpt: >
 ---
 
 
-Using Grafana Cloud Prometheus and the [k6 output for Prometheus remote write](/results-output/real-time/prometheus), you can send your k6 metrics into [Grafana Cloud](https://grafana.com/products/cloud) to better visualize your testing results, enabling you to correlate k6 metrics with other metrics of your observability stack.
+With Grafana Cloud Prometheus and the [k6 output for Prometheus remote write](/results-output/real-time/prometheus), you can send your k6 metrics into [Grafana Cloud](https://grafana.com/products/cloud) to better visualize your testing results.
+With k6 metrics in Grafana, you can correlate k6 metrics with other metrics of your observability stack.
 
 While this article focuses on Grafana Cloud, any remote write capable Prometheus installation is compatible with this approach.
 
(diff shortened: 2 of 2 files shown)


src/data/markdown/translated-guides/en/03 Results output/200 Real-time.md
```diff
@@ -34,7 +34,7 @@ You can also stream real-time metrics to the following services:
 - [InfluxDB](/results-output/real-time/influxdb-grafana)
 - [Netdata](/results-output/real-time/netdata)
 - [New Relic](/results-output/real-time/new-relic)
-- [Prometheus](/results-output/real-time/prometheus)
+- [Prometheus remote write](/results-output/real-time/prometheus-remote-write)
```
codebien, line 37: @MattDodsonEnglish this link does not work with the current change. Can I rename the file or could it break something?

MattDodsonEnglish: Rename the file is fine, but I don't think we need to rename the page path.

On the remote write page, I added a `slug` in the frontmatter so the path stays the same. The PR passed without any link breaks:

https://github.com/grafana/k6-docs/actions/runs/3714477284/jobs/6298652654

codebien: If I try it on staging it redirects to the home page. Is it an issue with staging?

MattDodsonEnglish: ohhh. Yes, it's a problem. I think all 404s have started redirecting to the home page, which means the linkchecker is useless :-( .

MattDodsonEnglish: Besides the linkchecker problem, this still goes to the right page for me. Am I missing something?
https://staging.k6.io/docs/results-output/real-time/prometheus/

MattDodsonEnglish, line 37:
```suggestion
- [Prometheus remote write](/results-output/real-time/prometheus)
```

src/data/markdown/translated-guides/en/03 Results output/200 Real-time/00 Grafana Cloud.md
```diff
@@ -1,13 +1,13 @@
 ---
 title: 'Grafana Cloud'
 excerpt: >
-  How to upload the test result metrics to Grafana Cloud using Grafana Cloud Prometheus and the k6 extension for Prometheus remote write'
+  How to upload the test result metrics to Grafana Cloud using Grafana Cloud Prometheus and the k6 output for Prometheus remote write'
 ---
 
 
-Using Grafana Cloud Prometheus and the [k6 extension for Prometheus remote write](https://github.com/grafana/xk6-output-prometheus-remote), you can send your k6 metrics into [Grafana Cloud](https://grafana.com/products/cloud/) to better visualize your testing results, enabling you to correlate k6 metrics with other metrics of your observability stack.
+Using Grafana Cloud Prometheus and the [k6 output for Prometheus remote write](/results-output/real-time/prometheus), you can send your k6 metrics into [Grafana Cloud](https://grafana.com/products/cloud) to better visualize your testing results, enabling you to correlate k6 metrics with other metrics of your observability stack.
```
MattDodsonEnglish, line 8:
```suggestion
With Grafana Cloud Prometheus and the [k6 output for Prometheus remote write](/results-output/real-time/prometheus), you can send your k6 metrics into [Grafana Cloud](https://grafana.com/products/cloud) to better visualize your testing results.
With k6 metrics in Grafana, you can correlate k6 metrics with other metrics of your observability stack.
```

MattDodsonEnglish (approved):
Looks good with some one nit. And the thing about the URI.

src/data/markdown/translated-guides/en/03 Results output/200 Real-time.md
codebien: @MattDodsonEnglish The main issue is that it is only `/prometheus` and we are trying to set everything as Prometheus **remote write**. It seems not consistent to me. WDYT?

MattDodsonEnglish: I see. I've been thinking that Prometheus Remote Write is the only viable export from k6 to Prometheus. Do you think k6 might export to more Prometheus services in the future?

If there will be more, we have two options:
 - Add a new item in the "Real time" list:
   ```
   RESULTS OUTPUT
  Real time >
    AWS
    Grafana
    Prometheus Remote Write
    Prometheus hypothetical
  ```
- Add a nested list
 ``` 
  Real time >
    AWS
    Grafana
    Prometheus >
      - Remote Write
      - Prometheus hypothetical
  ```
  
  Even thought it's not great to have so much nesting, the second option seem most logical. What's more, if we change the URL to `prometheus-rw` now, we'll have to redirect `prometheus` to that URI, creating another set of problems. I'm going to merge, but let this be a record of my decision-making process.

MattDodsonEnglish: Finally, if we do come into a situation with some messy redirects to figure out, we can at last come up with a "clean" set of URIs using `prom*` .

codebien: The Prometheus remote write is a protocol and it has many implementations the Prometheus server itself is one of them. Remote write alone is not correct, I see it creates a bunch of confusion. Naming is a difficult thing in IT world. :smile:

MattDodsonEnglish: Oh, that is annoying! I didn't think of it this way. I'll change the URI to `prometheus-rw` and program in some redirects. 

Why isn't `remote write` correct on its own? Just wondering?

MattDodsonEnglish: I guess it's incorrect because that's just not the name of the protocol.

codebien: >I guess it's incorrect because that's just not the name of the protocol.

Exactly. There isn't any official explanation, but this is the title in the spec, so this is my personal interpretation.

PR #953, opened 2022-12-19
Bump JavaScript modules API version

codebien:


src/utils/versioning.js
```diff
@@ -9,9 +9,10 @@ const SUPPORTED_VERSIONS = [
   'v0.38',
   'v0.39',
   'v0.40',
+  'v0.41',
```
olegbespalov, line 12: Maybe we could also clean up a few old versions, WDYT? :thinking:

olegbespalov (approved):

src/utils/versioning.js
codebien: What's the policy here, I also thought we should remove some of them, but how many? We should also [document it](https://github.com/grafana/k6-docs/wiki/Add-version-for-Javascript-API-documentation) to make it part of the release process. I think for this time I would prefer to open an issue, sync with the team and address it in the next few days.

codebien: https://github.com/grafana/k6-docs/issues/955

mstoykov (approved):

PR #954, opened 2022-12-19
Add "your first test" section to xk6-disruptor

pablochacin:
Existing documentation does not offer a clear introduction to the first steps for starting using xk6-disruptor

Signed-off-by: Pablo Chacin <pablochacin@gmail.com>

### DIFF
--- a/src/data/markdown/docs/40 xk6-disruptor/01 Get started/01 Welcome.md
+++ b/src/data/markdown/docs/40 xk6-disruptor/01 Get started/01 Welcome.md
@@ -36,7 +36,7 @@ Until this extension reaches v1.0 release, your k6 scripts may need to be update
 
 ## Your first test with xk6-disruptor
 
-1. Check the [requirements](/javascript-api/xk6-disruptor/get-started/requirements)
+1. Check the [requirements](/javascript-api/xk6-disruptor/get-started/requirements).
 
 2. [Install](/javascript-api/xk6-disruptor/get-started/installation) xk6-disrutor binary in your system.
 
--- a/src/data/markdown/docs/40 xk6-disruptor/01 Get started/01 Welcome.md
+++ b/src/data/markdown/docs/40 xk6-disruptor/01 Get started/01 Welcome.md
@@ -38,6 +38,6 @@ Until this extension reaches v1.0 release, your k6 scripts may need to be update
 
 1. Check the [requirements](/javascript-api/xk6-disruptor/get-started/requirements).
 
-2. [Install](/javascript-api/xk6-disruptor/get-started/installation) xk6-disrutor binary in your system.
+2. [Install](/javascript-api/xk6-disruptor/get-started/installation) the xk6-disrutor binary in your system.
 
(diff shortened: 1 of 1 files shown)


src/data/markdown/docs/40 xk6-disruptor/01 Get started/01 Welcome.md
```diff
@@ -33,3 +33,11 @@ xk6-disruptor is in the alpha phase, undergoing active development. k6 doesn't g
 Until this extension reaches v1.0 release, your k6 scripts may need to be updated on each release .
 
 </Blockquote>
+
+## Your first test with xk6-disruptor
+
+1. Check the [requirements](/javascript-api/xk6-disruptor/get-started/requirements)
```
MattDodsonEnglish, line 39:
```suggestion
1. Check the [requirements](/javascript-api/xk6-disruptor/get-started/requirements).
```

MattDodsonEnglish (commented):
Looks fine, but will the user not have to run the step in "Expose your service" first?

pablochacin:
> Looks fine, but will the user not have to run the step in "Expose your service" first?

The short answer is "it depends".  And the long answer is "it is complicated". That is explained in the example, so I don't think we should put as a step.

src/data/markdown/docs/40 xk6-disruptor/01 Get started/01 Welcome.md
```diff
@@ -33,3 +33,11 @@ xk6-disruptor is in the alpha phase, undergoing active development. k6 doesn't g
 Until this extension reaches v1.0 release, your k6 scripts may need to be updated on each release .
 
 </Blockquote>
+
+## Your first test with xk6-disruptor
+
+1. Check the [requirements](/javascript-api/xk6-disruptor/get-started/requirements).
+
+2. [Install](/javascript-api/xk6-disruptor/get-started/installation) xk6-disrutor binary in your system.
```
MattDodsonEnglish, line 41:
```suggestion
2. [Install](/javascript-api/xk6-disruptor/get-started/installation) the xk6-disrutor binary in your system.
```

MattDodsonEnglish (approved):

MattDodsonEnglish:
> > Looks fine, but will the user not have to run the step in "Expose your service" first?
> 
> The short answer is "it depends". And the long answer is "it is complicated". That is explained in the example, so I don't think we should put as a step at this level.
> 
> Also, notice I'll be iterating on the example(s) to simplify them and remove the friction points like this, but unfortunately there's some intrinsic complexity around K8s that we cannot magically remove :-(

Yeah I'm noticing that there's some inherent trickiness here. I think that, as long as we provide some conditional instructions and prereqs (if you are running x, be sure to set up y), we are on the right track.

I think it's best to improve the accuracy and flow in small iterations like this, especially since the application itself is still in experimental development. I also think this specific application requires the readers to fill in some information themselves: they must know how to interact with Kubernetes and apply examples to their specific configurations.
Probably the best we can do is explicitly state the technical expectations we assume of the reader (though the goal is always to be as accessible as possible).

Issue #781, opened 2022-08-30
Use SVGs for icons instead of emojis


MattDodsonEnglish:
This issue is a more general version of: https://github.com/grafana/k6-docs/issues/729

As @ankur22  and @imiric  explained to me:

> We use Unicode font glyphs ("emojis") for all icons, whose rendering is system/browser-dependent.
> The issue with using emojis is that we can't guarantee that the correct fonts/character sets will be installed on the end-users system, so we need an icon to replace this emoji (SVG?) -- 

For example, the WIP icons don't show up for me on Chromium/Linux on the [xk6-browser](https://k6.io/docs/javascript-api/xk6-browser/) docs, though they do show up on Firefox.

There are emojis peppered across the site; it's also a problem on the main /docs page:

![image](https://user-images.githubusercontent.com/47385188/187323784-c18b0d7f-5c4e-49ab-a13e-3dc9533dacac.png)

Possible solutions:

> So instead of relying on system fonts, all icons should be converted to SVGs. Or, at the very least, we should load an icon font that is known to contain the glyphs, but going with SVGs would be more reliable.

PR #959, opened 2022-12-20
Fix orientation of 'Running k6' car icon

imiric:
I'm not sure if this was caused by a font change on our side or some Unicode update on my system, since this used to look fine a few months ago, but on Linux and recent stable versions of Firefox and Chromium, the car icon is rendered pointing to the right, while the "wind" icon is pointing to the left. So it looks like the car is racing backwards :smile:

This still uses the same [U+1F3CE](https://www.utf8icons.com/character/127950/racing-car) glyph as before, but for some reason renders properly with the car pointing to the left for me. :man_shrugging: **Please test that this renders correctly on Windows, macOS, iOS and Android.**

This is why it would be a good idea to abandon emojis for this use case, and switch to SVGs (#781).

### Before

- Chromium 107.0.5304.87
  2022-12-20-181430_327x164_scrot

- Firefox 107.0.1
  2022-12-20-181444_291x165_scrot

### After

- Chromium 107.0.5304.87
  2022-12-20-181503_328x164_scrot

- Firefox 107.0.1
  2022-12-20-181518_314x177_scrot

### DIFF
--- a/src/components/pages/doc-welcome/quickstart/quickstart.component.js
+++ b/src/components/pages/doc-welcome/quickstart/quickstart.component.js
@@ -26,7 +26,7 @@ export const Quickstart = () => {
           urlLocale === 'es'
             ? '/es/empezando/ejecucion-de-k6/'
             : '/get-started/running-k6/',
-        title: `🏎💨 ${t('welcome.quickstart.running-k6.title')}`,
+        title: `🏎️💨 ${t('welcome.quickstart.running-k6.title')}`,
         text: t('welcome.quickstart.running-k6.text'),
       },
       {


MattDodsonEnglish:
I appreciate this attention to detail, but unfortunately, it still appears in the wrong orientation for me on Chromium in Linux. And the "wind" glyph doesn't appear it all. If I don't misunderstand how this works, these glyphs do exist on my system though, since they work on Firefox. I too would prefer the SVG approach.

Still, I'd merge this, since it's at least correctly oriented on Firefox―an improvement. Not sure if there's a better solution outside of just using SVGs. Or if the problem is only with my specific device.


image

imiric:
Strange. :confused: What version of Chromium are you using?

There have been some [emoji rendering issues](https://bugs.chromium.org/p/chromium/issues/detail?id=903205), but AFAICS, we don't use Arial in our font stack. Maybe the problem is that the font used for this `h3` (TT Norms Pro) doesn't have this character, and then it falls back to `sans-serif`, which can be anything on your system. So if your fallback font can't render the character, you get the weird default and missing block rendering.

Try changing your Sans-serif font configuration in Chromium (Settings > Appearance > Customize fonts > Sans-serif font). Mine is set to "Custom" for some reason. It's the way it's configured on Void Linux, and I haven't changed it, but I did install a bunch of additional fonts. Maybe also try installing the MS fonts package, which will give you some standard fonts you can fallback on. Or Roboto, if you don't have it.

Another solution would be to make our own build of TT Norms Pro, or any other font for that matter, that does have the emoji characters we use. That way there _shouldn't_ be any fallback, and it _should_ render the same on any system. I think... :sweat_smile::crossed_fingers: But ideally, just using static SVGs would be the way to go.

Anyway, this change is also an improvement on Windows 10:

### Before

- Chromium 100.0.4896.88
2022-12-21-105313_366x217_scrot

- Firefox 102.6.0esr
2022-12-21-105359_362x217_scrot


### After

- Chromium 100.0.4896.88
2022-12-21-105330_367x220_scrot

- Firefox 102.6.0esr
2022-12-21-105424_361x215_scrot


The fact it renders consistently between browsers on Windows hints that this is a system-dependent fallback issue.

MattDodsonEnglish (commented):
Setting my Chromium fonts to the same defaults as firefox's didn't change anything, but I think I'm an edge case. This is a pretty clear improvement on a variety of machines. So I'd prefer merging now and figuring out my weird edge case later.

MattDodsonEnglish (approved):

PR #965, opened 2022-12-22
Add documentation for new wildcard Hosts feature

imiric:
This documents support for using a wildcard with the `hosts` option, which was recently added in v0.42.0. See https://github.com/grafana/k6/pull/2747.

### DIFF
--- a/src/data/markdown/translated-guides/en/02 Using k6/05 k6 Options/02 Reference.md
+++ b/src/data/markdown/translated-guides/en/02 Using k6/05 k6 Options/02 Reference.md
@@ -421,7 +421,7 @@ $ k6 cloud --exit-on-running script.js
 ## Hosts
 
 An object with overrides to DNS resolution, similar to what you can do with `/etc/hosts` on
-Linux/Unix or `C:\\Windows\\System32\\drivers\\etc\\hosts` on Windows. For instance, you could set
+Linux/Unix or `C:\Windows\System32\drivers\etc\hosts` on Windows. For instance, you could set
 up an override which routes all requests for `test.k6.io` to `1.2.3.4`.
 
 You can also redirect only from certain ports or to certain ports.
--- a/src/data/markdown/translated-guides/en/02 Using k6/05 k6 Options/02 Reference.md
+++ b/src/data/markdown/translated-guides/en/02 Using k6/05 k6 Options/02 Reference.md
@@ -424,7 +424,9 @@ An object with overrides to DNS resolution, similar to what you can do with `/et
 Linux/Unix or `C:\Windows\System32\drivers\etc\hosts` on Windows. For instance, you could set
 up an override which routes all requests for `test.k6.io` to `1.2.3.4`.
 
-You can also redirect only from certain ports or to certain ports.
+Starting from v0.28.0, you can also redirect only from certain ports or to certain ports.
+
+Starting from v0.42.0, you can use an asterisk (`*`) at the start of the host name, to avoid repetition. For example, `*.k6.io` would apply the override for all subdomains of `k6.io`.
 
 <Blockquote mod="note" title="">
 
@@ -444,14 +446,14 @@ export const options = {
   hosts: {
     'test.k6.io': '1.2.3.4',
     'test.k6.io:443': '1.2.3.4:8443',
+    '*.k6.io': '1.2.3.4',  // applies for all subdomains of k6.io
   },
 };
 ```
 
 </CodeGroup>
 
-With the above code any request made to `test.k6.io` will be redirected to `1.2.3.4` without changing
-it port unless it's port is `443` which will be redirected to port `8443`.
+With the above code any request made to `test.k6.io` or any subdomains of `k6.io` will be redirected to `1.2.3.4` without changing its port, unless the port is `443` which will be redirected to port `8443`.
 
 ## HTTP debug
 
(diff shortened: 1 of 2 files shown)


src/data/markdown/translated-guides/en/02 Using k6/05 k6 Options/02 Reference.md
```diff
@@ -421,10 +421,12 @@ $ k6 cloud --exit-on-running script.js
...
 
 An object with overrides to DNS resolution, similar to what you can do with `/etc/hosts` on
-Linux/Unix or `C:\\Windows\\System32\\drivers\\etc\\hosts` on Windows. For instance, you could set
+Linux/Unix or `C:\Windows\System32\drivers\etc\hosts` on Windows. For instance, you could set
 up an override which routes all requests for `test.k6.io` to `1.2.3.4`.
 
-You can also redirect only from certain ports or to certain ports.
+Starting from v0.28.0, you can also redirect only from certain ports or to certain ports.
+
+Starting from v0.42.0, you can use an asterisk (`*`) at the start of the host name, to avoid repetition. For example, `*.k6.io` would apply the override for all subdomains of `k6.io`.
```
MattDodsonEnglish, line 429:
```suggestion
k6 also supports ways to narrow or widen the scope of your redirects:
- You can redirect only from or to certain ports.
- To save repetition, you can use an asterisk (`*`) as a wild card at the start of the host name.
For example, `*.k6.io` would apply the override for all subdomains of `k6.io`.
```

MattDodsonEnglish: I don't think we need to mention version numbers as it'll go stale. v0.28 is already ancient. I also tried to organize this information in some sort of coherent group. I added "wildcard" just to help `ctrl+f` or internet searchers.

src/data/markdown/translated-guides/en/02 Using k6/05 k6 Options/02 Reference.md
```diff
@@ -444,14 +446,14 @@ export const options = {
   hosts: {
     'test.k6.io': '1.2.3.4',
```
MattDodsonEnglish, line 447:
```suggestion
```

Should we delete this line? Doesn't the wildcard below already do this redirect?

src/data/markdown/translated-guides/en/02 Using k6/05 k6 Options/02 Reference.md
```diff
@@ -444,14 +446,14 @@ export const options = {
...
+    '*.k6.io': '1.2.3.4',  // applies for all subdomains of k6.io
   },
 };
 ```
 
 </CodeGroup>
 
-With the above code any request made to `test.k6.io` will be redirected to `1.2.3.4` without changing
-it port unless it's port is `443` which will be redirected to port `8443`.
+With the above code any request made to `test.k6.io` or any subdomains of `k6.io` will be redirected to `1.2.3.4` without changing its port, unless the port is `443` which will be redirected to port `8443`.
```
MattDodsonEnglish, line 456:
```suggestion
The preceding code will redirect any request made to any subdomain of `k6.io` to `1.2.3.4`.
The port will remain the same, unless the redirect is from `test.k6.io` on port `443`.
In this case, it redirects to port `8443`.
```

MattDodsonEnglish (changes requested):
I requested some minor views for language, and to smooth out the explanation. My suggestions also make some assumptions about wildcards that I'm not sure about, so please be sure I haven't added anything incorrect!

oleiade (approved):
🎉 🦖

src/data/markdown/translated-guides/en/02 Using k6/05 k6 Options/02 Reference.md
imiric: > I don't think we need to mention version numbers as it'll go stale. v0.28 is already ancient.

That's true, and I wondered about removing it as well, but I think it's important to mention when a new feature is added that it's only available since a specific version. Otherwise people will get confused when they try to use it, and it doesn't work.

I think we need better version management of the docs altogether. Am I wrong, or didn't we have a version dropdown users could select? In this case we could mark this feature as "New" for v0.42.0, and then remove the label in the next version.

Or we could have a strategy to update these "Starting from" labels, and be disciplined about removing them a few versions after they're introduced.

Anyway, I'm OK with your changes. I'll translate them to Spanish to be consistent.

MattDodsonEnglish: I just looked through our internal chat, and I saw that we recommend keeping a version reference for 6 releases / 1 year. I remember that I once removed a lot of old version references, but I forgot about the policy :facepalm: .

 > Otherwise people will get confused when they try to use it, and it doesn't work.
 
 This convinced me that it makes sense to put it in the new reference. I'm so used to seeing old, irrelevant references that I didn't think of how it would look from the other side.

src/data/markdown/translated-guides/en/02 Using k6/05 k6 Options/02 Reference.md
imiric: It does, but I think it's important to document all ways the feature can be used, and we should keep this most basic way of doing the mapping. The example itself doesn't have to represent a real-world usage scenario. We could always use a different domain instead, maybe `*.grafana.com`, but I think this is fine.

MattDodsonEnglish: >We could always use a different domain instead, maybe *.grafana.com, but I think this is fine.
 
I'd rather do this. Or `test-api.k6.io` or whatever. As it is right now, that line in the script doesn't really do anything. So I think it creates ambiguity. Readers will look at it and wonder, _wait, doesn't the wildcard already include that domain? Or do I not understand how this works?_ That was my thought process, at least.

imiric: OK. Done in 934582db.

MattDodsonEnglish (approved):

Issue #808, opened 2022-09-14
Restructure and rewrite Scenarios section


MattDodsonEnglish:
Scenarios are among the most important parts of the k6 API. But, frequent user questions and doubts about configuring them show that the docs could be confused. Subjectively, I can say that I find them confusing. I still need to read the docs for each executor carefully to figure out what workload to use.

Some preliminary issues:
- What's a "workhouse?" Let's reduce metaphorical language and be precise about what we mean.
- Similarly, what's a "workload?" We probably need to define that.
- "When to use" is buried in the topic for each executor. Let's move that to the top page.
- On the top page, a visual comparison of each workload would be nice, perhaps in a  [small mutiple](https://en.wikipedia.org/wiki/Small_multiple). 
- Perhaps a flow chart would help users choose an executor.

Sequence to solve this issue:

1. - [ ] Read and edit workload modules in [k6 learn](https://github.com/grafana/k6-learn).
2. - [ ] Read through and copy edit docs
3. - [ ] Restructure as necessary
4. - [ ] Add content as necessary

## Relevant links

- [Release for v0.27.0](https://github.com/grafana/k6/blob/master/release%20notes/v0.27.0.md), which contained PR https://github.com/grafana/k6/pull/1007, explains the then-new executors feature.

na--:
"workhorse", not "workhouse" :joy: But yeah, I agree the docs can be improved a ton here. One suggestion I have is to add a use-case based table with "if you want to do X, you should use executor Y in your scenario".

MattDodsonEnglish:
I was going to post about metioning the coordinated omission problem, but there's already an issue:

https://github.com/grafana/k6-docs/issues/231#issue-824581069

Besides the forum post, it got brought up in discussion about a blog that used k6 to bench mark load balancers:
https://news.ycombinator.com/item?id=32867684

Really, to fix this, we could stick it somewhere in https://k6.io/docs/using-k6/scenarios/arrival-rate/. That would help googlers.

MattDodsonEnglish:
This Office Hours has some good discussion on the different use cases for each scenario:
https://www.youtube.com/watch?v=ypli_y3UryU

MattDodsonEnglish:
Related: https://github.com/grafana/k6-docs/issues/615

MattDodsonEnglish:
As I went through the old issues, I collected all the ones about scenarios. It's quite a list! I guess to close this issue, we should try and close all of these, too:

[...]

PR #936, opened 2022-12-06
Misc scenario edits

MattDodsonEnglish:
This PR could be rebased. For context, read the commit messages. Each commit is atomic and closes a minor open issue about scenarios.

There's still a good deal of work to do about helping users choose executors, but at least a k6 doc will now appear when someone searches "k6 coordinated omission".

I'm not crazy about the title "Advanced examples," either―what makes them so advanced?―but that can be fixed in a different commit.

PR #974, opened 2023-01-05
808/about executors

MattDodsonEnglish:
Part of work in #808 .

As I tried to add explanatory text about VU allocation and dropped metrics, I struggled to put the content in place. Gradually, I moved the content from being grouped together in the top-level executors page, to being separate topics, to moving them to a new section for explanatory texts about scenarios. The organization still seemed incoherent, so I moved the other explanatory pages to a new section. This information architecture is the best I design I could come up to make it easy to find information and to make the work of positioning new topics much easier.

image

This work has a few steps:
- [x] Get technical review
- [ ] Get info arch and support review
- [x] Write link fixes and redirects
- [ ] Merge


MattDodsonEnglish:
@na-- I am requesting your review first.

I only request your review on the two new pages, related to dropped iterations and VU allocation. If you want to comment on the structural changes of the docs, feel free, but I really just want your expert opinion on the accuracy and usefulness.

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
```diff
@@ -0,0 +1,89 @@
...
+title: VU allocation
+excerpt: How k6 allocates VUs in the open-model, arrival-rate executors
+---
+
+This document explains how k6 allocates VUs in the arrival-rate executors.
+
+In arrival-rate executors, three properties determine the iterations per second:
+- The `rate` property determines how many iterations k6 starts,
+- `timeUnit` determines how frequently it starts the number of iterations
+- `preAllocatedVUs` sets the number of VUs to use to reach the target iterations per second
```
MattDodsonEnglish, line 11:
```suggestion
-  `rate`  determines how many iterations k6 starts.
- `timeUnit` determines how frequently it starts the number of iterations.
- `preAllocatedVUs` sets the number of VUs to use to reach the target iterations per second.
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
```diff
@@ -0,0 +1,89 @@
...
+1. If the target still exceeds available VUs, continue allocating VUs until reaching the number set by `maxVUs`.
+
+Though it seems convenient, you should avoid using `maxVUs` in most cases.
+Allocating VUs has CPU and memory costs, and allocating VUs as the test runs **can overload the load generator and skew results**.
+If you're running in k6 Cloud, the `maxVUs` will count against your subscription.
+In almost all cases, the best thing to do is to pre-allocate the number of VUs you need beforehand.
+
+Some of the times it might make sense to use `maxVUs` include:
+- To determine necessary allocation in first-time tests
+- To add a little "cushion" to an expected baseline `maxVUs`
```
MattDodsonEnglish, line 88:
```suggestion
- To add a little "cushion" to the pre-allocated VUs that you expect the test needs
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
```diff
@@ -0,0 +1,89 @@
...
+- For huge, highly distributed tests, in which you need to carefully scale load generators as you increment VUs.
```
MattDodsonEnglish, line 89:
```suggestion
- In huge, highly distributed tests, in which you need to carefully scale load generators as you increment VUs.
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/03 Dropped iterations.md
```diff
@@ -0,0 +1,41 @@
...
+### Configuration-related iteration drops
+
+Dropped iterations happen for different reasons in different types of executors.
+
+With `shared-iterations` and `per-vu-iterations`, iterations drop if the scenario reaches its `maxDuration` before all iterations finish.
+To mitigate this, you likely need to increase the value of the duration.
+  
+With `constant-arrival-rate` and `ramping-arrival-rate`, iterations drop if there are no free VUs.
+**If it happens at the beginning of the test, you likely just need to allocate more VUs.**
+If this happens later in the test, the dropped iterations might happen because of the SUT.
```
MattDodsonEnglish, line 23:
```suggestion
If this happens later in the test, the dropped iterations might happen because of the performance of the SUT is degrading.
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/03 Dropped iterations.md
```diff
@@ -0,0 +1,41 @@
...
+
+### SUT-related iteration drops
+
+If iterations drop later in the test run, your test SUT might have become slow to respond to requests, process iterations, or both.
+
+At a certain point of high latency or longer iteration durations, k6 can no longer send VUs at the configured rate. 
+There could be a variety of causes for these dropped iterations:
```
MattDodsonEnglish, line 30:
```suggestion
At a certain point of high latency or longer iteration durations, k6 can no longer send VUs at the configured rate.
As a result, the executor will drop iterations.

The reasons for these dropped iterations varies:
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/03 Dropped iterations.md
```diff
@@ -0,0 +1,41 @@
+---
+title: Dropped iterations
+excerpt: Explanations about how your scenario configuration or SUT performance can lead to dropped iterations
+---
+
+Sometimes, a scenario can't run the expected number of iterations.
+k6 tracks this number in a counter metric, `dropped iterations`.
```
MattDodsonEnglish, line 7:
```suggestion
Sometimes, a scenario can't run the expected number of iterations.
k6 tracks the number of unsent iterations in a counter metric, `dropped iterations`.
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
```diff
@@ -0,0 +1,87 @@
...
+k6 tries to reach the target iterations per second, and one of two things can happen.
+
+| If the executor       | Then..                                                                                                                                 |
+|-----------------------|----------------------------------------------------------------------------------------------------------------------------------------|
+| Has enough VUs        | the extra VUs are "idle," ready to be used when needed.                                                                                |
+| Has insufficient VUs. | k6 emits a [`dropped_iterations` metric](/using-k6/scenarios/about-scenarios/dropped-iterations) for each iteration that it can't run. |
+
+## Iteration duration affects the necessary allocation
+
+The necessary allocation depends on the iterations duration.
```
MattDodsonEnglish, line 53:
```suggestion
k6 tries to reach the target iterations per second,
and one of two things can happen:

| If the executor       | Then..                                                                                                                                 |
|-----------------------|----------------------------------------------------------------------------------------------------------------------------------------|
| Has enough VUs        | the extra VUs are "idle," ready to be used when needed.                                                                                |
| Has insufficient VUs. | k6 emits a [`dropped_iterations` metric](/using-k6/scenarios/about-scenarios/dropped-iterations) for each iteration that it can't run. |

## Iteration duration affects the necessary allocation

The necessary allocation depends on the iteration duration:
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
```diff
@@ -0,0 +1,87 @@
...
+Longer durations need more VUs.
+
+In a "perfect world", you could estimate the number of pre-allocated VUs with this formula
```
MattDodsonEnglish, line 56:
```suggestion
In a perfect world, you could estimate the number of pre-allocated VUs with this formula:
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
```diff
@@ -0,0 +1,87 @@
...
+
+```
+preAllocatedVUs = [median_iteration_duration * rate] + constant_for_variance
+```
+
+In the real world, if you know _exactly_ how long an iteration take, you likely don't need to run a test.
+What's more, as the test goes on, iteration duration will likely increase.
```
MattDodsonEnglish, line 63:
```suggestion
What's more, as the test goes on, iteration duration likely increases.
```

MattDodsonEnglish (commented):
Proofreading

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios.md
```diff
@@ -56,12 +56,13 @@ export const options = {
...
 For each k6 scenario, the VU workload is scheduled by an _executor_.
 For example, executors configure:
+- How long the test runs
 - Whether VU traffic stays constant or changes
 - Whether to model traffic by iteration number or by VU arrival rate.
 
 Your scenario object must define the `executor` property with one of the predefined executors names.
 Along with the generic scenario options, each executor object has additional options specific to its workload.
-For the list of the executors, refer to the [Executor guide](/using-k6/scenarios/executors/).
+For the list of the executors, refer [Executors](/using-k6/scenarios/executors/).
```
na--, line 65: hmm I don't think I've seen "refer something" in a sentence before, so it doesn't sound quite ok :thinking: shouldn't it be "see something" or "refer to something"?

```suggestion
For the list of the executors, see [Executors](/using-k6/scenarios/executors/).
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios.md
```diff
@@ -0,0 +1,18 @@
+---
+title: "About scenarios"
+excerpt: High-level explanations about how your executor configuration can change the test execution and test results
+---
+
+These topics explain the essential concepts of how scenarios and their executors work.
+
+Different scenario configurations can affect many different aspects of your system,
+including the generated load, utilized resources, and emitted metrics.
+If you know a bit about how scenarios work, you'll both design better tests for resources and goals, and interpret test results with more understanding.
```
na--, line 10: I am not sure I understand "design better tests for resources and goals"

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/01 Executors/06 ramping-arrival-rate.md
```diff
@@ -11,6 +11,14 @@ k6 will attempt to dynamically change the number of VUs to achieve the configure
 
 See the [arrival rate](/using-k6/scenarios/arrival-rate) section for details.
 
+<Blockquote mod="Note" title="Iteration starts are spaced fractionally">
+
+Iterations **do not** started at exactly the same time.
+At a `rate` of 10 with a `timeUnit` of `1s`, each iteration starts about every tenth of a second (that is, each 100ms).
+This jitter accidentally concurrency where all VUs start at the exact millisecond
```
na--, line 18: Something is wrong with this sentence, but I am not sure exactly what :sweat_smile:

MattDodsonEnglish: Oh I know what's wrong: it's not a sentence. It has no main verb. It should be "This jitter prevents." Fixing now...

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
```diff
@@ -0,0 +1,88 @@
+---
+title: VU allocation
```
na--, line 2: "VU allocation" seems a bit misleading, since we are only talking about arrival-rate scenarios

```suggestion
title: Arrival-rate VU allocation
```

or maybe even 

```suggestion
title: Arrival-rate configuration
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios.md
MattDodsonEnglish: The thinking was that if you understand scenarios better, you can:

- Make better decisions for your test goals, because certain scenarios correspond better to certain test design. To spike test a single component for raw throughput probably requires an arrival rate executor. To just see how quickly your system can churn through x number of iterations, `shared iterations` is a simpler choice.
- Use resources better, because understanding not to use maxVUs means you'll use CPU cycles more efficiently.

Of course now I realize that this is an enormous amount of implied information.

Will change to:
>If you know a bit about how scenarios work, you'll design better tests and interpret test results with more understanding.

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios.md
MattDodsonEnglish, line 65:
```suggestion
For the list of the executors, refer to [Executors](/using-k6/scenarios/executors/).
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios.md
MattDodsonEnglish, line 10:
```suggestion
If you know a bit about how scenarios work, you'll design better tests and interpret test results with more understanding.
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
MattDodsonEnglish: Need to think about this.
"Arrival-rate configuration" probably opens the door to more topics, like using options. That's not bad, but it may mean there's a better place to put this (not blocking for this PR). Is there any reason readers should know about non-arrival rate allocation? If so, maybe we could add it. If not, maybe it doesn't matter to document.

Not sure. It's a good point though.

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/01 Executors/06 ramping-arrival-rate.md
MattDodsonEnglish, line 18:
```suggestion
This jitter prevents accidental concurrency where all VUs start at the exact millisecond
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/01 Executors/05 constant-arrival-rate.md
```diff
@@ -15,6 +15,14 @@ useful for a more accurate representation of RPS, for example.
 
 See the [arrival rate](/using-k6/scenarios/arrival-rate) section for details.
 
+<Blockquote mod="Note" title="Iteration starts are spaced fractionally">
+
+Iterations **do not** started at exactly the same time.
+At a `rate` of 10 with a `timeUnit` of `1s`, each iteration starts about every tenth of a second (that is, each 100ms).
+This jitter accidentally concurrency where all VUs start at the exact millisecond
```
MattDodsonEnglish, line 22:
```suggestion
This jitter prevents accidental concurrency where all VUs start at the exact millisecond
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
```diff
@@ -0,0 +1,88 @@
...
+excerpt: How k6 allocates VUs in the open-model, arrival-rate executors
+---
+
+This document explains how k6 allocates VUs in the arrival-rate executors.
+
+In arrival-rate executors, three properties determine the iterations per second:
+-  `rate`  determines how many iterations k6 starts.
```
na--, line 9: Some more explanation is probably needed here :thinking: Something like this? Or is this too much?

```suggestion
The desired load in scenarios with arrival-rate executors is not configured in terms of how many VUs will loop through the iteration code in parallel. Instead, we need to specify at what interval k6 should try to start a new iteration, regardless of how many other iterations (and VUs) are currently running - the so-called ["open model"](/using-k6/scenarios/about-scenarios/open-vs-closed/#open-model). Or, put another way, we need to specify how many iterations per second (or per minute, hour, etc.) k6 should try and start.

Still, iterations need VUs to run on and k6 VUs, like other JavaScript runtimes, are single-threaded and can only run a single iteration (with its event loop) at any given time. This makes configuring scenarios with arrival-rate executors more complicated. We need to configure both the desired load, and to ensure we have enough pre-allocated initialized VUs to handle it. Which may be tricky to calculate, considering that the iteration duration depends entirely on what the script does and how quickly or slowly the system-under-test (SUT) responds. Slower SUT means slower iterations, which means more VUs will be needed, since some will be busy with running old iterations.

You can specify the desired arrival rate with the `rate` and `timeUnit` scenario options - k6 will try to start `rate` iterations evenly spread across a `timeUnit` (default `1s`) interval of time. So, given a scenario with a `constant-arrival-rate` executor and `rate: 10`, k6 will try to start one iteration every 100 milliseconds. However, if the scenario had `rate: 10, timeUnit: '1m'`, k6 will try to start a new iteration every 6 seconds.

VUs can be configured via the `preAllocatedVUs` option - it sets the number of VUs k6 will initialize before the test runs. There is also the `maxVUs` option, which allows k6 to initialize VUs mid-test, though that is mostly useful during development and generally is a bad practice to use in actual tests. VU initialization is somewhat CPU intensive and doing it in the middle of the test could potentially interfere with test execution or skew results.
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
```diff
@@ -0,0 +1,88 @@
...
+- `timeUnit` determines how frequently it starts the number of iterations.
+- `preAllocatedVUs` sets the number of VUs to use to reach the target iterations per second.
+
+In short, while `rate` and `timeUnit` set the target iterations per second, **you must allocate enough VUs to reach this target.**
+By pre-allocating VUs, k6 can initialize the VUs necessary for the test before the test runs,
+which ensures that the CPU cost of allocation doesn't interfere with test execution.
```
na--, line 15: Some more explanation is probably needed here :thinking: Something like this?

```suggestion
The desired load in scenarios with arrival-rate executors is not configured in terms of how many VUs will loop through the iteration code in parallel. Instead, we need to specify at what interval k6 should try to start a new iteration, regardless of how many other iterations (and VUs) are currently running - the so-called ["open model"](/using-k6/scenarios/about-scenarios/open-vs-closed/#open-model). Or, put another way, we need to specify how many iterations per second (or per minute, hour, etc.) k6 should try and start.

Still, iterations need VUs to run on and k6 VUs, like other JavaScript runtimes, are single-threaded and can only run a single iteration (with its event loop) at any given time. This makes configuring scenarios with arrival-rate executors more complicated. We need to configure both the desired load, and to ensure we have enough pre-allocated initialized VUs to handle it. Which may be tricky to calculate, considering that the iteration duration depends entirely on what the script does and how quickly or slowly the system-under-test (SUT) responds. Slower SUT means slower iterations, which means more VUs will be needed, since some will be busy with running old iterations.

You can specify the desired arrival rate with the `rate` and `timeUnit` scenario options - k6 will try to start `rate` iterations evenly spread across a `timeUnit` (default `1s`) interval of time. So, given a scenario with a `constant-arrival-rate` executor and `rate: 10`, k6 will try to start one iteration every 100 milliseconds. However, if the scenario had `rate: 10, timeUnit: '1m'`, k6 will try to start a new iteration every 6 seconds.

VUs can be configured via the `preAllocatedVUs` option - it sets the number of VUs k6 will initialize before the test runs. There is also the `maxVUs` option, which allows k6 to initialize VUs mid-test, though that is mostly useful during development and generally is a bad practice to use in actual tests. VU initialization is somewhat CPU intensive and doing it in the middle of the test could potentially interfere with test execution or skew results.
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/03 Dropped iterations.md
```diff
@@ -0,0 +1,41 @@
...
+With `shared-iterations` and `per-vu-iterations`, iterations drop if the scenario reaches its `maxDuration` before all iterations finish.
+To mitigate this, you likely need to increase the value of the duration.
+  
+With `constant-arrival-rate` and `ramping-arrival-rate`, iterations drop if there are no free VUs.
+**If it happens at the beginning of the test, you likely just need to allocate more VUs.**
+If this happens later in the test, the dropped iterations might happen because SUT performance is degrading and iterations are taking longer to finish.
+
+### SUT-related iteration drops
+
+At a certain point of high latency or longer iteration durations, k6 can no longer send VUs at the configured rate.
```
na--, line 27:
```suggestion
At a certain point of high latency or longer iteration durations, k6 will no longer have free VUs to start iterations with at the configured rate.
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/03 Dropped iterations.md
```diff
@@ -0,0 +1,41 @@
...
+As a result, the executor will drop iterations.
+
+The reasons for these dropped iterations vary:
+- The SUT response has become so long that k6 starts dropping scheduled VUs from the queue.
```
na--, line 31:
```suggestion
- The SUT response has become so long that k6 starts dropping scheduled iterations from the queue.
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/03 Dropped iterations.md
```diff
@@ -0,0 +1,41 @@
...
+- The SUT iteration duration has become so long that k6 needs to schedule more VUs to reach the target arrival rate, exceeding the number of scheduled iterations.
+- Some network errors between the generator and the SUT have caused iterations to drop.
```
na--, line 33:
```suggestion
```

This is not a thing that happens :sweat_smile: it might lead to interrupted iterations (e.g. an iteration that was already started but need to be stopped because we ran out of time), but we don't track these yet :disappointed:

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/01 Executors/05 constant-arrival-rate.md
```diff
@@ -15,6 +15,14 @@ useful for a more accurate representation of RPS, for example.
 
 See the [arrival rate](/using-k6/scenarios/arrival-rate) section for details.
 
+<Blockquote mod="Note" title="Iteration starts are spaced fractionally">
+
+Iterations **do not** started at exactly the same time.
+At a `rate` of 10 with a `timeUnit` of `1s`, each iteration starts about every tenth of a second (that is, each 100ms).
+This jitter prevents accidental concurrency where all VUs start at the exact millisecond
```
na--, line 22:
```suggestion
Iterations **do not** start at exactly the same time.
At a `rate` of `10` with a `timeUnit` of `1s`, each iteration starts about every tenth of a second (that is, each 100ms).
This jitter prevents accidental concurrency where all VUs start at the exact millisecond.
```

I am not actually sure about whether we should include the last sentence :thinking: the even spacing of iterations across the `timeUnit` interval is not exactly "jitter" and we might even add an option to group iteration starts in "buckets" (see 2nd point in https://github.com/grafana/k6/issues/1386)

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
```diff
@@ -0,0 +1,88 @@
...
+      timeUnit: "1m",
+    },
+  },
+};
+```
+
+
+<Blockquote mod="attention" title="">
+
+In cloud tests, **both `preAllocatedVUs` and `maxVUs` count against your subscription.**
```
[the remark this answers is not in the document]
MattDodsonEnglish: I don't want to move the whole block down, because it's in an admonition and I think it's most respectful if we alert readers about things that can affect subscription use up front. But I'll move your addition to that section.

image

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/01 Executors/05 constant-arrival-rate.md
MattDodsonEnglish: Sounds fine, will delete. I think "do not start at the same time" gets the point across.

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/01 Executors/06 ramping-arrival-rate.md
```diff
@@ -11,6 +11,14 @@ k6 will attempt to dynamically change the number of VUs to achieve the configure
 
 See the [arrival rate](/using-k6/scenarios/arrival-rate) section for details.
 
+<Blockquote mod="Note" title="Iteration starts are spaced fractionally">
+
+Iterations **do not** started at exactly the same time.
+At a `rate` of 10 with a `timeUnit` of `1s`, each iteration starts about every tenth of a second (that is, each 100ms).
+This jitter prevents accidental concurrency where all VUs start at the exact millisecond
+
```
MattDodsonEnglish, line 19:
```suggestion
Iterations **do not** start at exactly the same time.
At a `rate` of `10` with a `timeUnit` of `1s`, each iteration starts about every tenth of a second (that is, each 100ms).

```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
MattDodsonEnglish, line 15:
```suggestion
In arrival-rate executors, as long as k6 has VUs available, it starts iterations according to your target rate.
The ability to set iteration rate comes with a bit more configuration complexity: you must pre-allocate a sufficient number of VUs.
In other words, before the tests runs, you must both:
- Configure load (as iteration per unit of time)
- Ensure that you've scheduled enough VUs.

Read on to learn about how k6 allocates VUs in the arrival-rate executors.

## Pre-allocation in arrival-rate executors

As [open-model](/using-k6/scenarios/about-scenarios/open-vs-closed/#open-model) scenarios, arrival-rate executors start iterations according to a configured rate.
For example, you can configure arrival-rate executors to start 10 iterations each second, or minute, or hour.
This behavior is opposed to the closed-model scenarios, in which VUs wait for one iteration to finish before starting another

Each iteration need needs a VU to run it.
Because k6 VUs are single threaded, like other JS runtimes, a VU can only run the event loop of a single iteration at a time.
To ensure you have enough, you must pre-allocate a sufficient number.

In your arrival-rate configuration, three properties determine the iteration rate:
- `rate` determines how many iterations k6 starts.
- `timeUnit` determines how frequently it starts the number of iterations.
- `preAllocatedVUs` sets the number of VUs to use to reach the target iterations per second.
In practice, determining the right number of iterations might take some trial and error,
as the necessary VUs entirely depends on how quickly the SUT can process iterations.
```

MattDodsonEnglish: My initial thought is that the info is good...but dense.

One thing, I see that I was just wrong to type per second, it should be iteration rate.

 I also think a short bulleted intro should stay. Maybe not. This creates a [progressive disclosure](https://www.nngroup.com/articles/progressive-disclosure/), where readers who quickly scan could get the minimal info they need. Opening with four paragraphs in a row might scare many off.
 
 I've tried to synthesize your text and mine with:
 https://github.com/grafana/k6-docs/pull/974/commits/977c45d34e25c0defb079ef2c9f7bbb1e7279a24
 
 I wrote it sleepy, need to proofread, but let me know what you think.

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
MattDodsonEnglish: Now that I've thought about it, I don't want to go with "Arrival-rate configuration" because that should include Graceful stop and maybe more, which opens a whole new round of content structure. It's nice to keep info atomic.

Is VU allocation in non-arrival-rate ever important? If so, we could just add it to the doc later.

Either way, I choose these new titles, ranked by preference. You can pick and that's what will go with:
1. VU allocation.
2. VU pre-allocation
3. Arrival-rate VU allocation

na--: > because that should include Graceful stop and maybe more

graceful stop is not specific to arrival-rate executors and already has its own dedicated page that explains it: https://k6.io/docs/using-k6/scenarios/graceful-stop/

> Is VU allocation in non-arrival-rate ever important? If so, we could just add it to the doc later.

Well, considering the configuration of non-arrival-rate executors is specified in terms of VUs, there isn't really anything complicated to explain there :sweat_smile: 

Again, the complexity with arrival-rate is not just how VUs are allocated, but how to balance the VUs and the desired rate and how to find the right values for the former based on the latter and on iteration duration.

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
```diff
@@ -0,0 +1,102 @@
+---
+title: VU allocation
+excerpt: How k6 allocates VUs in the open-model, arrival-rate executors
+---
+
+In arrival-rate executors, as long as k6 has VUs available, it starts iterations according to your target rate.
+The ability to set iteration rate comes with a bit more configuration complexity: you must pre-allocate a sufficient number of VUs.
+In other words, before the tests runs, you must both:
+- Configure load (as iteration per unit of time)
```
na--, line 9:
```suggestion
- Configure load (as new iterations per unit of time)
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
```diff
@@ -0,0 +1,102 @@
...
+- Ensure that you've scheduled enough VUs.
```
na--, line 10: "Scheduled" is incorrect here, "allocated" or "initialized"

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
```diff
@@ -0,0 +1,102 @@
...
+Read on to learn about how k6 allocates VUs in the arrival-rate executors.
+
+## Pre-allocation in arrival-rate executors
+
+As [open-model](/using-k6/scenarios/about-scenarios/open-vs-closed/#open-model) scenarios, arrival-rate executors start iterations according to a configured rate.
+For example, you can configure arrival-rate executors to start 10 iterations each second, or minute, or hour.
+This behavior is opposed to the closed-model scenarios, in which VUs wait for one iteration to finish before starting another
+
+Each iteration need needs a VU to run it.
+Because k6 VUs are single threaded, like other JS runtimes, a VU can only run the event loop of a single iteration at a time.
```
na--, line 21:
```suggestion
Because k6 VUs are single threaded, like other JavaScript runtimes, a VU can only run a single iteration (and its event loop) at a time.
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
```diff
@@ -0,0 +1,102 @@
...
+To ensure you have enough, you must pre-allocate a sufficient number.
+
+In your arrival-rate configuration, three properties determine the iteration rate:
+- `rate` determines how many iterations k6 starts.
+- `timeUnit` determines how frequently it starts the number of iterations.
```
na--, line 26: Having these 2 in different lines, with a separate explanation for each, is more confusing than helpful in my opinion. The current disjointed explanation is both longer, more confusing and less correct than "k6 will try to start `rate` iterations evenly spread across a `timeUnit` (default `1s`) interval of time"

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
na--: bullet points are fine, except when they force you to unnecessarily break up explanations so much that they stop making sense and are harder to understand, e.g. https://github.com/grafana/k6-docs/pull/974/files#r1069127202

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
```diff
@@ -0,0 +1,102 @@
...
+- `preAllocatedVUs` sets the number of VUs to use to reach the target iterations per second.
+In practice, determining the right number of iterations might take some trial and error,
+as the necessary VUs entirely depends on how quickly the SUT can process iterations.
```
na--, line 29:
```suggestion
- `preAllocatedVUs` sets the number of VUs k6 will initialize before the test starts.
In practice, determining the right number of VUs to be able to reach the target iterations rate might take some trial and error,
as the necessary VUs entirely depends on what your iteration code looks like and how quickly the system-under-test can process requests.
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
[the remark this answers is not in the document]
na--: I just realized this might be confusing in another way, "both `preAllocatedVUs` and `maxVUs` count against your subscription" may confuse users that `preAllocatedVUs + maxVUs` will be counted against their subscription, while the reality is that just the bigger number between the two (i.e. `max(preAllocatedVUs, maxVUs)`) will be counted.

And, given that `maxVUs` will always be `>= preAllocatedVUs`, and if you don't specify `maxVUs` explicitly, it's implicitly equal to `maxVUs`, `maxVUs` is what will be counted against the subscription

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
```diff
@@ -0,0 +1,102 @@
...
+      timeUnit: "1m",
+    },
+  },
+};
+```
+
+
+<Blockquote mod="attention" title="">
+
+In cloud tests, **both `preAllocatedVUs` and `maxVUs` count against your subscription.**
```
na--, line 47: Another problem - you mention `maxVUs` here, but we haven't mentioned maxVUs anywhere before in this document... This was one of the reasons for my longer explanation that you discarded, it had a paragraph with a cohesive explanation for both `preAllocatedVUs` and `maxVUs`

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
```diff
@@ -0,0 +1,102 @@
...
+The arrival-rate executors also have a `maxVUs` property.
+If you set it, k6 runs in this sequence:
+1. Pre-allocate the `preAllocatedVUs`.
+1. Run the test, trying to reach the target iteration rate.
+1. If the target exceeds the available VUs, allocate another VU.
+1. If the target still exceeds available VUs, continue allocating VUs until reaching the number set by `maxVUs`.
+
+Though it seems convenient, you should avoid using `maxVUs` in most cases.
+Allocating VUs has CPU and memory costs, and allocating VUs as the test runs **can overload the load generator and skew results**.
+If you're running in k6 Cloud, the `maxVUs` will count against your subscription.
```
MattDodsonEnglish, line 96:
```suggestion
If you're running in k6 Cloud, the `maxVUs` will count against your subscription.
This is because k6 must allocate sufficient resources for `maxVUs` to be initialized, even if they never are.
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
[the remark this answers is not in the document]
MattDodsonEnglish: Maybe:

> The number of VUs you allocate (including `maxVUs`) counts against your subscription.

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
MattDodsonEnglish: I'm still kind of of the opinion that `maxVUs` shouldn't be mentioned anywhere :-). What I'll do is just make the first admonition only about `preAllocatedVUs`, and then a second admonition in the `maxVUs` section. It's not very elegant, but it doesn't cram so much information together.

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
MattDodsonEnglish: > Again, the complexity with arrival-rate is not just how VUs are allocated, but how to balance the VUs and the desired rate and how to find the right values for the former based on the latter and on iteration duration.

Is this not encompassed in `VU Pre-allocation`? Basically, I'm looking for the shortest way to say the most in the most accurate way.

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
```diff
@@ -0,0 +1,103 @@
+---
+title: VU allocation
+excerpt: How k6 allocates VUs in the open-model, arrival-rate executors
+---
+
+In arrival-rate executors, as long as k6 has VUs available, it starts iterations according to your target rate.
+The ability to set iteration rate comes with a bit more configuration complexity: you must pre-allocate a sufficient number of VUs.
+In other words, before the tests runs, you must both:
+- Configure load (as iteration per unit of time)
+- Ensure that you've scheduled enough VUs.
```
MattDodsonEnglish, line 10:
```suggestion
- Ensure that you've allocated enough VUs.
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
na--: Again, if I could rewrite history, maxVUs probably would not exist :sweat_smile:  But now that it exists, we should try to make it as clear as possible how it function and why it might not be a good idea to use. 

An admonition only for `preAllocatedVUs` doesn't make sense, this is how cloud subscriptions normally work, i.e. no admonition needed just for it.

And if we want to tuck `maxVUs` only at the end of the document, somewhat out of sight (which I don't necessarily mind), then we should only have an admonition there.

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
na--: "VU Pre-allocation" only makes sense to you because you already know it applies for arrival-rate executors. A new user won't know that fact and won't click on that menu entry at all, even if this is exactly the information they are looking for.

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 About scenarios/02 VU allocation.md
MattDodsonEnglish: Here's how I've handled it: two admonitions. No points for subtlety, but I don't think anyone can say we're being sneaky:

image


image

na--: Nobody will read the text you have highlighted in the middle of that paragraph :sweat_smile: People will just see the two admonitions and get the impressions that in the cloud we will charge them for `preAllocatedVUs + maxVUs`

MattDodsonEnglish: Point taken. With https://github.com/grafana/k6-docs/pull/974/commits/19784d4173ca6376585990b447f1b39cbcc58706, it looks like this:

image

MattDodsonEnglish:
@ppcano , na-- and I have done a pretty big job on the two new docs, and I don't think anyone has the desire to get to in to them anymore, but would you mind doing a quick review of the info architecture changes?

There's no doubt that there's more work to do to organize this information, but I think it can wait for more PRs.

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios.md
```diff
@@ -56,20 +56,21 @@ export const options = {
...
 For each k6 scenario, the VU workload is scheduled by an _executor_.
 For example, executors configure:
+- How long the test runs
 - Whether VU traffic stays constant or changes
 - Whether to model traffic by iteration number or by VU arrival rate.
 
 Your scenario object must define the `executor` property with one of the predefined executors names.
 Along with the generic scenario options, each executor object has additional options specific to its workload.
-For the list of the executors, refer to the [Executor guide](/using-k6/scenarios/executors/).
+For the list of the executors, refer to [Executors](/using-k6/scenarios/executors/).
```
ppcano, line 65: @MattDodsonEnglish I think we should list here all the executors. If not, readers might move to Concepts without getting the "general" idea of the different executor options.

I suggest adding something similar than the [Executors table](https://k6.io/docs/using-k6/scenarios/executors/). For example:

- [Shared iterations](https://k6.io/docs/using-k6/scenarios/executors/shared-iterations): a fixed amount of iterations are "shared" between a number of VUs. 
- [Per VU iterations](https://k6.io/docs/using-k6/scenarios/executors/per-vu-iterations): each VU executes an exact number of iterations. 
- ...

ppcano (approved):
added a comment with a suggestion

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios.md
MattDodsonEnglish: @ppcano I deleted that list in https://github.com/grafana/k6-docs/pull/936. I didn't like duplicating content in this way.

I'll make a new PR to put it back in, maybe just as a summary:

You can configure executors in to distribute workload according to:
- Iterations. Either shared by the VUs, or distributed across them.
- VUs. Either a constant number or a ramping number.
- Iterations per second. Either constant or ramping.

MattDodsonEnglish: >I'll make a new PR to put it back in

#985

folex:
Could you please clarify a little bit regarding connection between Open Model and Dropped Iterations?

The Test lifecycle doc states the following:
> The default function life-cycle
> A VU executes the default() function from start to end in sequence. Once the VU reaches the end of the function, it loops back to the start and executes the code all over.

But doesn't it contradict the Open Model? Which states:
> Compared to the closed model, the open model decouples VU iterations from the iteration duration. The response times of the target system no longer influence the load on the target system.

As I understand it, to avoid coordinated omission, the VU's `default()` function should be called on a strict schedule, every N ms, without waiting for `default()` to finish.

Now, there's also a question of what it means for a VU to be "free" or "busy".

The Dropped Iterations doc states the following:
> With constant-arrival-rate and ramping-arrival-rate, iterations drop if there are no free VUs. 

Here I don't really understand what "free VU" or "busy VU" means. Is a VU that has an iteration in progress considered "busy"?  How does that interplay with Open Model, which must run VU's function on a strict schedule?

----

Another question about free/busy arises from the fact that `default()` is an async function. So there's some time taken to created Promise, and then there's some time taken for that Promise to be completed. Which of these is considered to be "iteration time"?

Dropped Iterations article sheds some light on that:
> The reasons for these dropped iterations vary:
> The SUT response has become so long that k6 starts dropping scheduled iterations from the queue.
> The SUT iteration duration has become so long that k6 needs to schedule more VUs to reach the target arrival rate, exceeding the number of scheduled iterations.

So there's clearly a separation between "iteration duration" and "response". But how exactly are these defined? Is it related to `default()`'s Promise creation and Promise completion, or is it something else entirely?

----

Hope my questions make sense! And thank you very much for the amazing work you are doing here, systemazing and documenting benchmark approaches is super important!

MattDodsonEnglish:
Hey @folex , these are great questions; maybe you should make an issue. I no longer work on the k6 docs, and I don't know if anyone else will see this thread.

But I'm still a k6 fan, and I did write the sentences you quoted, so I'll try to comment on a few things.

- About the lifecycle: I think you're definitely right: that sentence describes the closed model. Maybe it could be re-written: "Depending on the configuration, VUs either execute in loops, from start to finish, or at set rates of time."

- About "busy" VUs. I think this just means they are running an iteration. Have you by chance read this doc? It might help a little.    https://k6.io/docs/using-k6/scenarios/concepts/arrival-rate-vu-allocation/

     > Before an arrival-rate scenario starts, k6 first initializes the number of preAllocatedVUs. When the test runs, the number of available preAllocatedVUs determines how many iterations k6 can start. k6 tries to reach the target iterations per second, and one of two things can happen: [they are idle; or all are busy and k6 emits a drop] 
     
I can't remember the difference between iteration time and response time, but I think each iteration could make many requests, all with unique response durations. The iteration duration is the entire VU code, including all requests, sleep time, console logs, etc.


     
About the async stuff, I have no idea. I defer to @mstoykov  :bow:

mstoykov:
Hi @folex, 

A busy VU is one that is executing an iteration. A free one is one that isn't executing one ;). 

An iteration started ends only ones it has no async jobs to finish irregardless of if the `default` function is `async` or just used `setTimeout` or used pure promises with no `async`/`await`. 

This is predominantly as anything else would've been really confusing.

```javascript
export default function() {
  http.asyncRequest("GET", someurl).then((res) => {
    // do something
  })
}
```

Should this code with ... `ramping-vus` executor just keep starting `asyncRequest` request as it gets to the end of the iteration?

The answer we(me specifically) choose that this isn't really a good idea and instead k6 waits for everything async to finish. It already needed to wait for them at the end of k6 run for example - so it made sense to not "contaminate" between iterations. 

So in this case a busy VU is busy even if it is waiting on async operation to finish. 

And as such an iteration and its duration is the whole time between the iteration starting and the *last* if any async operation finishing.  

As @MattDodsonEnglish pointed out iterations can have multiple responses - or none. K6 just happens to be primarily used in cases where there are things such as request and responses, but this is not really a requirement.

> As I understand it, to avoid coordinated omission, the VU's default() function should be called on a strict schedule, every N ms, without waiting for default() to finish.

While in theory this is true - reality makes it impossible to just continue to "start" stuff - which is why if we can't, we tell users that we dropped iterations - which means that the test should probably be marked failed.

Even outside of k6 this isn't really possible - you will run out of resource - memory, CPU, file descriptors and so on.

In the case of k6 another thing that is limiting is k6 runs JS - js is a single-threaded language by specification. As such in order for k6 to run an iteration it needs a js VM that is not running any at the moment. Which is basically what a VU is.

In theory, you can keep making VUs, but:
1. you then hit the above problems
2. my experience helping users has shown that the moment you start doing that you usually end up just doing less and less iterations. As more and more resources go towards making more VUs then actually doing the requests.
3. usually shows that something is very wrong.


Part of what k6 does is run the *same* test so you can compare and contrast. As such if the test used to pass with X preallocatedVUs, and now it is dropping a bunch of iterations - that is probably pretty bad. 

On the other hand there was some discussion around letting arrival rate specifically running multiple async default functions. But that has not been implemented and likely will need some limits as well - as just running 2000 default iterations in the same VU will likely not run great.

Additionally, at the time this was discussed we were just adding async code and still mostly did not have many uses for it. 

Not certain if I answered all your questions, but I hope this helps. :bow: 

Also commenting on merged PRs isn't the best idea. Maybe try the [community forum](https://community.grafana.com/c/grafana-k6/70) as I very much by chance saw that someone asked me something on a merged PR.

PR #984, opened 2023-01-16
feat: replaced algolia search with algolia docsearch

yasell:
This PR replaces Algolia search with [Algolia DocSearch](https://docsearch.algolia.com/)


ppcano:
@yasell,  this PR has the WIP label, is is ready from your side?

1 - DocSearch UI looks awesome.

2 - TODO: The search results should not return content from previous API versions. 


Screenshot 2023-02-03 at 11 52 20




Screenshot 2023-02-03 at 11 53 04


This is populating results from:

Screenshot 2023-02-03 at 11 53 23


3 - 

@bolotskydev, recently, we have discussed how to improve the search results returned by Algolia. It is a recurring issue whose importance is growing.    

You and @MattDodsonEnglish have researched a bit this. I expect we'll work on it soon. 

Before migrating, I want to double check that the new implementation will NOT prevent any future optimizations. 

I have checked the DocSearch admin panel and is the same than the admin panel of the custom implementation. At first glance, both looked the same. 

Kirill, please, can you confirm us that DocSearchalso allows the same types of optimizations we want to implement?

yasell:
Hi @ppcano!

This PR is not quite finished yet. We want to improve the look a bit and think about the logic for the crawler. The settings are now completely default.

bolotskydev:
@ppcano yes, the controls are the same, but DocSearch uses many more parameters for ranking, so any tweak to the default configuration shall be made consciously because it can ruin many things. However, on the first sight, besides the versioning issue you mentioned, the results looking promising as their relevance improved significantly - for example, we don't need to set up `synonyms` like we did previously for `metrics <-> custom metrics` , the results are aligned 👌   

I and @yasell still looking into DocSearch config while doing the migration, but no caveats so far

heitortsergent:
Closing this PR since we'll be moving the docs soon.

PR #992, opened 2023-01-19
Update "Running large tests" article

### LINKED ISSUE #208: Update "Running Large Tests" article
With several new features in k6, it's time to re-run the performance tests and update the results. We should be able to squeeze significantly more requests/VUs from a single machine now.

imiric:
This is an update to the ["Running large tests" article](https://k6.io/docs/testing-guides/running-large-tests/), and closes #208. I'm creating the PR to get early feedback on the changes, but let's not merge it yet until the benchmarking is finished.

See the commits for details, but some of the major changes include:

- Removal of the benchmark results section.
  We plan to move all benchmark results to the [k6-hardware-benchmark](https://github.com/grafana/k6-hardware-benchmark/) repository, so that we can automate running the tests and generating the report. This article will contain only general suggestions, which won't need to be updated as often, and will also make it shorter and more readable.

  I'm currently running the benchmarks, and will update the repository once we have the results.

- Removal of the suggestion of `--compatibility-mode=base`. See 2ffca0a4 for explanation.


immavalls (approved):
Great updates @imiric, just added some minor suggestions. @MattDodsonEnglish should probably review as he is the k6 tech writing expert 😄

mstoykov (approved):

MattDodsonEnglish:
A rendered version in staging is here:

https://mdr-ci.staging.k6.io/docs/refs/pull/992/merge/testing-guides/running-large-tests/

FYI, I'm going to only read it and see if I have any structural changes. I'll wait to do copy edits till the final version, since that's the most obvious and mechanical work.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -236,191 +221,46 @@ If you run into errors during the execution, it's good to understand if they wer
...
 
-- Maximum VUS reached: 30.000
-- Memory used: ~120 GB (out of 370 available)
-- CPU load (avg): ~45 (out of 96.0).
-- Peak RPS: ~61.500.
-- `sleep(1)` in each iteration.
+We have executed a few large tests on different EC2 machines to see how much load k6 can generate. Our general observation is that k6 scales proportionally to the hardware. A machine with 2x more resources can generate roughly 2x more traffic. The limit to this scalability is in the number of open connections. A single Linux machine can open up to 65,535 sockets per IP. This means that a maximum of 65k requests can be sent simultaneously on a single machine. The RPS limit depends on the response time of the SUT. If responses are delivered in 100ms, the RPS limit is 650,000.
 
-### Testing for RPS
+We maintain a [repository](https://github.com/grafana/k6-hardware-benchmark) with some scripts used to benchmark k6 and create reports. These tests are run for every new k6 version, and you can see the results in the [`results/` directory](https://github.com/grafana/k6-hardware-benchmark/tree/master/results).
```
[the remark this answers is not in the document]
imiric: Nice catch, but like I mentioned in the PR description, we shouldn't merge this before the benchmarking is finished. That will happen after the results are added to the k6-hardware-benchmark repo, so this link will eventually work.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -5,18 +5,17 @@ excerpt: 'How to run large-scale k6 tests without distributed-execution'
...
 
-k6 is different from many other load testing tools in the way it handles hardware resources. A single k6 process will efficiently use all CPU cores on a load generator machine.
-A single instance of k6 is often enough to generate load of 30,000-40,000 simultaneous users (VUs). This number of VUs can generate upwards of 300,000 requests per second (RPS).
+k6 is different from many other load testing tools in the way it handles hardware resources. A single k6 process will efficiently use all CPU cores on a load generator machine. Depending on the available resources, and taking into consideration some of the advice below, a single instance of k6 is able to run 30,000-40,000 simultaneous users (VUs). In some cases this number of VUs can generate up to 300,000 HTTP requests per second (RPS).
 
 Unless you need more than 100,000-300,000 requests per second (6-12M requests per minute), a single instance of k6 will likely be sufficient for your needs.
 
 Below we will explore what hardware and considerations are needed for generating different levels of load.
 
 ## OS fine-tuning
```
MattDodsonEnglish, line 16: As a casual, I feel uncomfortable running these commands.

These processes require on root and it looks like they modify kernal interaction and pre-defined user resource limits. As a reader, I think "why are these limits here? Am I doing something dangerous?" As a writer, I think, "Am I encouraging readers to copy and paste something without thinking about potentially serious ramifications?"

I looked at the "Fine Tuning OS" and see there's a bit more guidance. Still, though this section links that page, I'd suggest putting an admonition in this section.

```
<Blockquote mod="attention" title="">

The default limits protect your system from programs that might leak memory in huge quantities.
Consider increasing the limits only for the shell session that runs the test.

</Blockquote>

```


Maybe the language isn't correct, but something like that. Also, I see that the `ulimit` command by default doesn't persist, does the `sysctl` configuration change persist? If so, it might be good to provide details for reverting it.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -5,18 +5,17 @@ excerpt: 'How to run large-scale k6 tests without distributed-execution'
 
 This document explains how to launch a large-scale k6 test on a single machine without the need for distributed execution.
```
MattDodsonEnglish, line 6: Maybe we should prepare readers for what they are going to read and what they need?

```
This document explains how to launch a large-scale k6 test on a single machine without the need for distributed execution. It most cases, the process involves the following:
- Shell commands to increase the default network and user limits of the operating system
- Monitoring to ensure safe resource usage from the load generator
- Efficient test design, with attention to scripting, k6 options, and file uploads.
- Some error monitoring to detect network and SUT failures.
```

Something like this would serve as a TL;DR, and it would prepare readers for what they need to do (as it's more than just adding a flag,  `--iterations 1000000`)

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -5,18 +5,17 @@ excerpt: 'How to run large-scale k6 tests without distributed-execution'
...
+k6 is different from many other load testing tools in the way it handles hardware resources. A single k6 process will efficiently use all CPU cores on a load generator machine. Depending on the available resources, and taking into consideration some of the advice below, a single instance of k6 is able to run 30,000-40,000 simultaneous users (VUs). In some cases this number of VUs can generate up to 300,000 HTTP requests per second (RPS).
 
 Unless you need more than 100,000-300,000 requests per second (6-12M requests per minute), a single instance of k6 will likely be sufficient for your needs.
 
 Below we will explore what hardware and considerations are needed for generating different levels of load.
 
 ## OS fine-tuning
 
-The following OS changes allow k6 to use the **full network capacity** of the machine for maximum performance.
+The following OS changes for GNU/Linux allow k6 to use the **full network capacity** of the machine for maximum performance.
```
MattDodsonEnglish, line 18:
```suggestion
The following OS changes for GNU/Linux let k6 use the **full network capacity** of the machine for maximum performance.
```

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -58,15 +57,17 @@ Memory consumption heavily depends on your test scenarios. To estimate the memor
...
 
 Simple tests will use ~1-5MB per VU. (1000VUs = 1-5GB).
-Tests that are using file uploads can consume tens of megabytes per VU.
+Tests that are using file uploads, or load large JS modules, can consume tens of megabytes per VU.
+Keep in mind that each VU has a copy of all JS modules your test uses.
+If you need to share memory between VUs, consider using [SharedArray](/javascript-api/k6-data/sharedarray/), or an external data store, such as [Redis](/javascript-api/k6-experimental/redis/).
 
 ## General advice
 
 ### Make your test code resilient
```
MattDodsonEnglish, line 66:
```suggestion
### Resilient error handling
```

I think we should give headings names that are as precise as possible.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -58,15 +57,17 @@ Memory consumption heavily depends on your test scenarios. To estimate the memor
...
 
 When running large stress tests, your script can't assume anything about the HTTP response.
 Often performance tests are written with a "happy path" in mind.
```
MattDodsonEnglish, line 69:
```suggestion
A large test might elicit unexpected behavior from the SUT, so your test script should handle exceptions well.

In testing literature, the correct, predictable response is often called the [_Happy path_](https://en.wikipedia.org/wiki/Happy_path).
A common mistake in performance testing happens when the script assumes that SUT will take the happy path.
```

MattDodsonEnglish: I'd seen this "Happy path" term twice now, and I only just realized that it's an industry term. I think we should define it. Now I'm not the brightest, but even if it's a common term in CS, not all of our readers will come from the same language and educational backgrounds.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -58,15 +57,17 @@ Memory consumption heavily depends on your test scenarios. To estimate the memor
...
-For example, a "happy path" check like the one below is something that we see in k6 often.
+For example, a "happy path" check like the one below is something that we often see in k6 scripts.
```
MattDodsonEnglish, line 70: Now that I understand what "Happy path" is, is this really a happy path check? It seems like it does not assume the happy path, since it accepts the possibility of an incorrect body size. 

Instead, it seems like it's more a case of error handling that's not general enough.

Proposed change:

```
Even if scripts accept the SUT won't always respond according to the happy path, it still might not handle a sufficiently wide range of errors. 
For example, the following check is something we often see in k6 scripts:

File: src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md [hunk 8]
```diff
@@ -78,11 +79,11 @@ const checkRes = check(res, {
...
 ```
 
-Code like this runs fine when the system under test (SUT) is not overloaded and returns proper responses.
+Code like this runs fine when the [system under test](/misc/glossary/#system-under-test) (SUT) is not overloaded and returns proper responses.
 When the system starts to fail, the above check won't work as expected.
 
-The issue here is that the check assumes that there's always a body in a response. The `r.body` may not exist if server is failing.
-In such case, the check itself won't work as expected and error similar to the one below will be returned:
+The issue here is that the check assumes that there's always a body in a response. The `r.body` may not exist if the server is failing.
+In such case, the check itself won't work as expected and an error similar to the one below will be returned:
```
MattDodsonEnglish, line 86: Shouldn't we add an example of how to handle this exception? Right now it seems we diagnose a problem, but don't give any way to find a solution.

MattDodsonEnglish: Oh, I realize now that this is in the text, just not in the diffs. Still, it looks like the example only fails a check if there is a body to begin with.

```javascript
const checkRes = check(res, {
  'Homepage body size is 11026 bytes': (r) => r.body && r.body.length === 11026,
});
```

Wouldn't it make more sense to have two checks? One that checks that a body is returned and other that, when returned, it's the correct length?

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -104,17 +105,19 @@ const checkRes = check(res, {
...
-1. To run k6
-2. To monitor CPU and memory
-3. To monitor the network
+We recommend the following command-line tools for monitoring resources:
 
-For monitoring CPU and memory we recommend [htop](https://en.wikipedia.org/wiki/Htop).
-For monitoring network, we recommend [iftop](https://en.wikipedia.org/wiki/Iftop)
+- CPU and memory: [htop](https://en.wikipedia.org/wiki/Htop) or [nmon](https://nmon.sourceforge.net/).
+
+- Network: [iftop](https://en.wikipedia.org/wiki/Iftop) or [nmon](https://nmon.sourceforge.net/).
```
MattDodsonEnglish, line 117:
```suggestion
- CPU and memory: [`htop`](https://en.wikipedia.org/wiki/Htop) or [`nmon`](https://nmon.sourceforge.net/).

- Network: [`iftop`](https://en.wikipedia.org/wiki/Iftop) or [`nmon`](https://nmon.sourceforge.net/).
```

MattDodsonEnglish: Try to put code and terminal programs in `monospace`, which gives the reader a little clue about the essential form of whatever's being described.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -104,17 +105,19 @@ const checkRes = check(res, {
...
+
+If you prefer graphical applications, you can use your OS's system monitoring tool (e.g. GNOME System
+Monitor or Plasma System Monitor), or use a standalone tool like [SysMonTask](https://github.com/KrispyCamel4u/SysMonTask).
```
MattDodsonEnglish, line 120: Maybe we could put a little summary repetition of what the reader should be looking for here?

Something like:

```
As general guidelines, look for the the following:
- CPU utilization doesn't exceed 80%
- Network traffic isn't constant (?)
- Memory doesn't approach full utiliazation (?)

Comment (MattDodsonEnglish) [reply src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md:120]: I put (?) when I'm not sure.

File: src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md [hunk 11]
```diff
@@ -123,88 +126,66 @@ Here's a screenshot of 3 terminal sessions showing k6, iftop and htop.
...
-cd k6-hardware-benchmark
-yarn install
-
-yarn run to-es5 someplace/yourscript.js
-# builds the ES5 script in someplace/yourscript.es5.js
-
-k6 run --compatibility-mode=base someplace/yourscript.es5.js
-```
-
 ### discardResponseBodies
```
MattDodsonEnglish, line 129: I need to think about whether to put these in `monospace` . I'm leaning towards yes, even though it's unsightly.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -123,88 +126,66 @@ Here's a screenshot of 3 terminal sessions showing k6, iftop and htop.
...
   --vus=20000 \
   --duration=10m \
-  --compatibility-mode=base \
   --no-thresholds \
-  --no-summary \
+  --no-summary
 ```
 
-### Remove unnecessary checks, groups and custom metrics
+### Test script optimizations
```
MattDodsonEnglish, line 160: Shouldn't this be its own H2? It seems incoherent to be nested under "Options"

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -123,88 +126,66 @@ Here's a screenshot of 3 terminal sessions showing k6, iftop and htop.
...
+
+k6 v0.41.0 introduced a change to support metric time-series. A side-effect of this is that every unique URL will create a new time-series object, which may consume more RAM than expected.
+
+The solution to this is to use the [URL grouping](https://k6.io/docs/using-k6/http-requests/#url-grouping) feature.
+
+#### JavaScript optimizations
+
+Finally, if all of the above suggestions are insufficient, there might be some JavaScript optimizations you can do. This includes general improvements to minimize script complexity: avoid deeply nested `for` loops, don't keep references to large objects in memory if it can be avoided, keep external JS dependencies to a minimum, perform tree shaking of the k6 script if you have a build pipeline, etc.
+
+See [this article](https://javascript.info/garbage-collection) about garbage collection in the V8 runtime. While the JavaScript VM k6 uses is very different and runs on Go, the general principles are applicable. Keep in mind that memory leaks are still possible in k6 scripts, and might lead to much quicker RAM exhaustion if not fixed.
```
MattDodsonEnglish, line 186:
```suggestion
Refer to this article about [garbage collection](https://javascript.info/garbage-collection) in the V8 runtime. While the JavaScript VM k6 uses is very different and runs on Go, the general principles are applicable. Keep in mind that memory leaks are still possible in k6 scripts, and might lead to much quicker RAM exhaustion if not fixed.
```

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -217,17 +198,21 @@ The network throughput of the load generator machine, as well as the SUT will li
 
 k6 needs a significant amount of memory when uploading files, as every VU is independent and has its own memory.
 
-### Data transfer costs
+### Costs
+
+#### Data transfer
```
MattDodsonEnglish, line 203: Getting into `h4s` is kind of "prose smell". So much nesting makes it hard for the reader to understand what this relates to.

In any case, we can at least use description lists: https://github.com/grafana/k6-docs/blob/main/CONTRIBUTING/gatsby-reference.md#description-list

Elsewhere, it probably makes sense to flatten the structure. I'll look into this.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -236,191 +221,46 @@ If you run into errors during the execution, it's good to understand if they wer
...
-
-Results
+## Benchmarking k6
 
-- Maximum VUS reached: 30.000
-- Memory used: ~120 GB (out of 370 available)
-- CPU load (avg): ~45 (out of 96.0).
-- Peak RPS: ~61.500.
-- `sleep(1)` in each iteration.
+We have executed a few large tests on different EC2 machines to see how much load k6 can generate. Our general observation is that k6 scales proportionally to the hardware. A machine with 2x more resources can generate roughly 2x more traffic. The limit to this scalability is in the number of open connections. A single Linux machine can open up to 65,535 sockets per IP. This means that a maximum of 65k requests can be sent simultaneously on a single machine. The RPS limit depends on the response time of the SUT. If responses are delivered in 100ms, the RPS limit is 650,000.
```
MattDodsonEnglish, line 260: > A single Linux machine can open up to 65,535 sockets per IP. 

Could we site a source here?

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -123,88 +126,66 @@ Here's a screenshot of 3 terminal sessions showing k6, iftop and htop.
...
-  --no-summary \
+  --no-summary
 ```
 
-### Remove unnecessary checks, groups and custom metrics
+### Test script optimizations
 
-If everything else has failed and you are trying to squeeze more performance out of the hardware,
-you can consider optimizing the code of the load test itself.
+If you are trying to squeeze more performance out of the hardware, you can consider optimizing the code of the test script itself.
```
MattDodsonEnglish, line 162: I think each of these sections should mention how the proposed change would manifest in terms of system performance. If all improve memory, then we could retitle the heading: "Script optimizations to reduce memory"

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -123,88 +126,66 @@ Here's a screenshot of 3 terminal sessions showing k6, iftop and htop.
...
 
 k6 records the result of every individual check and group separately. If you are using many checks and groups, you may consider removing them to boost performance.
 
-**Custom metrics**
+#### Custom metrics
 
 Similar to checks, values for custom metrics (Trend, Counter, Gauge and Rate) are recorded separately. Consider minimizing the usage of custom metrics.
 
-**Thresholds with abortOnFail**
+#### Thresholds with abortOnFail
```
MattDodsonEnglish, line 172: Is this a CPU cost?

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
MattDodsonEnglish: Also, some of these features are quite important. It raises the question, if users have to disable them, is it worth running the test. Why should someone run such a test?

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -123,88 +126,66 @@ Here's a screenshot of 3 terminal sessions showing k6, iftop and htop.
...
-If you need response body for some requests you can set [Params.responseType](/javascript-api/k6-http/params).
+If you need the response body for some requests you can override this with [Params.responseType](/javascript-api/k6-http/params).
+
 
 ### --no-thresholds --no-summary
 
-If you are running a local test and streaming results to the cloud (`k6 run -o cloud`), you may want to disable the terminal summary
-and local threshold calculation because thresholds and summary will be displayed in the cloud.
+If you are running a local test and streaming results to the k6 Cloud (`k6 run -o cloud`), you may want to disable the terminal summary
+and local threshold calculation because thresholds and summary will be displayed in the Cloud.
```
MattDodsonEnglish, line 146:
```suggestion
If you are running a local test and streaming results to the k6 Cloud (`k6 run -o cloud`), you may want to disable the terminal summary
and local threshold calculation,
because k6 Cloud will display the summary and calculate the thresholds.

Without these options, the operations will be duplicated by both the local machine and the k6 Cloud servers.
```

MattDodsonEnglish (commented):
Great updates! I've left a lot of technical questions. 

One thing that seems missing is an early preliminary discussion of _when_ and _why_ you'd want run such a test. In some cases, getting the most out of k6 requires stripping out some critical features, like live threshold evaluation and custom metrics. 

So in what cases would a user even want to do this? Perhaps:
- To component test isolated parts
- To experiment with load generator limits

Maybe this could go at the beginning.

I also would recommend restructuring some sections for a flatter structure. Too much nesting gets hard to read.

I propose this structure, where each third-level indention (what are currently h4s) are items in a description list: https://github.com/grafana/k6-docs/blob/main/CONTRIBUTING/gatsby-reference.md#description-list


```

- [OS fine-tuning](#os-fine-tuning)
- [Hardware considerations](#hardware-considerations)
  - [Network](#network)
  - [CPU](#cpu)
  - [Memory](#memory)
- [Monitoring the load generator](#monitor-the-load-generator-machine)
- [Making error-resilient scripts](#make-your-test-code-resilient)
- [k6 script optimizations](#general-advice)
  - [discardResponseBodies](#discardresponsebodies)
  - [--no-thresholds --no-summary](#--no-thresholds---no-summary)
  - Remove resource-intensive operations
  - [Checks and groups](#checks-and-groups)
    - [Custom metrics](#custom-metrics)
    - [Thresholds with abortOnFail](#thresholds-with-abortonfail)
    - [URL grouping](#url-grouping)
  - [JavaScript optimizations](#javascript-optimizations)
- [File upload considerations](#file-upload-testing)
  - [Network throughput](#network-throughput)
  - [Memory](#memory-1)
  - [Costs](#costs)
    - [Data transfer](#data-transfer)
    - [Virtual server](#virtual-server)
- [Errors](#errors)
  - [read: connection reset by peer](#read-connection-reset-by-peer)
  - [context deadline exceeded](#context-deadline-exceeded)
  - [dial tcp 52.18.24.222:80: i/o timeout](#dial-tcp-52182422280-io-timeout)
  - [socket: too many open files](#socket-too-many-open-files)
- [Benchmarking k6](#benchmarking-k6)
- [Distributed execution](#distributed-execution)
- [Large-scale tests in k6 Cloud](#large-scale-tests-in-k6-cloud)
- [Read more](#read-more)
```

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
imiric: > As a casual, I feel uncomfortable running these commands.

Well, this article is directed more towards advanced k6 users, who are reaching the limits imposed by their OS. I agree that it should be as friendly as the rest of our documentation, but at some point we have to expect some familiarity from the reader.

We do explain the purpose of these commands below, and that they require root privileges (only the `sysctl` commands do, but we gloss over some of these subtleties that are better explained in the "Fine tuning OS" article). These are not "dangerous" by any means, but tweak the system to allow k6 (and any other program) to achieve maximum performance. These settings differ from OS and Linux distribution, but are usually set rather low by default, since the general public doesn't typically run programs that use tens of GBs of RAM, and opens thousands of concurrent TCP connections.

Also note that the settings changed by these commands are temporary, and will be reset to the system default after a reboot. If the user wants to make them permanent (presumably after they've confirmed the improvement), they can change some system files, which also require root privileges. The `ulimit` command applies only for the current shell session, so it has even less impact. This is all better explained in the "Fine tuning OS" article.

So I'm not sure what kind of warning would be helpful here, given the audience this article is aimed towards.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
imiric: Sure, see 8d725942. I changed some of the wording slightly.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
imiric: It's a pretty common term in testing, even outside of load/performance testing.

Instead of defining it here, let's add it to the [glossary](https://k6.io/docs/misc/glossary/)? Then we can link it here, just as we did with SUT.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
imiric: It is a happy path check, since it doesn't take into account failure conditions under load, when the response body wouldn't exist. The body size check is a standard assertion, and the user might want the test to fail if it's incorrect.

Your suggestion sounds more confusing, sorry, so I'd rather leave this as is.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
MattDodsonEnglish: > Well, this article is directed more towards advanced k6 users, 

Two things here: the article doesn't state directly who it's for (but the new intro does a much better job of setting expectations :+1:); how much does "advanced k6" use imply Linux knowledge?

I don't see what information with these extra two sentences.

```
These default limits are in place because ordinary programs don't need to open thousands of concurrent TCP connections or use dozens of GBs of RAM.
The `ulimit` changes apply only for the current shell, and the other settings reset when the system reboots.
For details, refer to [Fine-tuning OS](/misc/fine-tuning-os). 
```

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
MattDodsonEnglish: > It is a happy path check, since it doesn't take into account failure conditions under load, when the response body wouldn't exist. The body size check is a standard assertion, and the user might want the test to fail if it's incorrect.

I don't understand. Isn't an incorrect body size another indication of system failure? Or are we defining failure as "no body response at all?". From the Wikipedia page, I thought a happy path is where there are no error conditions.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
MattDodsonEnglish: > saw some examples of how description lists are rendered, and I'm not sure if they would be a good fit here. > The titles are too large, and we don't need the indentation of the bodies.

I agree, I wont to reduce the size of description list items, but that's a CSS change. Anyway, they're pretty much the size of the h4s so no difference there.

The indentation is fine, IMO: it creates a clear sequence of repetition and alignment. With h4 there's only one level of alignment, the definition makes a clearer sequence of Item and description.

It would look like this: https://k6.io/blog/reverse-load-test/#analysis-bottlenecks-and-planned-improvements

Really I think it's fine.

Anyway the docs table of contents shows only h2s, so the best way to improve that navigation is by flattening the structure out, reducing h3s and h4s.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
MattDodsonEnglish: I agree with RFC for authority.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
imiric: Matt, I think you're nitpicking at this point. It's fine as is :)

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
imiric: I don't think there's an objective best practice here, regardless of what some authority says. :smile:

Consistency matters more, so if you strongly prefer it this way, feel free to change it, and I'll make sure to use it from now on.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
imiric: > the article doesn't state directly who it's for

None of our articles mention the intended audience AFAIK, but whoever is interested in "running large tests" and reaching the limits of their OS, would likely expect to read about advanced topics.

> how much does "advanced k6" use imply Linux knowledge?

I initially wrote "advanced user" in general, and still think it applies. I think there's some overlap between an advanced k6 user and an advanced computer user, so we don't need to explain every topic from scratch, especially since we already mention these details on the "Fine tuning OS" page.

That said, I expanded the intro a bit in 7e64d232. Let me know if this is an improvement.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
imiric: Let me know if 63f1f8e9 resolves this.

MattDodsonEnglish:
Nice, I made some edits for structure and sentences to merge into this branch. #1003  @imiric , I'd appreciate a review. I don't think you need to analyze it too deeply―just if anything feels off to you. I'll proofread the text after the benchmarks are added.

 It's a hard article to organize because of it's length and diversity, but I'm happy with how it shaped up. After my merges, I'll be totally fine with the article (except my necessary proofreading).

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -104,107 +111,94 @@ const checkRes = check(res, {
...
+If you prefer graphical applications, you can use your OS's system monitoring tool (e.g. GNOME System
+Monitor or Plasma System Monitor), or use a standalone tool like [SysMonTask](https://github.com/KrispyCamel4u/SysMonTask).
 
 Here's a screenshot of 3 terminal sessions showing k6, iftop and htop.
 ![k6 iftop htop](./images/large-scale-testing-3-terminals.png)
 
+As general guidelines, look for the following:
+- CPU utilization doesn't exceed 90%. If all CPU cores are 100% utilized during the test run, you might notice performance degradation in your test results.
+- Network utilization is at an acceptable level. Depending on your test, you might expect to fully saturate the network bandwidth, or this might be a signal that your test is bound by the available bandwidth. In other scenarios, you might want to minimize network usage to keep costs down.
+- Memory utilization doesn't exceed 90%. If you're close to exhausting available physical RAM, the system might start swapping memory to disk, which will affect performance and system stability. In extreme cases, running out of memory on Linux will cause the system to end the k6 process.
```
imiric, line 134: Should we recommend running without swap? Our loadgens don't use it, AFAIK. cc: @mstoykov

mstoykov: I would recommend people to run without swap *in general* ;) The only use at this point IMO are : 
1. systems which don't have enough ram at all
2. hibernation (which is why I still have swap on my laptop)

In (my) practice the difference between rogue process running out of memory and dying and going into swap is just that in the second case it dies a lot slower and later usually thrashing the whole system in the process ;). 

I guess with NVMe SSDs this is a bit better but even on my laptop whenever I go to swap I find out as the whole system goes really slow.

Given that k6 is a performance tool - it performance degrading significantly because swap is used seems like a terrible idea even if the process itself doesn't die.

na--:
Removed myself from the list of reviewers since this doesn't need my review, I likely can't contribute anything that @mstoykov and @imiric haven't already .

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
imiric: Fully agree, so I think this deserves a separate section that explains it. Will add, thanks.

imiric: Resolved by 1ed89f2c.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -27,46 +36,79 @@ ulimit -n 250000
...
+: k6 likes memory, but isn't [as greedy as other load testing tools](https://k6.io/blog/comparing-best-open-source-load-testing-tools#memory-usage).
 Memory consumption heavily depends on your test scenarios. To estimate the memory requirement of your test,
 run the test on your development machine with 100VUs and multiply the consumed memory by the target number of VUs.
+: Simple tests use ~1-5MB per VU. (1000VUs = 1-5GB).
+Tests that use file uploads, or load large JS modules, can consume tens of megabytes per VU.
+Note that each VU has a copy of all JS modules your test uses.
+To share memory between VUs, consider using [SharedArray](/javascript-api/k6-data/sharedarray/), or an external data store, such as [Redis](/javascript-api/k6-experimental/redis/).
+: If you're using [swap space](https://en.wikipedia.org/wiki/Memory_paging), consider disabling it. If the system runs out of physical memory, the process of swapping memory to much slower secondary storage will have erratic effects on performance and system stability. Instead, plan ahead for the memory usage you expect your tests to achieve, and ensure that you have enough physical RAM for the usage to not exceed 90%.
```
mstoykov, line 71:
```suggestion
: If you're using [swap space](https://en.wikipedia.org/wiki/Memory_paging), consider disabling it. If the system runs out of physical memory, the process of swapping memory to much slower secondary storage will have erratic effects on performance and system stability. Which likely will invalidate any test results as the load generator had different performance in different parts of the test. Instead, plan ahead for the memory usage you expect your tests to achieve, and ensure that you have enough physical RAM for the usage to not exceed 90%.
```

It is probably good idea to spell out what the consequence for the testing is

ppcano:
@imiric any blocker - can we merge it?

imiric:
@ppcano Not yet. It's waiting for [PRs in k6-hardware-benchmark](https://github.com/grafana/k6-hardware-benchmark/pulls) to be approved and merged.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -3,20 +3,29 @@ title: 'Running large tests'
 excerpt: 'How to run large-scale k6 tests without distributed-execution'
 ---
 
-This document explains how to launch a large-scale k6 test on a single machine without the need for distributed execution.
+A common misconception about load testing is that it requires [distributed execution](#distributed-execution) (that is, load generated from multiple machines).
+This isn't the case with k6.
```
mstoykov, line 7: This is a very strong statement that arguably is wrong. 

Can we have not so marketing message of 
```suggestion
A common misconception about load testing is that it requires [distributed execution](#distributed-execution) (that is, load generated from multiple machines).
This isn't necessarily the case with k6.
```

mstoykov (approved):
LGTM! 

This reads as marketing material with a bunch of hyperbole and statements bordering on "k6 is the only load testing tool that users your resources fully". Which arguably is not true. Apart from `locust` I think all the other fairly known load testing tools use the resources fully - some of them just have a lot of other things that eat into their actual performance (jMeter for example). 

And a lot of those statements would've been a lot better if we were *comparing* with other tools, and we had "see graph k6 does 10kRPs, jMeter(for example) does 2k RPS, both at 80% CPU". But we show only k6 and then are "it's better than most" which IMO is pointless - better than most is still worse than some which might be what people use. And with no concrete comparison there is no way for someone to know if with this same setup their load testing tool will not give them twice the performance.


I personally do not think the docs are a place for marketing, so I would prefer if we tone it down a bunch on the amount of comparison to others where we ... just say stuff instead of show them.

I still think there is a lot of valuable information and showing that you can do X RPS for some situation is still beneficial to users, and we should have it. But a bunch of the article reads as "k6 exceptionalism" to me.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
ppcano: I support changing the wording here and in other places. 

Haven't reviewed it. @mstoykov, it can help if you can point other sentences.

imiric:
@mstoykov You're right that the article has a marketing tone biased towards k6. I wouldn't say it reads like k6 is the _only_ tool that fully utilizes system resources, but we're pointing out a clear advantage over other tools that struggle with that. This is not a comparison article like the ["Open source load testing tool review"](https://k6.io/blog/comparing-best-open-source-load-testing-tools/), so mentioning the performance of other tools would be out of context. This article is about how to configure the system and k6 to fully utilize the available hardware resources.

> I personally do not think the docs are a place for marketing

Agreed, so you'll be happy to know that this article will be moved to the blog. :smile:


> a bunch of the article reads as "k6 exceptionalism" to me

Can you point out all those places?

This shouldn't read as a marketing article, but then again, it should point out k6's strengths, so that's a thin line to tread. I can try toning down some of these parts, with @MattDodsonEnglish's help, of course.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
imiric: Hhmm what would you say is wrong about it? That it's not a common misconception, or that k6 does require distributed execution?

I'm not sure how common of a misconception it is, but k6 certainly doesn't _require_ it. I'd say there's some truth in both statements.

I can add necessarily to soften this, but maybe we should rephrase this altogether? WDYT @MattDodsonEnglish?

MattDodsonEnglish: How about this?

> k6 can generate a lot of load from a small machine.
> With proper monitoring and script optimization, you might be able to run a rather large load test without needing distributed execution.

MattDodsonEnglish:
> , with @MattDodsonEnglish's help, of course. Ah, sorry, Matt has limited availability right now, so we can change it ourselves.

Thanks, @imiric ! :pray:  I can make a quick comment though. Besides that first line, I also don't really see the "k6 exceptionalism" that @mstoykov pointed out. The only other overt line that I found was "not as greedy as other tools" but that does link to a data-driven comparison.

I believe that for technical articles like this, the best marketing is honesty and usefulness. So if there are other places where the doc seems over the top, I'd prefer to delete those lines. I just don't know what lines those are exactly.

However, the first line makes such a strong impression that might it frame the whole article as marketing material, when really the material is quite practical and technical. That's also from the original version, and since then there's been much revision. So it isn't very harmonious with the rest of the article. I proposed cooling it down here:

https://github.com/grafana/k6-docs/pull/992#discussion_r1102636685.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
imiric: Thanks, Matt!

I changed this, and a few other places that were too harsh.

@mstoykov Let us know if 62bd7e8d is an improvement.

imiric:
I know that you already approved this @mstoykov, but it sounded like you had major objections to the marketing tone of the article, so please take another look if the recent changes are an improvement.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
mstoykov: For me that changes is great. 

IMO it was wrong on a lot of levels but especially on the hardness of "it isn't the case".

But distributed mode isn't just used for *scale* it also used to make the test come from multiple (usually) geo distributed machines. 

And we are not doing some very complex test - for something more complex we will need even more resources.

Sometimes there are other considerations (as the article already goes through) such as the number of ports and the need for multiple IPs. 

Arguably those are fixable *without* distributed execution, but probably only to an extend.

So the whole hardness of the "it isn't the case" is the *biggest* problem.

mstoykov (approved):
I guess I over-reacted as the start was particularly marketing like IMO.

Although that in general seems to be mostly from the previous version then anything done to this one. 

With the latest changes IMO there are no unfound comparisons.

But maybe we should also try to get https://github.com/grafana/loadgentest updated? Maybe with just less tools?

imiric:
@mstoykov Yeah, we should definitely update the comparison article and benchmark results at some point as well. This year would be great.

Can we get Ragnar to do it again? :smile:

Issue #932, opened 2022-12-05
Add relevant JS- and testing-related terms to Glossary


MattDodsonEnglish:
This came up in @immavalls' Week of Load Testing:

- [ ] ES6
- [ ] BDD
- [ ] load generator (can't believe we don't have this!)
- [ ] ramp-up
- [ ] ramp-down
- [ ] workload
- [ ] transactions (?)

PR #997, opened 2023-01-24
Gloss; add definition for happy path

MattDodsonEnglish:
part of #932

### DIFF
--- a/src/data/markdown/translated-guides/en/07 Misc/09 Glossary.md
+++ b/src/data/markdown/translated-guides/en/07 Misc/09 Glossary.md
@@ -79,7 +79,7 @@ Graceful stop
 : A period that lets VUs finish an iteration at the end of a load test. Graceful stops prevent abrupt halts in execution.<br/><br/>[Graceful stop reference](/using-k6/scenarios/graceful-stop/)
 
 Happy path
-: The default system behavior that happens when a known input produces an expected output. A common mistake in performance testing happens when scripts account only for the happy path. Most load tests try to discover system errors, so test scripts should include exception handling.<br/><br/>[Happy path (Wikipedia)](https://en.wikipedia.org/wiki/Happy_path)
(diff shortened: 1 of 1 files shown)


src/data/markdown/translated-guides/en/07 Misc/09 Glossary.md
```diff
@@ -75,7 +76,10 @@ Goja
 : A JavaScript engine written in Go. k6 binaries are embedded with Goja, enabling test scripting in JavaScript.<br/><br/>[Goja repository](https://github.com/dop251/goja)
 
 Graceful stop
-: A period that lets VUs finish an iteration at the end of a load test. Graceful stops prevent abrupt halts in execution.<br/><br/>[Graceful stop reference](/using-k6/scenarios/concepts/graceful-stop/)
+: A period that lets VUs finish an iteration at the end of a load test. Graceful stops prevent abrupt halts in execution.<br/><br/>[Graceful stop reference](/using-k6/scenarios/graceful-stop/)
+
+Happy path
+: The default system behavior that happens when a known input produces an expected output. A common mistake in performance testing happens when scripts account only for the happy path. Most load tests try to discover system errors, so test scripts should include exception handling.<br/><br/>[Happy path (Wikipedia)](https://en.wikipedia.org/wiki/Happy_path)
```
imiric, line 82:
```suggestion
: The default system behavior that happens when a known input produces an expected output. A common mistake in performance testing happens when scripts account only for the successful scenario. Most load tests try to discover system errors, so test scripts should include exception handling.<br/><br/>[Happy path (Wikipedia)](https://en.wikipedia.org/wiki/Happy_path)
```

To avoid using the term name in the definition.

imiric (approved):
:+1:

src/data/markdown/translated-guides/en/07 Misc/09 Glossary.md
MattDodsonEnglish: I've put it in parenthesis, because  I want the text to drive home how the happy path relates specifically to testing. I want to be explicit that "best case" is synonymous with "happy path". And I avoided "scenario" only because it could mean something else in this gloss :-)

Issue #1002, opened 2023-01-25
Let description lists have codeblocks


MattDodsonEnglish:
Unless I misunderstand, the definitions in a `DescriptionList` element can't have codeblocks. It's not a common need, but it could be useful in long explanatory articles.

imiric:
Hey, as mentioned in https://github.com/grafana/k6-docs/pull/1003#discussion_r1089087450, I think we should close this issue, since we shouldn't need to implement code blocks in description list definitions. `<dl>` elements should be used for short term descriptions, so we shouldn't need to support advanced formatting in `<dd>` elements.

MattDodsonEnglish:
Fair enough. I still think it's reasonable to expect all components to parse recursively, but you've convinced me there's truly no use for this particular case :-)

Issue #1001, opened 2023-01-25
Resize description list terms


MattDodsonEnglish:
The `<dd>` element is too big and prominent. The items are arguably more prominent than `h3` headings, which is too much. To change this, I assume we need just change one place in the CSS.

Issue #998, opened 2023-01-25
Add parameter to show h3s in Tables of contents


MattDodsonEnglish:
The docs site shows only h2s for the in-page ToC. In most cases, this is fine, but for long articles, it could improve scannability to extend the table of contents to h3s. This is especially true in docs like the testing guides, where the content  is long and has many facets.

ppcano:
We'll be moving testing guides to the blog. Then, on which other pages will the H3 ToC be necessary?

MattDodsonEnglish:
That's true. The only time I really miss it is in these long articles anyway.

PR #1003, opened 2023-01-25
large tests, edit structure, format, and sentences

MattDodsonEnglish:
- Grouping related elements as closely as possible
- Trimming sentences down and removing unneeded verbiage
- Organizing items in a flatter hierarchy 
- Recommends 80% CPU utilization in the monitoring section, since that's recommended in the hardware section.

This PR uses description lists. In fact, this document is essentially a series of lists that describe different considerations for different facets of a load test.

The goal of the lists is to use standardized formatting, with multiple levels of alignment, to create a visual clue that the elements all relate to a single topic.

Ideally, all the scripting stuff would be in one h2, the dd elements would be smaller, and the error messages and options sections would be in definition lists too. But, I decided against it because the ToC goes to only level two, and because our description list component can't handle code blocks. I've made issues about this problems, #998 , #1001 , #1002 .

But I think some aesthetic issues are worth the cleaner organization. And these issues should be fixable.

### DIFF
--- a/src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
+++ b/src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
@@ -3,19 +3,23 @@ title: 'Running large tests'
 excerpt: 'How to run large-scale k6 tests without distributed-execution'
 ---
 
-This document explains how to launch a large-scale k6 test on a single machine without the need for distributed execution. In most cases, the process involves the following:
+A common misconception says that a large load test requires [distributed execution](#distributed-execution) (that is, with load generated from multiple machines).
+This isn't the case with k6.
+
+This document explains how to launch a large-scale k6 test on a single machine.
+Maximizing the load a machine generates is a multi-faceted process, which includes:
+
 - Changing operating system settings to increase the default network and user limits.
-- Monitoring the load generator machine to ensure adequate resource usage.
-- Efficient test design, with attention to scripting, k6 options, and file uploads.
+- Monitoring the load-generator machine to ensure adequate resource usage.
+- Designing efficient tests, with attention to scripting, k6 options, and file uploads.
 - Monitoring the test run to detect errors logged by k6, which could indicate limitations of the load generator machine or the [system under test](/misc/glossary/#system-under-test) (SUT).
 
-A common misconception of many load testers is that [distributed execution](#distributed-execution) (the ability to launch a load test from multiple machines) is required to generate a large load. This is not the case with k6.
+k6 differs from many other load testing tools in how it handles hardware resources.
+A single k6 process efficiently uses all CPU cores on a load generator machine. Depending on the available resources, and with the guidelines described in this document, a single instance of k6 can run 30,000-40,000 simultaneous users (VUs).
+In some cases, this number of VUs can generate up to 300,000 HTTP [requests per second](/misc/glossary/#requests-per-second) (RPS).
 
-k6 is different from many other load testing tools in the way it handles hardware resources. A single k6 process will efficiently use all CPU cores on a load generator machine. Depending on the available resources, and taking into consideration some of the advice below, a single instance of k6 is able to run 30,000-40,000 simultaneous users (VUs). In some cases this number of VUs can generate up to 300,000 HTTP [requests per second](/misc/glossary/#requests-per-second) (RPS).
-
-Unless you need more than 100,000-300,000 requests per second (6-12M requests per minute), a single instance of k6 will likely be sufficient for your needs.
-
-Below we will explore what hardware and considerations are needed for generating different levels of load.
+Unless you need more than 100,000-300,000 requests per second (6-12M requests per minute), a single instance of k6 is likely sufficient for your needs.
+Read on to learn about how to get the most load from a single machine.
 
 ## OS fine-tuning
 
@@ -38,42 +42,72 @@ For detailed information about these settings, how to make them permanent, and i
 
 ## Hardware considerations
 
-### Network
+As load increases, you need to pay attention to hardware constraints.
 
-Network throughput of the machine is an important consideration when running large tests.
-Many AWS EC2 machines come with 1Gbit/s connection which may limit the amount of load k6 can generate.
+<DescriptionList>
 
-When running the test, you can use `iftop` in the terminal to view in real-time the amount of network traffic generated.
+Network
+: Network throughput of the machine is an important consideration when running large tests.
+Many AWS EC2 machines come connections of with 1Gbit/s, which may limit the load k6 can generate.
+: When running the test, use a tool like `iftop` in the terminal to view the amount of network traffic generated in real time.
 If the traffic is constant at 1Gbit/s, your test is probably limited by the network card. Consider upgrading to a different EC2 instance.
 
-### CPU
-
-Unlike many other load testing tools, k6 is heavily multi-threaded. It will effectively use all available CPU cores.
-
-The amount of CPU you need depends on your test files (sometimes called test script).
+CPU
+: Unlike many other load testing tools, k6 is heavily multi-threaded. It will effectively use all available CPU cores.
+: The amount of CPU you need depends on your test script and associated files.
 Regardless of the test file, you can assume that large tests require a significant amount of CPU power.
 We recommend that you size the machine to have at least 20% idle cycles (up to 80% used by k6, 20% idle).
 If k6 uses 100% of the CPU to generate load, the test will experience throttling limits, and this may
 cause the result metrics to have a much larger response time than in reality.
 
-### Memory
-
-k6 likes memory, but [it isn't as greedy as other load testing tools](https://k6.io/blog/comparing-best-open-source-load-testing-tools#memory-usage).
+Memory
+: k6 likes memory, but isn't [as greedy as other load testing tools](https://k6.io/blog/comparing-best-open-source-load-testing-tools#memory-usage).
 Memory consumption heavily depends on your test scenarios. To estimate the memory requirement of your test,
 run the test on your development machine with 100VUs and multiply the consumed memory by the target number of VUs.
+: Simple tests use ~1-5MB per VU. (1000VUs = 1-5GB).
+Tests that use file uploads, or load large JS modules, can consume tens of megabytes per VU.
+Note that each VU has a copy of all JS modules your test uses.
+To share memory between VUs, consider using [SharedArray](/javascript-api/k6-data/sharedarray/), or an external data store, such as [Redis](/javascript-api/k6-experimental/redis/).
 
-Simple tests will use ~1-5MB per VU. (1000VUs = 1-5GB).
-Tests that are using file uploads, or load large JS modules, can consume tens of megabytes per VU.
-Keep in mind that each VU has a copy of all JS modules your test uses.
-If you need to share memory between VUs, consider using [SharedArray](/javascript-api/k6-data/sharedarray/), or an external data store, such as [Redis](/javascript-api/k6-experimental/redis/).
+</DescriptionList>
 
-## General advice
+## Monitoring the load generator
 
-### Resilient error handling
+The previous section described how hardware limits the load your machine can generate.
+Along with understanding conceptually, you also should monitor resources while the test runs&mdash;especially when running tests for the first time.
+
+
+The easiest way to monitor the load generator is to open multiple terminals on the machine: one to run k6, and others to monitor the CPU, memory, and network .
+We recommend the following tools:
+
+- CPU and memory: [htop](https://en.wikipedia.org/wiki/Htop) or [nmon](https://nmon.sourceforge.net/).
+- Network: [iftop](https://en.wikipedia.org/wiki/Iftop) or [nmon](https://nmon.sourceforge.net/).
+
+If you prefer graphical applications, use the system monitoring tool from your OS (e.g. GNOME System
+Monitor or Plasma System Monitor), or standalone tools like [SysMonTask](https://github.com/KrispyCamel4u/SysMonTask).
+
+Here's a screenshot of 3 terminal sessions showing k6, iftop and htop.
+![k6 iftop htop](./images/large-scale-testing-3-terminals.png)
+
+### Monitoring guidelines
+
+As the test runs, these are the indicators to monitor:
+
+- **CPU utilization stays within 80%.**
+  If all CPU cores are 100% utilized during the test run, you might notice performance degradation in your test results.
+
+- **Network utilization is at an acceptable level.**
+Depending on your test, you might expect to fully saturate the network bandwidth, or this might be a signal that your test is bound by the available bandwidth. In other scenarios, you might want to minimize network usage to keep costs down.
+
+- **Memory utilization doesn't exceed 90%.**
+If you're close to exhausting available physical RAM, the system might start swapping memory to disk, which will affect performance and system stability. In extreme cases, running out of memory on Linux will cause the system to end the k6 process.
+
+## Error handling should be resilient
 
 When running large stress tests, your script can't assume anything about the HTTP response.
-Often performance tests are written with a "happy path" in mind.
-For example, a "happy path" check like the one below is something that we often see in k6 scripts.
+It's a mistake to script a performance test with only the [happy path](/misc/glossary#happy-path "The default behavior that happens when the system returns no errors") in mind.
+
+For example, in k6 scripts, we often see something like this _happy path_ check:
 
 ```javascript
 import { check } from 'k6';
@@ -85,17 +119,19 @@ const checkRes = check(res, {
 });
 ```
 
-Code like this runs fine when the [SUT](/misc/glossary/#system-under-test) is not overloaded and returns proper responses.
-When the system starts to fail, the above check won't work as expected.
+Code like this runs fine when the SUT is not overloaded and returns proper responses.
+When the system starts to fail, this check won't work as expected.
 
-The issue here is that the check assumes that there's always a body in a response. The `r.body` may not exist if the server is failing.
-In such case, the check itself won't work as expected and an error similar to the one below will be returned:
+The issue is that the check assumes that the response always has a body.
+But if the server is failing, `r.body` might not exist.
+In such cases, the check itself won't work as expected, and an error such as the following will be returned:
 
 ```bash
 ERRO[0625] TypeError: Cannot read property 'length' of undefined
 ```
 
-To fix this issue your checks must be resilient to any response type. This change will fix the above problem.
+To fix this issue, make your checks resilient to any response type.
+This change will handle the exception.
 
 <CodeGroup labels={["resilient check"]}>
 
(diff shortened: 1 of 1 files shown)


src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -311,22 +338,21 @@ k6 run --execution-segment "3/4:1"     --execution-segment-sequence "0,1/4,2/4,3
...
+- Each k6 instance evaluates [Thresholds](/using-k6/thresholds) independently - excluding the results of the other k6 instances. If you want to disable the threshold execution, use [`--no-thresholds`](/using-k6/k6-options/reference#no-thresholds).
 - k6 reports the metrics individually for each instance. Depending on how you store the load test results, you'll have to aggregate some metrics to calculate them correctly.
 
-With the limitations mentioned above, we built a [Kubernetes operator](https://github.com/grafana/k6-operator) to distribute the load of a k6 test across a **Kubernetes cluster**. For further instructions, check out [the running distributed k6 tests on Kubernetes tutorial](https://k6.io/blog/running-distributed-tests-on-k8s/).
+With the limitations mentioned above, we built a [`Kubernetes operator](https://github.com/grafana/k6-operator) to distribute the load of a k6 test across a **Kubernetes cluster**. For further instructions, check out [the running distributed k6 tests on Kubernetes tutorial](https://k6.io/blog/running-distributed-tests-on-k8s/).
```
MattDodsonEnglish, line 347:
```suggestion
With the limitations mentioned above, we built a [Kubernetes operator](https://github.com/grafana/k6-operator) to distribute the load of a k6 test across a **Kubernetes cluster**. For further instructions, check out [the running distributed k6 tests on Kubernetes tutorial](https://k6.io/blog/running-distributed-tests-on-k8s/).
```

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -38,42 +42,72 @@ For detailed information about these settings, how to make them permanent, and i
...
+Tests that use file uploads, or load large JS modules, can consume tens of megabytes per VU.
+Note that each VU has a copy of all JS modules your test uses.
+To share memory between VUs, consider using [SharedArray](/javascript-api/k6-data/sharedarray/), or an external data store, such as [Redis](/javascript-api/k6-experimental/redis/).
+
+</DescriptionList>
+
+## Monitoring the load generator
+
+The previous section described how hardware limits the load your machine can generate.
+Along with understanding conceptually, you also should monitor resources while the test runs&mdash;especially when running tests for the first time.
```
imiric, line 77:
```suggestion
Keeping this in mind, you should monitor resources while the test runs&mdash;especially when running tests for the first time.
```

"Along with understanding [this] conceptually" sounds too... wordy.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -38,42 +42,72 @@ For detailed information about these settings, how to make them permanent, and i
...
+
+
+The easiest way to monitor the load generator is to open multiple terminals on the machine: one to run k6, and others to monitor the CPU, memory, and network .
```
imiric, line 80:
```suggestion
The easiest way to monitor the load generator is to open multiple terminals on the machine: one to run k6, and others to monitor the CPU, memory, and network.
```

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -38,42 +42,72 @@ For detailed information about these settings, how to make them permanent, and i
...
+Monitor or Plasma System Monitor), or standalone tools like [SysMonTask](https://github.com/KrispyCamel4u/SysMonTask).
+
+Here's a screenshot of 3 terminal sessions showing k6, iftop and htop.
+![k6 iftop htop](./images/large-scale-testing-3-terminals.png)
 
-## General advice
+### Monitoring guidelines
 
-### Resilient error handling
+As the test runs, these are the indicators to monitor:
```
imiric, line 94:
```suggestion
As the test runs, these are good indicators to monitor:
```

These are not the only ones.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -38,42 +42,72 @@ For detailed information about these settings, how to make them permanent, and i
...
+
+- **Memory utilization doesn't exceed 90%.**
+If you're close to exhausting available physical RAM, the system might start swapping memory to disk, which will affect performance and system stability. In extreme cases, running out of memory on Linux will cause the system to end the k6 process.
+
+## Error handling should be resilient
 
 When running large stress tests, your script can't assume anything about the HTTP response.
-Often performance tests are written with a "happy path" in mind.
-For example, a "happy path" check like the one below is something that we often see in k6 scripts.
+It's a mistake to script a performance test with only the [happy path](/misc/glossary#happy-path "The default behavior that happens when the system returns no errors") in mind.
```
imiric, line 108: I wouldn't phrase it as an outright "mistake". It's certainly an oversight, and it's good practice to think about failure scenarios.

Can we rephrase this in a similar way as it was before?

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -38,42 +42,72 @@ For detailed information about these settings, how to make them permanent, and i
...
+
+- **Network utilization is at an acceptable level.**
+Depending on your test, you might expect to fully saturate the network bandwidth, or this might be a signal that your test is bound by the available bandwidth. In other scenarios, you might want to minimize network usage to keep costs down.
+
+- **Memory utilization doesn't exceed 90%.**
+If you're close to exhausting available physical RAM, the system might start swapping memory to disk, which will affect performance and system stability. In extreme cases, running out of memory on Linux will cause the system to end the k6 process.
+
+## Error handling should be resilient
 
 When running large stress tests, your script can't assume anything about the HTTP response.
```
imiric, line 107:
```suggestion
When running large stress tests, your script shouldn't assume anything about the HTTP response.
```

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -85,17 +119,19 @@ const checkRes = check(res, {
...
+But if the server is failing, `r.body` might not exist.
+In such cases, the check itself won't work as expected, and an error such as the following will be returned:
 
 ```bash
 ERRO[0625] TypeError: Cannot read property 'length' of undefined
 ```
 
-To fix this issue your checks must be resilient to any response type. This change will fix the above problem.
+To fix this issue, make your checks resilient to any response type.
+This change will handle the exception.
```
imiric, line 134:
```suggestion
To fix this issue, make your checks resilient to any response type.

The following change will handle the exception.
```

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -170,67 +186,72 @@ k6 run scripts/website.js \
...
-### Thresholds with abortOnFail
+Checks and groups
+: k6 records the result of every individual check and group separately. If you are using many checks and groups, you may consider removing them to boost performance.
 
-If you have configured [abortOnFail thresholds](/using-k6/thresholds#aborting-a-test-when-a-threshold-is-crossed), k6 needs to evaluate the result constantly to verify that the threshold wasn't crossed. Consider removing this setting.
+Custom metrics
+: Similar to checks, values for custom metrics (Trend, Counter, Gauge and Rate) are recorded separately. Consider minimizing the usage of custom metrics.
 
-### URL grouping
+Thresholds with abortOnFail
```
imiric, line 208: Should `abortOnFail` be formatted as code here, same as `discardResponseBodies`? Or is this not possible currently because of #1002?

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -170,67 +186,72 @@ k6 run scripts/website.js \
...
+Finally, if the preceding suggestions are insufficient, you might be able to do some general JavaScript optimizations:
+- Avoid deeply nested `for` loops.
+- Avoid references to large objects in memory as much as possible.
+- Keep external JS dependencies to a minimum.
+- Perform tree shaking of the k6 script if you have a build pipeline, and so on.
 
-Refer to this article about [garbage collection](https://javascript.info/garbage-collection) in the V8 runtime. While the JavaScript VM k6 uses is very different and runs on Go, the general principles are applicable. Keep in mind that memory leaks are still possible in k6 scripts, and might lead to much quicker RAM exhaustion if not fixed.
+Refer to this article about [garbage collection](https://javascript.info/garbage-collection) in the V8 runtime. While the JavaScript VM that k6 uses is very different and runs on Go, the general principles apply. Note that memory leaks are still possible in k6 scripts, and, if not fixed, they might exhaust RAM much quicker.
```
imiric, line 225:
```suggestion
Refer to this article about [garbage collection](https://javascript.info/garbage-collection) in the V8 runtime. While the JavaScript VM that k6 uses is very different and runs on Go, the general principles apply. Note that memory leaks are still possible in k6 scripts, and, if not fixed, they might exhaust RAM much more quickly.
```

Sounds more natural to me.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -170,67 +186,72 @@ k6 run scripts/website.js \
...
-Special considerations must be taken when testing file uploads.
+<DescriptionList>
 
-### Network throughput
+Network throughput
+: The network throughputs of the load generator machine and of the SUT are likely bottlenecks for file uploads.
 
-The network throughput of the load generator machine, as well as the SUT will likely be the bottleneck.
+Memory
+: k6 needs a significant amount of memory when uploading files, as every VU is independent and has its own memory.
```
imiric, line 237: Might be good to expand on this.

```suggestion
: k6 needs a significant amount of memory when uploading files. Every VU is independent and has its own memory, and files will be copied in all VUs that upload them. You can follow [issue #1931](https://github.com/grafana/k6/issues/1931) that aims to improve this situation.
```

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -170,67 +186,72 @@ k6 run scripts/website.js \
...
+: If your infrastructure is already hosted on AWS, consider running your load generator machine within the same AWS region and availability zone. In some cases, this traffic will be much cheaper or even free. For additional data cost-saving tips, check this [article on how to reduce data transfer costs on AWS](https://www.stormit.cloud/blog/aws-data-transfer-pricing-how-to-reduce-costs/). Our examples are made with AWS in mind. However, the same suggestions also apply to other cloud providers.
 
-k6 needs a significant amount of memory when uploading files, as every VU is independent and has its own memory.
+Virtual server costs
+: The AWS EC2 instances are relatively cheap. Even the largest instance we have used in this benchmark (m5.24xlarge) costs only $4.6 per hour.
+: Make sure you turn off the load generator servers once you are done with your testing. A forgotten EC2 server might cost thousands of dollars per month.
+**Tip.** It's often possible to launch "spot instances" of the same hardware for 10-20% of the cost.
```
imiric, line 248:
```suggestion
**Tip:** it's often possible to launch "spot instances" of the same hardware for 10-20% of the cost.
```

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -242,49 +263,55 @@ WARN[0013] Request Failed       error="Get http://test.k6.io: read tcp 172.31.72
 
 ### context deadline exceeded
 
-This happens when k6 was able to send a request, but the target system didn't respond in time. The default timeout in k6 is 60 seconds. If your system doesn't produce the response in this time frame, this error will appear.
+This happens when k6 could send a request, but the target system didn't respond in time. The default timeout in k6 is 60 seconds. If your system doesn't produce the response in this time frame, this error will appear.
```
imiric, line 266: I prefer "was able to" here. It has more of an "ability" tone of something that happened in the past.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -242,49 +263,55 @@ WARN[0013] Request Failed       error="Get http://test.k6.io: read tcp 172.31.72
...
 
 ```bash
 WARN[0064] Request Failed    error="Get http://test.k6.io: context deadline exceeded"
 ```
 
 ### dial tcp 52.18.24.222:80: i/o timeout
 
-This is a similar error to the one above, but in this case, k6 wasn't even able to make an HTTP request. The target system isn't able to establish a TCP connection.
+This error is similar `context deadline exceeded`, but in this case, k6 could not even make an HTTP request. The target system can't establish a TCP connection.
```
imiric, line 274:
```suggestion
This error is similar to `context deadline exceeded`, but in this case, k6 wasn't even able to make an HTTP request. The target system can't establish a TCP connection.
```

Same here about "could not".

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -242,49 +263,55 @@ WARN[0013] Request Failed       error="Get http://test.k6.io: read tcp 172.31.72
...
 WARN[0057] Request Failed     error="Get http://test.k6.io/: dial tcp 52.18.24.222:80: i/o timeout"
 ```
 
 ### socket: too many open files
 
-This error means that the load-generator machine isn't able to open TCP sockets because it reached the limit of open file descriptors.
-Make sure that your limit is set sufficiently high. See our ["Fine tuning OS" article](/misc/fine-tuning-os/#viewing-limits-configuration).
+This error means that the load generator can't open TCP sockets because it reached the limit of open file descriptors.
+Make sure that your limit is set sufficiently high.
+Refer to ["Fine tuning OS" article](/misc/fine-tuning-os/#viewing-limits-configuration).
```
imiric, line 284:
```suggestion
Refer to the ["Fine tuning OS" article](/misc/fine-tuning-os/#viewing-limits-configuration).
```

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -311,10 +338,10 @@ k6 run --execution-segment "3/4:1"     --execution-segment-sequence "0,1/4,2/4,3
...
+- k6 does not provide the functionality of a "primary" instance to coordinate the distributed execution of the test. Alternatively, you can use the [k6 REST API](/misc/k6-rest-api) and [`--paused`](/using-k6/k6-options/reference#paused) to synchronize the multiple k6 instances' execution.
+- Each k6 instance evaluates [Thresholds](/using-k6/thresholds) independently - excluding the results of the other k6 instances. If you want to disable the threshold execution, use [`--no-thresholds`](/using-k6/k6-options/reference#no-thresholds).
 - k6 reports the metrics individually for each instance. Depending on how you store the load test results, you'll have to aggregate some metrics to calculate them correctly.
 
 With the limitations mentioned above, we built a [Kubernetes operator](https://github.com/grafana/k6-operator) to distribute the load of a k6 test across a **Kubernetes cluster**. For further instructions, check out [the running distributed k6 tests on Kubernetes tutorial](https://k6.io/blog/running-distributed-tests-on-k8s/).
```
imiric, line 347:
```suggestion
With the limitations mentioned above, we built a [Kubernetes operator](https://github.com/grafana/k6-operator) to distribute the load of a k6 test across a **Kubernetes cluster**. For further instructions, check out [the tutorial for running distributed k6 tests on Kubernetes](https://k6.io/blog/running-distributed-tests-on-k8s/).
```

imiric (changes requested):
Nice work @MattDodsonEnglish! :clap:

It looks much better now. :+1: And you were right, `DescriptionList` is a good fit, and separates some of the topics nicely.

Just left some minor comments. Some more nitpicky than others :wink:

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
```diff
@@ -38,42 +42,72 @@ For detailed information about these settings, how to make them permanent, and i
...
 
-### CPU
-
-Unlike many other load testing tools, k6 is heavily multi-threaded. It will effectively use all available CPU cores.
-
-The amount of CPU you need depends on your test files (sometimes called test script).
+CPU
+: Unlike many other load testing tools, k6 is heavily multi-threaded. It will effectively use all available CPU cores.
+: The amount of CPU you need depends on your test script and associated files.
 Regardless of the test file, you can assume that large tests require a significant amount of CPU power.
```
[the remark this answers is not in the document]
MattDodsonEnglish: It's a feature of extended markdown: https://www.markdownguide.org/extended-syntax#definition-lists
I don't like the syntax, but it's gained adoption.


The parsers of many SSGs use it, so it'd work in Hugo out of the box. Our Gatsby component has the disadvantage of requiring a `<DescriptionList>` tag, but the advantage of automatically generating `id` attributes, so each term is has an anchor.

I don't think it works it Github though
: does it?

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
MattDodsonEnglish, line 108:
```suggestion
An oversight of some scripts is to test with only the [happy path](/misc/glossary#happy-path "The default behavior that happens when the system returns no errors") in mind.
```

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
MattDodsonEnglish: Exactly. I'd prefer to format this and the error headings as titles, but those are the breaks.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
MattDodsonEnglish: How about "can"? I should've realized the real problem was the needless past tense.

MattDodsonEnglish, line 266:
```suggestion
This happens when k6 can send a request, but the target system doesn't respond in time. The default timeout in k6 is 60 seconds. If your system doesn't produce the response in this time frame, this error will appear.
```

MattDodsonEnglish:
> It looks much better now. +1 And you were right, `DescriptionList` is a good fit, and separates some of the topics nicely.
> 
> Just left some minor comments. Some more nitpicky than others wink

Thanks! These were all great. Really, you did all the work I should have done myself.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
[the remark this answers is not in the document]
imiric: Ah, got it. I just noticed it's not just line breaks, but separate elements.

Then I'd argue we're abusing the purpose of _definition_ lists, just because of how they're currently styled. It's akin to using tables for layout in the old days, and goes against the semantic HTML concept.

There's no reason these couldn't be wrapped in an element just like `<DescriptionList>`, but contain plain `h3`s or whatever level is appropriate, and then the paragraphs could just have a small `padding-left`. It would require trivial amounts of CSS, the markup would be much simpler, and you could just do a double line break as usual to get new paragaphs. And we wouldn't need to fix issues like #1002, at least not right now.

I don't want to block this PR from merging because of this, but I can create an issue if you agree.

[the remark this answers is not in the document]
MattDodsonEnglish: I don't think it's abusing a `<dl`>`. It is actually called a _description list_ by MDN. I swear it used to be called definition list and then they broadened the name. I'm not going to dig, but the standard says the lists can be any name-value groups:

```
Name-value groups may be terms and definitions, metadata topics and values, questions and answers, or any other groups of name-value data.
```

In these cases, I think the description list works well semantically, because these all are lists in one distinct category, e.g. "Specific hardware considerations".  In another form, the info could be represented as a table, since their structure is so similar. On the other hand, the grouping of h3s can be a little looser and usually don't have the same parallellism between heading and content.

But, this is very, very pedantic and I'm being a little bit vague and mystical. Still, I don't think it's so important and I don't think we're doing anything that abuses web conventions or harms accessibility. If you think we are, please make an issue.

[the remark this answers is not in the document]
imiric: Fair enough, the naming is confusing, but all the usage examples point to this being used for defining or describing terms, for metadata, and generally simpler usage than what we're using them for (#1002 would be something we shouldn't need).

The reason the colon is used is precisely to define the term, and that's not our use case here. On the glossary page it makes perfect sense, but rarely elsewhere, and definitely not just for indentation.

I'd also point out [this note from MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl#notes):

> Do not use this element (nor `<ul>` elements) to merely create indentation on a page. Although it works, this is a bad practice and obscures the meaning of description lists.
>
> To change the indentation of a description term, use the CSS `margin` property.

Which is exactly what we're doing here. It also has accessibility implications for screen readers, etc., and we don't need to complicate things.

So let's merge this as is, but I'll create the issue to change this.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
imiric: But the sentence is in past tense already, "can" wouldn't be a good fit either.

Why are you against "was able to"?

imiric (approved):
LGTM, besides https://github.com/grafana/k6-docs/pull/1003#discussion_r1089087450, which we can address later.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
MattDodsonEnglish: The main clause is "This happens", which is present.

> Why are you against "was able to"?

Ha, yes, you've hit the heart of the issue.

"Is able to" longer than "can". That's the entire reason.

 I'm also generally against the past tense because it creates longer constructions and creates a grammatical distance that isn't necessary unless describing real pat tenses.

MattDodsonEnglish: But this is an extremely minor point. I'm fine with you choosing whatever you prefer, and merging when you think it's ready.

MattDodsonEnglish:
Nice, thanks for your great help (and patience :-D), I don't want to merge
and create any conflicts for you in the main PR, so feel free to merge
whenever you think it's ready.

src/data/markdown/translated-guides/en/06 Testing Guides/04 Running large tests.md
imiric: We're really bikeshedding at this point :sweat_smile:, and I don't want to drag this on, but "this happens" is describing an action that happened in the past. I.e. you ran the test, got the error, and then "this happens when ...". You could write it either way, but describing a past action is more relatable to how the user would experience it.

Still, regardless of the tense, even the present "is able to" would be better than "can", since "can" implies that k6 has the _capability_ to do something, which, of course it does, it _can_ make requests; but "able to" implies the _ability_ in that moment, i.e. it's _unable_ to because of this or that reason.

But nevermind, it's fine as is. :smile:

MattDodsonEnglish: If you search google for "this {happens|occurs} when", you'll see most sentences are in the present tense. It's because the phrase introduces a general condition. In our sentences, no event _really_ occurred; we just describe the general conditions that create the error, as we understand them presently.

But, it's an interesting point about "can" vs. "is able to". I think in most cases, I'd only need "is unable to" but you've definitely got me questioning my "no-is-able-to" orthodoxy here.

I now realize the most direct way would be to remove all auxiliary verbs:

> This happens when k6 sends a request and the target system doesn't respond in time.

This particular case is not a big deal in the scheme of things, but I do appreciate any discussion that brings more attention to precision and clarity in language.

imiric: Good points :+1: I agree that the simplest form is to remove all auxiliary verbs, as in your example there. Consistency is probably most important, and you have the big picture in mind. So I leave this up to you.

Thanks for the discussion :slightly_smiling_face:

Issue #900, opened 2022-11-16
Document the importance of sleep


MattDodsonEnglish:
For developers who don't come from a testing background, the `sleep()` function can be confusing. Why waste CPU cycles doing nothing?

We have sleep recommendations in the browser recorder, a mention in the "high CPU" insight, and some mentions in k6-learn, but it'd be nice to have a more findable description, since sleep is in so many of our example scripts, and little "best practices" like this are some of the most requested docs.

imiric:
We could just link to https://www.sleepfoundation.org/how-sleep-works/why-do-we-need-sleep :trollface:

But to add something constructive: `sleep()` is very important to simulate actual user behavior. Users don't just hammer a web site like a bot would, and a lot of time is spent rendering and interacting with the frontend. Running a browser test represents that scenario better, and part of the use case for k6 _is_ to stress the SUT and not behave exactly like a user would. But even in that case, `sleep()` is useful to pace the generated load, and not overload the load generator machine. Given that k6 is not strictly a _load_ testing tool, but encompasses many different areas of observability, waiting for a specific amount of time is useful in other testing scenarios as well. So I agree that we should have a document to mention some of these examples.

sniku:
I think the most important is to document that sleep() is a bad idea for arrival rate executors. At least, sleep should not be added at the end of the iteration, as we recommend for other executors.

mstoykov:
On similar note :point_up:  we likely should get something like [delay](https://deno.land/x/delay@v0.2.0) for asynchronous code as `sleep` is blocking and synchronous.

MattDodsonEnglish:
> I think the most important is to document that sleep() is a bad idea for arrival rate executors. At least, sleep should not be added at the end of the iteration, as we recommend for other executors.

I just made a small note to account for this, if someone wants to review.

https://github.com/grafana/k6-docs/pull/1008/files

PR #1008, opened 2023-01-27
Sleep isn't needed at end of arrival rate iter

MattDodsonEnglish:
From discussion in #900

### DIFF
--- a/src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/01 Executors/05 constant-arrival-rate.md
+++ b/src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/01 Executors/05 constant-arrival-rate.md
@@ -39,8 +39,7 @@ this executor has the following options:
 ## When to use
 
 When you want to maintain a constant number of iterations without being affected by the
-performance of the system under test. To reliably achieve a fixed request rate, try to keep test function simple, with preferably only a single request call,
-and no additional processing or `sleep()` calls.
+performance of the system under test.
 
 <Blockquote mod="note" title="Don't put sleep at the end of an iteration">
 
--- a/src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/01 Executors/05 constant-arrival-rate.md
+++ b/src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/01 Executors/05 constant-arrival-rate.md
@@ -16,7 +16,6 @@ useful for a more accurate representation of RPS, for example.
 For explanations about how this executor works, refer to [Open and Closed models](/using-k6/scenarios/concepts/open-vs-closed)
 and [Arrival-rate VU allocation](/using-k6/scenarios/concepts/arrival-rate-vu-allocation).
 
-
 <Blockquote mod="Note" title="Iteration starts are spaced fractionally">
 
 Iterations **do not** start at exactly the same time.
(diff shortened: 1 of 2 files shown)


src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/01 Executors/05 constant-arrival-rate.md
```diff
@@ -40,7 +39,15 @@ this executor has the following options:
 ## When to use
 
 When you want to maintain a constant number of iterations without being affected by the
-performance of the system under test.
+performance of the system under test. To reliably achieve a fixed request rate, try to keep test function simple, with preferably only a single request call,
```
mstoykov, line 42: I don't think that recommending to only have 1 request is a good thing. While this will simplify calculating how many VUs you need in normal cases it :

1. in practice makes it seem like the executor can't handle anything more complicated, which is far from the truth
2. again in the case that the SUT is failing the "request should take no more than 1s" thinking is still not going to fix anything when it is taking 1m ;)

MattDodsonEnglish: Yeah I just moved that from the old section; it struck me as pretty much invalidating all complex uses of arrival rate. Will delete.

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/01 Executors/05 constant-arrival-rate.md
```diff
@@ -40,7 +39,15 @@ this executor has the following options:
...
+and no additional processing or `sleep()` calls.
```
MattDodsonEnglish, line 43:
```suggestion
performance of the system under test.
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/01 Executors/05 constant-arrival-rate.md
MattDodsonEnglish: [e8751a1](https://github.com/grafana/k6-docs/pull/1008/commits/e8751a19a8555e8ba79f77d2e80c51b50e476cf5)

MattDodsonEnglish:
One other thing to note, I wrote "It's unneeded" whereas @sniku wrote "It's a bad idea". We can make the wording a little stronger if the reason it is a bad idea is something more than "It's unnecessary." But if we do that, we should also explain a bit more.

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/01 Executors/05 constant-arrival-rate.md
```diff
@@ -42,6 +41,13 @@ this executor has the following options:
 When you want to maintain a constant number of iterations without being affected by the
 performance of the system under test.
 
+<Blockquote mod="note" title="Don't put sleep at the end of an iteration">
+
+The arrival-rate executors already pace the iteration rate through the `rate` and `timeUnit` properties.
+It's unnecessary to use a `sleep()` function at the end of the VU code.
```
mstoykov, line 47:
```suggestion
So it's unnecessary to use a `sleep()` function at the end of the VU code.
```

mstoykov (approved):

PR #1016, opened 2023-01-30
Executors, explain vu cycling in arrival-rate

### LINKED ISSUE #136: Document how VUs are cycled in arrival-rate executors
Prompted by https://github.com/loadimpact/k6/pull/1623#discussion_r505206848, we should make sure to explain that the arrival-rate executors are going to cycle over all of the initialized VUs, given enough time, even if they only need a small fraction of them at any given point to run their allotted iterations per second.

MattDodsonEnglish:
Closes #136

### DIFF
--- a/src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 Concepts/02 Arrival-rate VU allocation.md
+++ b/src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 Concepts/02 Arrival-rate VU allocation.md
@@ -112,3 +112,13 @@ Some of the times it might make sense to use `maxVUs` include:
 - To determine necessary allocation in first-time tests
 - To add a little "cushion" to the pre-allocated VUs that you expect the test needs
 - In huge, highly distributed tests, in which you need to carefully scale load generators as you increment VUs.
+
+## Arrival-rate executors cycle over all VUs
+
+Given enough time, k6 cycles over all VUs initialized, even if only a small fraction of them are needed to reach the target rate.
+
+This fact has implications for the [execution context variables](/docs/using-k6/execution-context-variables/).
+For example, if your arrival-rate scenario allocates 15 vus runs long enough,
+then the value of `{exec.vu.iterationInScenario}` will reach `15`,
+even if k6 never requires that number of VUs to reach the target iteration.
+
--- a/src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 Concepts/02 Arrival-rate VU allocation.md
+++ b/src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 Concepts/02 Arrival-rate VU allocation.md
@@ -113,12 +113,8 @@ Some of the times it might make sense to use `maxVUs` include:
 - To add a little "cushion" to the pre-allocated VUs that you expect the test needs
 - In huge, highly distributed tests, in which you need to carefully scale load generators as you increment VUs.
 
-## Arrival-rate executors cycle over all VUs
+## You can't guarantee which VU runs an iteration
 
-Given enough time, k6 cycles over all VUs initialized, even if only a small fraction of them are needed to reach the target rate.
-
-This fact has implications for the [execution context variables](/docs/using-k6/execution-context-variables/).
-For example, if your arrival-rate scenario allocates 15 vus runs long enough,
-then the value of `{exec.vu.iterationInScenario}` will reach `15`,
-even if k6 never requires that number of VUs to reach the target iteration.
+As with all executors, you can't predict the specific VU that an arrival-rate executor uses for a specific iteration.
+As the test runs on, the executor might use some or all of the allocated VUs, even if it never needs the entire allocated number to reach the iteration rate.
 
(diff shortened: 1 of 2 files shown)


src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 Concepts/02 Arrival-rate VU allocation.md
```diff
@@ -112,3 +112,13 @@ Some of the times it might make sense to use `maxVUs` include:
 - To determine necessary allocation in first-time tests
 - To add a little "cushion" to the pre-allocated VUs that you expect the test needs
 - In huge, highly distributed tests, in which you need to carefully scale load generators as you increment VUs.
+
+## Arrival-rate executors cycle over all VUs
+
+Given enough time, k6 cycles over all VUs initialized, even if only a small fraction of them are needed to reach the target rate.
+
+This fact has implications for the [execution context variables](/docs/using-k6/execution-context-variables/).
+For example, if your arrival-rate scenario allocates 15 vus runs long enough,
```
MattDodsonEnglish, line 121:
```suggestion
For example, if your arrival-rate scenario allocates 15 vus and runs long enough,
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 Concepts/02 Arrival-rate VU allocation.md
```diff
@@ -112,3 +112,13 @@ Some of the times it might make sense to use `maxVUs` include:
...
+then the value of `{exec.vu.iterationInScenario}` will reach `15`,
```
MattDodsonEnglish, line 122:
```suggestion
then the value of `{exec.vu.iterationInScenario}` will reach `15` at some points,
```

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 Concepts/02 Arrival-rate VU allocation.md
```diff
@@ -112,3 +112,13 @@ Some of the times it might make sense to use `maxVUs` include:
...
+
+## Arrival-rate executors cycle over all VUs
+
+Given enough time, k6 cycles over all VUs initialized, even if only a small fraction of them are needed to reach the target rate.
+
+This fact has implications for the [execution context variables](/docs/using-k6/execution-context-variables/).
+For example, if your arrival-rate scenario allocates 15 vus and runs long enough,
+then the value of `{exec.vu.iterationInScenario}` will reach `15` at some points,
+even if k6 never requires that number of VUs to reach the target iteration.
+
```
mstoykov, line 124:
```suggestion
then the value of `{exec.vu.iterationInScenario}` will reach `15` at some points,
even if k6 never requires that number of VUs to reach the target iteration *at the same time*.

```

mstoykov: Otherwise I can read this as even if I only had to to 10 iterations (1iter/s for 10s) but got 15 VUs I will still use all 15 VUs ... somehow

MattDodsonEnglish: Hmm I'm a little lost at both yours and mine.

Maybe:

> even if the executor never requires all 15 VUs to reach the target rate.

WDYT?

MattDodsonEnglish:
If it's still confusing with that sentence, I think it might be easier to communicate visually.

> For example, if you allocate 4 vus and the target never requires more than two active, k6 still cycles through all four if the test runs long enough. This is to say that the value of `{exec.vu.iterationInScenario}` will be `3` and `4` at some iterations.

image

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/00 Concepts/02 Arrival-rate VU allocation.md
mstoykov: I mean it is better as *rate* is the key word. My problem is that somebody who doesn't understand this all that well likely will be just as confused. 

IMO it is better to be more explicit in what is that we are trying to say when it is confusing to begin with. 

In this case for me the key parts are:
1. k6 will use at most as many VUs *at a time* to hit the provided rate
2. this VUs will not necessarily be the first N VUs *only* even if only N or less VU are needed at any given moment.

I am not certain guaranteeing that it will go through all the VUs is a good idea - this is an implementation detail that likely is never necessary to be known.

src/data/markdown/translated-guides/en/02 Using k6/14 Scenarios/01 Executors.md
```diff
@@ -9,6 +9,19 @@ want to model to test your services.
 
 Possible values for `executor` are the executor name separated by hyphens.
 
+<Blockquote mod="note" title="VUs might not distribute uniformely over iterations">
+
+For any given scenario, you can't guarantee that a specific VU can run a specific iteration.
+
+With [`SharedArray`](/javascript-api/k6-data/sharedarray/) and [execution context variables](/using-k6/execution-context-variables/), you can map a specific VU to a specific value in your test data.
+So the tenth VU could use the tenth item in your array.
```
mstoykov, line 17:
```suggestion
So the tenth VU could use the tenth item in your array. Or the sixth iteration to the sixth item.
```

I would expect in most cases people care about the iteration getting *specific* or (more usually IMO) different data from the SharedArray and less about a specific VU getting one or the other.

MattDodsonEnglish: Yeah thanks, this is better. My main goal was to prevent readers from thinking its impossible to link VUs to specific test data, which would be a much bigger hassle

na-- (approved):

mstoykov (approved):

PR #1026, opened 2023-02-02
 redo eslint configuration for examples 

mstoykov:
- Don't forbid async/await
- Don't enable everything that will work in node
- Add a bit more globals
- raise parser version to support async/await

This still means that if a syntax is not supported by the parser it will
give an error asking the user to add `requireConfigFile: false` instead
of telling them "you can't use this syntax".

Unfortunately it seems like the only way to fix that is to configure
this per each syntax which is a PITA. Hopefully k6 will soon support
even bigger parts of the syntax and that will be even less of a problem.

### DIFF
--- a/src/data/markdown/.eslintrc.js
+++ b/src/data/markdown/.eslintrc.js
@@ -1,12 +1,11 @@
 module.exports = {
   root: true,
   env: {
-    browser: true,
-    node: true,
-    es6: true,
+    es2017: true,
+    commonjs: true,
   },
   parserOptions: {
-    ecmaVersion: 7,
+    ecmaVersion: '2017',
     sourceType: 'module',
   },
   extends: ['plugin:mdx/recommended', 'plugin:prettier/recommended'],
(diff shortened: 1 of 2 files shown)


ppcano:
Thanks Mihail. This should also be ported to the [blog](https://github.com/grafana/k6-blog/blob/main/src/data/markdown/posts/.eslintrc.js).

src/data/markdown/.eslintrc.js
```diff
@@ -8,6 +8,7 @@ module.exports = {
   parserOptions: {
     ecmaVersion: 7,
```
EdvinasDaugirdas, line 9: Change this to `ecmaVersion: 8` and you shouldn't need `requireConfigFile: false`.

7 version does not recognise `async/await` syntax, although it's pretty weird that it notifies such things as a `babel` config error 🤷

mstoykov: Yeah that is true - I have now redone the whole PR

EdvinasDaugirdas (approved):
LGTM 🚀

Issue #955, opened 2022-12-19
Remove old JavaScript modules versions


codebien:
- [x] Define a policy for the definition of old (e.g. not expected to be maintained). Last 4 versions.

- [ ] Update the [release process documentation](https://github.com/grafana/k6-docs/wiki/Add-version-for-Javascript-API-documentation) including the defined policy.

- [ ] Remove the old JavaScript modules' versions.

ppcano:
Should be closed by https://github.com/grafana/k6-docs/commit/5cfe9eb2d77afacb0f4e4b0e5c09be8bef3591f5

Will start with a limit of the latest 6 versions. But we'll change the value of `MAX_JS_API_VERSIONS_TO_BUILD` depending on final decision. 

> Remove the old JavaScript modules' versions.

To start, we can do this manually.

> Update the [release process documentation](https://github.com/grafana/k6-docs/wiki/Add-version-for-Javascript-API-documentation) including the defined policy.

This guide is for creating a new version. We can wait for updating it until we have decided the removal process.

ppcano:
@olegbespalov, reopened.

Issue #966, opened 2022-12-23
Improve documentation versioning


imiric:
The current way of documenting specific k6 versions (introduced in #283) has some drawbacks and limitations:

- The process involves duplicating the entire `javascript api` tree, which is inefficient and bloats the repository, but more importantly carries layout and content changes that need to be repeated on all versions to be consistent.
  If you take a look at [the landing page for the latest v0.42.0 version](https://k6.io/docs/javascript-api/), you'll see that the "Init context" explanation is different from v0.40.0 and earlier. If we need to make layout or style changes, it needs to be done on all versioned copies.

- It only versions JS API docs. Sometimes we make changes to [k6 options](https://k6.io/docs/using-k6/k6-options/reference/), but there's no way for the user to see which version is documented. So we resort to phrases like "New since v...", which often goes stale, and is not relevant after those versions are widely deployed. See the discussion in https://github.com/grafana/k6-docs/pull/965#discussion_r1055905689.

- It doesn't version any of the extension docs. xk6-browser and xk6-disruptor are stuck with "latest". This documentation shouldn't be only under the "JavaScript API" section, since it has an introduction, installation, "Getting started" guides, etc., but maybe that's a separate issue.

I don't have a specific proposal for fixing this, but I'm happy to brainstorm some solutions with the PixelPoint team.

**Related issues:** #675, #955

ppcano:
> you'll see that the "Init context" explanation is different from v0.40.0 and earlier. If we need to make layout or style changes, it needs to be done on all versioned copies.

Why? I am not sure this is a must. My approach is to fix the latest version and the old ones are unmaintained (saved for reference).

> So we resort to phrases like "New since v...", which often goes stale

Agree. We should do better on updating these notes. 

After how many new versions should we drop the version note? 2, 4, 6, 8?

> It doesn't version any of the extension docs. xk6-browser and xk6-disruptor are stuck with "latest". 

Agree. This has been considered internally but not prioritized.

heitortsergent:
Closing this since we've moved the docs to Grafana.

PR #1029, opened 2023-02-06
Drop version notes

ppcano:
This PR removes old version notes. 

The general consensus is to preserve notes from the latest four k6 versions. 

- TODO: document it somewhere as commented at https://github.com/grafana/k6-docs/issues/955


imiric (changes requested):
Looks good!

I think we're missing a couple from the Spanish translation:
```bash
$ rg -c -g '*.md' -g '!**/v0.*/**' '<sup>.*v0'
src/data/markdown/translated-guides/es/02 Using k6/02 Metrics.md:1
src/data/markdown/translated-guides/es/02 Using k6/01 HTTP requests.md:1
```

Given our discussion in #966, you probably don't want to change any of the old versions, but I think we should at least update the v0.4x versions as well, since they're fairly recent.

There are 30 files with the `<sup>` notation:
```bash
> rg -l -g '**/v0.4*/**/*.md' '<sup>.*v0' | wc -l
30
```

But there might be more with the "since" phrasing and other forms you removed here.

BTW, this is another example of the maintenance issue with copying entire directory trees for versioning. Note that it also applies for translated versions, since we're not using internationalized strings everywhere.

ppcano:
The current plan is to remove the translations on the docs but preserve them on the blog.

> you probably don't want to change any of the old versions

Exactly. 

Even, if we finally agree on updating a limited number of previous versions, my suggestion would be to update only fixes. Those are, imo, cosmetic improvements.

PR #1050, opened 2023-02-16
Fix code example in 01 Javascript Extensions.md

ifross89:
The code sample has some compilation errors that stop it from running.

This change fixes the errors so the example is able to be built & run.


imiric (approved):
Thank you! :bow:

PR #978, opened 2023-01-10
Test types, make spike ramping arrival rate

MattDodsonEnglish:
- And some general language and formatting clean up.
- Fixes #826

PR #1056, opened 2023-02-21
Fix images and text on "Stress testing" guide

imiric:
Some of the images on the ["Stress testing" guide](https://k6.io/docs/test-types/stress-testing/) were removed and some text was mangled in #978.

It seems @MattDodsonEnglish intentionally removed them in 74b33b39 because they were "misleading", but I'm not sure why he thought that. The images help visualize the test behavior, and not having them is worse, IMO. Maybe the exact VU number will change on different test runs, but the point is to show how the script options affect the load.

In any case, if we decide to remove them, we should also remove the text that introduces them, also change the Spanish translation, and remove the images themselves from the repo.

### DIFF
--- a/src/data/markdown/translated-guides/en/05 Test Types/03 Stress testing.md
+++ b/src/data/markdown/translated-guides/en/05 Test Types/03 Stress testing.md
@@ -93,8 +93,9 @@ export default function () {
 </CodeGroup>
 
 The VU chart of a stress test should look similar to this:
-5 minutes. We have also included a recovery stage at the end, where the system is gradually
-decreasing the load to 0.")
+![Virtual user chart of an API stress test](./images/stress-test.png)
+
+This configuration increases the load by 100 users every 2 minutes and stays at this level for 5 minutes. We have also included a recovery stage at the end, where the system is gradually decreasing the load to 0.
 
 If your infrastructure is configured to auto-scale, this test will help you to determine:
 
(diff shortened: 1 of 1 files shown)


MattDodsonEnglish:
FYI, I removed that image after a suggestion from mystoykov: https://github.com/grafana/k6-docs/pull/978#pullrequestreview-1249733302. Sorry, I should have linked that in my commit.

However, I think specific graph could run the risk of confusing users if it doesn't match the script. Maybe instead we could abstract these images to simple line graphs, not connected to any number or graph. Obviously this example is rather crude:

image


https://community.k6.io/t/about-the-number-of-users-in-the-stress-test-documentation/6131

imiric:
Well, we still need to label the line in _some_ way, otherwise it's completely meaningless. Maybe RPS would be a better indicator, but I really don't see why expressing the load in terms of VUs would be confusing, since there's usually a direct correlation between RPS and amount of VUs.

[This reporting issue](https://github.com/grafana/k6-docs/pull/978#discussion_r1070952585) was recently fixed, so it would report the actual number of VUs in use. Yet if we graphed it now, it would appear very "jumpy", and it wouldn't illustrate the behavior properly.

The confusion in that forum post is mostly because of the precise phrasing of "this configuration increases the load by 100 users every 2 minutes". I think it would be enough to soften the language that describes the graphs, without putting so much emphasis on VUs, and explain that the graphs are illustrative of the load, and don't describe the script options _exactly_.

But I definitely wouldn't remove the graphs altogether, as they're very helpful to understand the behavior.

PR #1060, opened 2023-02-27
Enable display in sidebar

javaducky:
This page is being referenced from the externally-controlled executor page but is not discoverable otherwise. Discovered while updating incorrect link reported by https://k6io.slack.com/archives/C5C0281HB/p1677269825314019.


ppcano:
This is intentional. 

Initially, we used this API for two use cases:
- building a web app to visualizing results in real time. 
- coordinate distributed execution on k6 Cloud.

We now recommend output extensions or the operator for these cases. So, we didn't want to promote the REST API because it is experimental and might be deprecated.

PR #1143, opened 2023-04-21
Update example file path in Docker setup instructions

ch4r-ch4r:
Fix incorrect file path in Docker setup instructions for k6, InfluxDB, and Grafana. Change the path from $PWD/samples to $PWD/examples to reflect the correct location of the es6sample.js file.

### DIFF
--- a/src/data/markdown/translated-guides/en/04 Results output/200 Real-time/00 InfluxDB - Grafana.md
+++ b/src/data/markdown/translated-guides/en/04 Results output/200 Real-time/00 InfluxDB - Grafana.md
@@ -117,7 +117,7 @@ $ docker-compose up -d \
     influxdb \
     grafana
 $ docker-compose run -v \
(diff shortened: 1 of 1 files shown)


javaducky (approved):
LGTM! Thank you!

MattDodsonEnglish:
Oh, now I see that this entire page has been refactored now and that example no longer exists :grimacing: . I'm sorry to both of you for being late.

However this is a great callout, and it made me think about some other places where the docs may need to be fixed to update that samples dir.

MattDodsonEnglish:
Closing because the reference to that file seems to have disappeared. Still, thanks @ch4r-ch4r for pointing this out: it led to fixes on other pages.

[2d2737b](https://github.com/grafana/k6-docs/commit/2d2737b6e7e73a1953001ae339ae8dfabfd4573c)

PR #1146, opened 2023-04-24
Add xk6-chai extension

szkiba:
Add [xk6-chai](https://github.com/szkiba/xk6-chai) extension that embeds Chai.js into the k6 binary.
Using this extension the test will not have runtime external JavaScript dependencies and also can speed up the test launch (since the embedded Chai.js is compiled only once during a k6 run)


javaducky (approved):
LGTM!

PR #1147, opened 2023-04-24
Fix, reorder and simplify threshold instructions

MattDodsonEnglish:
- Move instructions to set syntax to top of page
- Move extensive annotation directly to code comments
- Simplify the language and make it more scannable


src/data/markdown/translated-guides/en/02 Using k6/04 Thresholds.md
```diff
@@ -66,8 +67,138 @@ In this case, the test met the criteria for both thresholds.
...
+    // avoid using the same metric more than once here
+    // metric_name: [ 'count<100' ],
+    // metric_name: [ 'rate<50' ],
+  },
+};
+```
+
+</CodeGroup>
+
+Only the last one will remain.
```
MattDodsonEnglish, line 199:
```suggestion
    // don't use the same metric more than once here
    // metric_name: [ 'count<100' ],
    // metric_name: [ 'rate<50' ],
  },
};
```

</CodeGroup>

In such cases, only the last threshold will be evaluated.
```

ppcano (approved):

PR #1150, opened 2023-04-25
feat; document when http-metric timestamps happen

### LINKED ISSUE #1051: Document what is the timestamp for each metric
For this [forum question](https://community.k6.io/t/constant-rate-arrival-and-http-requests-timestamps/5910) it was relevant that the timestamp of http_req_* is of the *end* of the request not the beginning. 

While this might be less relevant for other metrics it might be a good idea to *actually* document it so that it is easily findable.

MattDodsonEnglish:
Fixes #1051

### DIFF
--- a/src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
+++ b/src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
@@ -95,7 +95,7 @@ These metrics are generated only when the test makes HTTP requests.
 <Blockquote mod="note" title="">
 
 For all `http_req_*` metrics, **the timestamp is emitted the end of the request.**
-In other words, the timestamp happens when k6 receives a return status, or the request times out.
+In other words, the timestamp happens when k6 receives the end of the response body, or the request times out.
 
 </Blockquote>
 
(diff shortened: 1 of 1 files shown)


MattDodsonEnglish:
@mstoykov, I'm flagging you because you opened the issue. Also, I'm not sure that my definition of when "a request ends" is strictly correct, but I figured it'd be better to just make a PR and receive a correction than carry on some pre-discussion.

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
```diff
@@ -92,6 +92,14 @@ k6 always collects the following built-in metrics:
 
 These metrics are generated only when the test makes HTTP requests.
 
+<Blockquote mod="note" title="">
+
+For all `http_req_*` metrics, **the timestamp is emitted the end of the request.**
+In other words, the timestamp happens when k6 receives a return status, or the request times out.
```
mstoykov, line 98:
```suggestion
In other words, the timestamp happens when k6 receives the end of the response body, or the request times out.
```

It is not just the status. We do actually read the whole body before we emit the metrics

PR #1153, opened 2023-04-27
refactor: remove networkidle option on browser tests

mdcruz:
As agreed with the k6 browser team, we will remove the use of `networkidle` on our browser docs because this option is not recommended by Playwright.


ankur22 (approved):
LFTM! Thanks for this 🎉

PR #1162, opened 2023-05-05
docs: add k6 version for the browser module

mdcruz:
As suggested by one of our users, https://k6io.slack.com/archives/C3E688J8Z/p1683213296899049, this PR adds which k6 version you need to be in to use the browser module


immavalls (approved):
LGTM, thanks @mdcruz 🙏

Issue #1160, opened 2023-05-04
Link JS API reference with corresponding metrics referenc


MattDodsonEnglish:
With #1152 , the metrics have been restructured into discrete topics.

However, we still need to add some connective tissue linking them with the API docs. We should link the reference of metrics  to the corresponding JS API doc. Specifically we need to document metrics from experimental modules (if they exist).

I'm not sure what's the best strategy:

- Duplicate the metrics. Tables on both k6.io/using-k6/metrics/reference and the corresponding API doc. If so, it probably would be better to make a shared component to ensure that both versions stay accurate.
- Just link to the metrics from the API reference. A little less convenient, perhaps, but simpler to write and less text to read.

PR #1173, opened 2023-05-10
Link api modules to the metric reference

MattDodsonEnglish:
This PR could be rebased; the commits build on each other

Part of #1160


olegbespalov (approved):
Looks good to me :+1: 

One possible improvement that we could apply is removing the legacy API methods ( `Socket.send(data)` and the `Socket.on('message', callback).`) from the WebSocket metrics list(`/using-k6/metrics/reference#websockets`). That way, the list will be API agnostic.

MattDodsonEnglish:
That's a great point. I think I fixed with  https://github.com/grafana/k6-docs/pull/1173/commits/23d2be44b486619af96337db354ade93f691a851.

Do you mind confirming? I want to make sure that I should only remove the link to the methods, and not the entire metric row.

olegbespalov (approved):
yep, looks good to me :+1:

PR #1181, opened 2023-05-18
Add `Dockerfile-standalone` for a static docker build

roobre:
This PR provides an alternative way of building the docs locally, that aims to be more reproducible and shareable than the default `Dockerfile`.

The `Dockerfile-standalone` copies all files in the build step and instructs Gatsby to produce a production build, so it does not have a dependency on the local environment to run. It leverages docker build-time mounts to reuse Gatsby cache, making subsequent builds as fast as if they were local. Finally it serves the generated HTML using nginx, which makes the image lighter for sharing or hosting it somewhere.

Not sure if you will be interested on having this, feel free to just close the PR if you aren't :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,4 +56,17 @@ Then visit http://localhost:8100
 
 > If you want to re-run the `--build` command, you may get an error about not having access to delete a `cache/` folder. Use `sudo` to delete this manually before retrying.
 
+### Static Docker build
 
+Additionally, the `Dockerfile-standalone` Dockefile allows to produce a lightweight static build which also does not need local files to run. This can be useful in certain scenarios, for example for sharing the build output easily.
+
+In order to build the static version in docker, run:
+
+```shell
+docker build -f Dockerfile-standalone . -t k6-docs
+```
+
+To serve the site locally, run:
+```shell
+docker run --rm -ti -p 8080:80 k6-docs
+```
--- a/Dockerfile-standalone
+++ b/Dockerfile-standalone
@@ -1,0 +1,31 @@
+FROM node:16-slim as builder
+
+EXPOSE 8000
+
+RUN mkdir /app
+WORKDIR /app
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -56,4 +56,17 @@ Then visit http://localhost:8100
 
 > If you want to re-run the `--build` command, you may get an error about not having access to delete a `cache/` folder. Use `sudo` to delete this manually before retrying.
 
+### Static Docker build
 
+Additionally, the `Dockerfile-standalone` Dockefile allows to produce a lightweight static build which also does not need local files to run. This can be useful in certain scenarios, for example for sharing the build output easily.
```
pablochacin, line 61: Nit: given that you are calling it "static" and not "standalone" in the documentation, shouldn't the Dockefile also be called static?

pablochacin: The meaning of "does not need local files" is not clear to me, as it actually uses the local cache. Maybe it would be more accurate to say "does not need installing dependencies locally"? 

Second, I'm not sure I understand the use case this option is intended for.

pablochacin, line 61: > This can be useful in certain scenarios, for example for sharing the build output easily.

I'm not sure I understand the use case.

roobre: > Nit: given that you are calling it "static" and not "standalone"

Fair enough, if people find this interesting I'll rename it.

> it actually uses the local cache

It caches the output of the build step, but does not need to mount the local folder in runtime:

https://github.com/grafana/k6-docs/blob/aa4f410cf198cd13b283989798dd69091e4037fd/docker-compose.yml#L13

roobre: As the image does not depend on the local folder, you can `docker push` the image to serve it elsewhere if, for example, I wanted to share the site with someone I'm working with.

Probably not something you would do every day, but if you want to, this will help.

PR #1183, opened 2023-05-18
ci/pr-deploy: Combine build and upload jobs

roobre:
As discussed in https://github.com/grafana/k6-docs/pull/1182#issuecomment-1553334116, we spend around seven minutes copying a hundred-megabyte artifact across jobs. While bigger jobs are often not ideal, in this case I think it is preferable than the artifact upload/download overhead.

Setting this as a draft until #1182 is merged.


roobre:
During development of this PR, I noticed it was not able to reuse the build cache from other PRs. This is expected according to [the docs](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows):

> Workflow runs cannot restore caches created for child branches or sibling branches. For example, a cache created for the child feature-b branch would not be accessible to a workflow run triggered on the parent main branch. Similarly, a cache created for the feature-a branch with the base main would not be accessible to its sibling feature-c branch with the base main

In order for PRs to be able to reuse each other's cache, we need to also cache the gatsby build in the `staging` job, which in turn would also speed up this job. I'll set this up on a different (hopefully final) PR :)

imiric (approved):
Makes sense :+1:

I did a double-take and couldn't believe the 3m36s run time :eyes::grinning:

2023-05-22-125551_1354x234_scrot

That's a radical improvement over the previous 23m! :clap:

I'm slightly concerned that the aggressive caching won't bite us in the ass in some way later (inconsistent dependencies, wrong cache shared between PRs, etc.), but I currently can't see any drawbacks to this strategy, and we can address any issues with it later.

ppcano:
@roobre, thanks for improving the build process! 💯

PR #1188, opened 2023-05-22
ci/staging: leverage node_modules and gatsby cache on staging as well

roobre:
As mentioned in https://github.com/grafana/k6-docs/pull/1183#issuecomment-1556951247, and contrary to what I intended with #1182, the gatsby cache is not being currently reused across PRs. The reason why is outlined in the [GHA docs](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache):

> Workflow runs cannot restore caches created for child branches or sibling branches. For example, a cache created for the child feature-b branch would not be accessible to a workflow run triggered on the parent main branch. Similarly, a cache created for the feature-a branch with the base main would not be accessible to its sibling feature-c branch with the base main.

An easy way to fix this is to also use the cache on workflows targeting `main`, such as the deploy to staging job. This will also make this workflow faster.

### DIFF
--- a/.github/workflows/staging.yml
+++ b/.github/workflows/staging.yml
@@ -26,29 +26,9 @@ env:
   GATSBY_NEWSLETTER_FORM_ID: 1420
 
 jobs:
-  update-dependencies:
-    name: Update node module dependencies
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout project
-        uses: actions/checkout@v2
-      - name: Use Node.js 16.xs
-        uses: actions/setup-node@v2
-        with:
-          node-version: '16.x'
-          registry-url: 'https://registry.npmjs.org'
-      - name: Restore cache
-        uses: actions/cache@v2
-        with:
-          path: '**/node_modules'
-          key: ${{ runner.os }}-modules-${{ hashFiles('**/package-lock.json') }}
-      - name: Install dependencies
-        run: npm install
   lint:
     name: Lint code
     runs-on: ubuntu-latest
-    needs:
-      - update-dependencies
     steps:
       - name: Checkout project
         uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/staging.yml
```diff
@@ -57,13 +37,28 @@ jobs:
...
+        uses: actions/cache@v3
+        with:
+          # Gatsby requires both .cache and public. It will refuse to use .cache/ if public/ is not present.
+          path: |
+            .cache
+            public
+          # Cache will not be used by Gatsby if the following files change:
+          # https://www.gatsbyjs.com/docs/build-caching/
+          # We use the commit SHA on the key to prevent an ever-living cache entry keyed as "main".
+          key: pr-gatsby-${{ hashFiles('package.json', 'gatsby-config.js', 'gatsby-node.js') }}-${{ github.ref }}-${{ github.sha }}
```
imiric, line 58: Doesn't using the commit SHA in the key invalidate the purpose of the cache, since it won't be hit for each new commit?

roobre: Great question! The short answer is that it doesn't, because of `restore-keys`. The long answer follows:

The way `actions/cache` works is by checking first if an entry exactly matching `key` exists. If it does, it restores it, records it as a perfect match, and calls it a day. If it doesn't, it scans through the lines specified in `restore-keys`, and restores the most recent entry that matches as a prefix. In this case, and this is the key part, it does not record a perfect match. Why does it matter? Because at the end of the job, if a perfect match occurred, no new entry is created. However, if there wasn't a perfect match, a new entry is created with the name as specified by `key`.

This may help to make sense of the comment above: If the key did _not_ include a hash, there would only be one entry in the cache named `pr-gatsby-somethingsomething-main`, which would be perfect match for every run. As a result, it would never get refreshed after it gets created. Over time, it would become more and more stale until it reaches a point where it is useless.

By adding the `sha` to `key`, we create a key named like `pr-gatsby-somethingsomething-main-shasha`. There will almost never be a perfect match for this, but as we are also looking for `pr-gatsby-somethingsomething-main-` as a prefix, we will restore the most recent build from main. As it wasn't a perfect match, the modified cache will be uploaded to GH, becoming the new most recent entry.

Hopefully this made sense! LMK if it's still not fully clear.

roobre: An example of this can be seen in the job for #1183, where a perfect hit occurs so the cache is not re-uploaded:
image
https://github.com/grafana/k6-docs/actions/runs/5044385369/jobs/9047751762

imiric: Ah, I missed that `restore-keys` are used as a prefix. Makes sense now. Thanks for explaining :bow:

imiric (approved):
:+1:

roobre:
Merging now, crossing fingers so I didn't make any stupid syntax mistake 🤞🏻

PR #1189, opened 2023-05-22
feat: re-enable xk6-prometheus

szkiba:
Please add the [xk6-prometheus](https://github.com/szkiba/xk6-prometheus) extension. It was added earlier, but has since disappeared from the list (see [this](https://github.com/grafana/k6-docs/blob/fc6f7e24496202f22a9af177a716695c1d3ca397/src/data/ecosystem/extensions.js#L412) from 2021)

xk6-prometheus is a k6 extension implements Prometheus HTTP exporter as k6 output extension.

### DIFF
--- a/src/data/doc-extensions/extensions.json
+++ b/src/data/doc-extensions/extensions.json
@@ -527,6 +527,21 @@
       "cloudEnabled": false
     },
     {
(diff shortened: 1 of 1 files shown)


javaducky (approved):
LGTM!

PR #1224, opened 2023-06-15
Update k6/crypto to recommend the webcrypto API

mstoykov:

PR #1226, opened 2023-06-16
Update `k6/ws` blockquote to refer the experimental module

ppcano:
As per https://github.com/grafana/k6-docs/pull/1224 discussion: 

Screenshot 2023-06-16 at 12 05 18

### DIFF
--- a/src/data/markdown/docs/02 javascript api/12 k6-ws.md
+++ b/src/data/markdown/docs/02 javascript api/12 k6-ws.md
@@ -2,15 +2,8 @@
 title: "k6/ws"
 excerpt: "k6 WebSocket API"
 ---
+<WsBlockquote />
 
-<Blockquote mod='info'>
-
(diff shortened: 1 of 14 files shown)


src/data/markdown/docs/02 javascript api/12 k6-ws.md
```diff
@@ -2,15 +2,8 @@
 title: "k6/ws"
 excerpt: "k6 WebSocket API"
 ---
+<WsBlockquote />
```
mstoykov, line 5: I guess there is no way to tell gatsby - "add this to the top for each page in this folder?" But instead we need to add it to everyone?

mstoykov (approved):
LGTM, although I am not so certain it is worth to edit all the pages.

I guess with at least the blockquote being defined in one place we won't need to go edit it everywhere everytime we want to change something :smile:

PR #1232, opened 2023-06-21
mention headers in canonical form

costela:
This small PR just mentions the fact that `Response.headers` are in canonical form (according to go's definition).
This should avoid mistakes like checking for the inexistent header names.


mstoykov (approved):

ppcano:
@costela, thanks for adding this to the k6 docs.

PR #1242, opened 2023-06-30
Update context, newContext and newPage description for browser module

ankur22:
Due to the change in https://github.com/grafana/xk6-browser/pull/929 and https://github.com/grafana/xk6-browser/pull/945, we need to update the following to reflect this new behaviour where a `browser` can only hold a single `browserContext` and if a user wishes to create a new `browserContext` then they can only close the existing one and create a new one.

- Change `browser.contexts` to `browser.context` to retrieve the current `browserContext`.
- Change the description in `browser.newPage` to help user understand how they can work with more than one `browserContext` sequentially.
- - Change the description in `browser.newContext` to help user understand how they can work with more than one `browserContext` sequentially.


src/data/markdown/docs/02 javascript api/07 k6-experimental/01 browser/context.md
```diff
@@ -0,0 +1,49 @@
...
+    },
+  },
+}
+
+export default function () {
+  console.log(browser.context()); // null
+
+  const page = browser.newPage();
+  const context = browser.context();
+  console.log(context); // {"base_event_emitter":{}}
```
ka3de, line 44: I'm not sure we should reference the `base_event_emitter` in JS API docs example, as it seems a pretty internal element, even thought that is what would actually be printed, but I think It might confuse users. What do you think?

Would it work an example such as:
```suggestion
  console.log(browser.context()); // null

  const page1 = browser.newPage(); // implicitly creates a new browserContext
  const context = browser.context(); // underlying live browserContext associated with browser
  const page2 = context.newPage(); // shares the browserContext with page1
```
Maybe too many comments though.

ka3de (approved):
Overall LGTM :clap: Just made a small comment.

src/data/markdown/docs/02 javascript api/07 k6-experimental/01 browser/context.md
ankur22: Yeah, i was wondering the same. I like your suggestion 👍

mdcruz (approved):

PR #1244, opened 2023-07-02
Update 00 InfluxDB.md

eitama:
The InfluxDB V2 API alllows:
org
orgID
bucket
As parameters for the write API.
K6 uses "org" which is not the orgID.  the documentation is misleading, and this commit fixes that. Also it makes it more explicit that you need the bucket name and not the bucket id.


immavalls:
Thanks for the PR @eitama can you sign the [CLA](https://cla-assistant.io/grafana/k6-docs?pullRequest=1244)? Then we can merge.

eitama:
@immavalls Signed!

PR #1245, opened 2023-07-04
Better dynatrace extra header description

mstoykov:


mstoykov:
This was reverese engineered from the code: 

https://github.com/Dynatrace/xk6-output-dynatrace/blob/93c29c10f8aa05d40b90be45dc1935a84a1f69c3/pkg/dynatracewriter/config.go#L194-L202

and the message is basically the same as the one for [prometheus remote write](https://k6.io/docs/results-output/real-time/prometheus-remote-write/#options)

PR #1249, opened 2023-07-07
Fix example code in browser module context file

ankur22:
This just ensures that the correct `page` is referenced during the cleanup steps.


ka3de (approved):
My bad, my suggestion in previous PR was incomplete!
LGTM. Thanks for reviewing :raised_hands:

PR #1250, opened 2023-07-07
disruptor/examples/http: fix backwards scenario startTimes

roobre:
Current example would start the injeciton scenario immediately after the load scenario finishes, making injection noop.

Additionally I've surrounded the scenarios object with `export const options = {` so it's easier to copypaste.


src/data/markdown/docs/40 xk6-disruptor/04 Examples/02 Inject HTTP faults into Pod.md
```diff
@@ -306,16 +307,15 @@ export const options = {
...
-            duration: "30s",
+            startTime: "10s",
+            duration: "20s",
         },
         disrupt: {
             executor: 'shared-iterations',
             iterations: 1,
             vus: 1,
             exec: "disrupt",
-            startTime: "0s",
```
pablochacin, line 317: Even when it is not necessary, I would prefer to leave this to make it more evident.

pablochacin (commented):
LGTM

pablochacin (approved):

PR #1251, opened 2023-07-09
Create guide for distributed tests

javaducky:
This is not a comprehensive guide for all available options of the operator. We'll be expanding documentation for the operator in a new section at a later time.

### DIFF
--- a/src/data/markdown/translated-guides/en/07 Testing Guides/05 Running distributed tests.md
+++ b/src/data/markdown/translated-guides/en/07 Testing Guides/05 Running distributed tests.md
@@ -1,0 +1,354 @@
+---
+title: 'Running distributed tests'
+excerpt: 'How to run distributed tests in Kubernetes'
+---
+
+It has already been established that k6 can [run large load tests](/testing-guides/running-large-tests/) from a single instance, but what about _multiple instances running a single test_?
+
+Several reasons why you may wish to run a distributed test include:
+
+- Your [system under test](/misc/glossary/#system-under-test) (SUT) should be accessed from multiple IP addresses.
+- A fully optimized node cannot produce the load required by your extremely large test.
+- Kubernetes is already your preferred operations environment.
+
+For scenarios such as these, we've created the [k6-operator](https://github.com/grafana/k6-operator).
+
+## Introducing k6-operator
+[k6-operator](https://github.com/grafana/k6-operator) is an implementation of the [operator pattern](/misc/glossary/#operator-pattern) in Kubernetes, defining [custom resources](/misc/glossary/#custom-resource) in Kubernetes.
+The intent is to automate tasks that a _human operator_ would normally do; tasks like provisioning new application components, changing configurations, or resolving run-time issues.
+
+The k6-operator defines the custom `K6` resource type and listens for changes to, or creation of, `K6` objects.
+Each `K6` object references a k6 test script, configures the environment, and specifies the amount of [parallelism](/misc/glossary/#parallelism) for a test run.
+Once a change is detected, the operator will react by modifying the cluster state, spinning up k6 test jobs as needed.
+Based on the desired parallelism, the operator split the workload between the jobs using [execution segments](/misc/glossary/#execution-segment).
+
+## Get started with k6-operator
+Let's walk through the process for getting started with the k6-operator.
+The only requirement being access to a Kubernetes cluster and having the appropriate access and tooling.
+
+<Blockquote mod="note" title="Experiment in Docker">
+
+This process can be performed on a local Kubernetes cluster running within [Docker](https://docs.docker.com/get-docker/)!
+Using [kind](https://kind.sigs.k8s.io/) or [k3d](https://k3d.io/) are awesome options to experiment with the process.
+
+</Blockquote>
+
+- [1. Install the operator](#install-the-operator)
+- [2. Create a test script](#create-a-test-script)
+- [3. Adding test scripts](#add-test-scripts)
+  - [Add as a ConfigMap](#add-as-a-configmap)
+  - [Add within a PersistentVolume](#add-inside-a-persistentvolume)
+- [4. Create a custom resource](#create-a-custom-resource)
+  - [Script in a ConfigMap](#script-in-a-configmap)
+  - [Script in a PersistentVolume](#script-in-a-persistentvolume)
+  - [Configure the environment](#configure-the-environment)
+  - [Change command-line arguments](#change-command-line-arguments)
+- [5. Run your test](#run-your-test)
+- [6. When things go wrong](#when-things-go-wrong)
+
+## 1. Install the operator
+The first step to running distributed tests in Kubernetes is to install the operator if not already installed in the cluster.
+At this time, installation does require having the project source code downloaded onto your system.
+Installation commands must be run from the source directory.
+
+<Blockquote mod="note" title="Prerequisites">
+
+Besides privileged access to a Kubernetes cluster, installation will require that the system performing the installation has the following tools installed: 
+- [Git](https://git-scm.com/downloads)
+- [Go](https://go.dev/doc/install)
+- [Kubectl](https://kubernetes.io/docs/tasks/tools/#kubectl)
+- [Kustomize](https://kubectl.docs.kubernetes.io/installation/kustomize/)
+- [Make](https://www.gnu.org/software/make/).
+
+</Blockquote>
+
+From your command-line, execute the following:
+```shell
+git clone https://github.com/grafana/k6-operator && cd k6-operator
+```
+Ensure that your `kubectl` tool is set for the appropriate Kubernetes cluster.
+Then, from the `k6-operator` directory, you may now perform the installation:
+```shell
+make deploy
+```
+By default, the operator will be installed into a new namespace, `k6-operator-system`.
+You can verify the successful installation by listing available resources within the namespace, ensuring the status is `Running`:
+```shell
+kubectl get pod -n k6-operator-system
+```
+Once installed, we're ready to start creating and executing test scripts.
+
+## 2. Create a test script
+Creating k6 test scripts for Kubernetes is no different from creating the script for the command-line.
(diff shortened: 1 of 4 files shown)


yorugac (changes requested):
Thanks for putting this together, Paul!

For the "1st iteration" of the guide, I think it's pretty good: straight-forward and precise. I've only left a few small suggestions for improvements.

Overall, most noticeable is, of course, that not all options are being mentioned, e.g. `LocalFile` or `cleanup`. But probably adding everything here will make the guide quite unwieldy and hard to read through so it makes sense to keep it brief. I guess if at some point we could add a "k6-operator reference" page where we can keep a list of all supported options, that'd be a solution. OTOH, current Readme also contains things like "Multi-file tests" with `k6 archive` suggestion: a very common question which is neither a part of API reference nor part of a guide. More like a "howto" or "k6-operator tricks"... Either way, :+1: to have a guide for now :smile:

javaducky:
> we could add a "k6-operator reference" page where we can keep a list of all supported options

Absolutely. My goal is to have an operator section similar to what is available for the [xk6-disruptor](https://mdr-ci.staging.k6.io/docs/refs/pull/1251/merge/javascript-api/xk6-disruptor/). A goal will be to provide architecture information, comprehensive listing of options, and very specific usage examples.

src/data/markdown/translated-guides/en/07 Testing Guides/05 Running distributed tests.md
```diff
@@ -0,0 +1,354 @@
...
+- Kubernetes is already your preferred operations environment.
+
+For scenarios such as these, we've created the [k6-operator](https://github.com/grafana/k6-operator).
+
+## Introducing k6-operator
+[k6-operator](https://github.com/grafana/k6-operator) is an implementation of the [operator pattern](/misc/glossary/#operator-pattern) in Kubernetes, defining [custom resources](/misc/glossary/#custom-resource) in Kubernetes.
+The intent is to automate tasks that a _human operator_ would normally do; tasks like provisioning new application components, changing configurations, or resolving run-time issues.
+
+The k6-operator defines the custom `K6` resource type and listens for changes to, or creation of, `K6` objects.
+Each `K6` object references a k6 test script, configures the environment, and specifies the amount of [parallelism](/misc/glossary/#parallelism) for a test run.
```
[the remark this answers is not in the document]
javaducky: I'll incorporate this but move the glossary link.

src/data/markdown/translated-guides/en/07 Testing Guides/05 Running distributed tests.md
```diff
@@ -0,0 +1,354 @@
...
+Let's create our ConfigMap as `my-test` with the content of our `test.js` script we created in the previous step:
+
+```shell
+kubectl create configmap my-test --from-file test.js
+```
+
+<Blockquote mod="attention" title="">
+
+Limitations exist on how large your test script can be when deployed within a `ConfigMap`.
+Kubernetes imposes a size limit of 1,048,576 bytes (1 MiB) for the data, therefore if your test scripts exceed this limit, you'll need to mount a `PersistentVolume`.
```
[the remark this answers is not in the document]
javaducky: I do have a link to the `ConfigMap` reference in the text just above this warning to drive folks to the reference documentation. 

I am a little apprehensive about adding this particular link to the size limit warning as the anchor is a "when you should consider using a ConfigMap"-type of section. I then thought about adding a link with the text "When should you use a ConfigMap," but then felt we were working too hard to incorporate that particular link to reiterate the limit. For that, I think we should omit the additional link 🤔.

Another thought...should we link to a related GitHub issue? https://github.com/kubernetes/kubernetes/issues/19781#issuecomment-1512262944

WDYT?

javaducky:
@yorugac, I pushed some updates. Let me know what you think.

src/data/markdown/translated-guides/en/07 Testing Guides/05 Running distributed tests.md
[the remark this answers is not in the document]
yorugac: It seems you've ended up adding the k8s link in the next paragraph :smile:
I suggest to resolve this for now :+1: (unless you prefer to change the link to Github issue one after all). I suspect we'll probably need to revise this part when there's a separate section for k6-operator docs and with fresh eyes at some point.

src/data/markdown/translated-guides/en/07 Testing Guides/05 Running distributed tests.md
```diff
@@ -0,0 +1,379 @@
...
+
+</Blockquote>
+
+To learn more about creating `PersistentVolume` resources, review the [Kubernetes documentation](https://kubernetes.io/docs/concepts/storage/persistent-volumes/).
+
+## 4. Create a custom resource
+During [installation](#install-the-operator), the `K6` [Custom Resource](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) was added to the Kubernetes API.
+The data we provide in the custom resource contains all the information necessary for the k6 operator to start a distributed load test.
+
+The main elements defined within the `K6` resource relate to the name and location of the test script to run, and the amount of [parallelism](/misc/glossary/#parallelism) to utilize.
```
yorugac, line 172:
```suggestion
During [installation](#install-the-operator), the `K6` [Custom Resource definition](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) was added to the Kubernetes API.
The data we provide in the custom resource `K6` object should contain all the information necessary for the k6-operator to start a distributed load test.

Specifically, the main elements defined within the `K6` object relate to the name and location of the test script to run, and the amount of [parallelism](/misc/glossary/#parallelism) to utilize.
```
Explanation: here, I thought it'd be nice to clarify what is being installed automatically during "installation of operator" step VS what a user is supposed to define themselves. Otherwise, it sounds as if user is supposed to define CRD itself - which is not true.

src/data/markdown/translated-guides/en/07 Testing Guides/05 Running distributed tests.md
```diff
@@ -0,0 +1,379 @@
...
+
+<Blockquote mod="attention" title="">
+
+It is important that the `ConfigMap` and `CustomResource` are created in the same [Namespace](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/).
+
+</Blockquote>
+
+### Script in a PersistentVolume
+If the test script to be executed is contained within a `PersistentVolume`, creation of a [PersistentVolumeClaim](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims) will be required.
+We won't go into the details of PersistentVolumes and PersistentVolumeClaims, but to learn more, you should review the [Kubernetes documentation](https://kubernetes.io/docs/concepts/storage/persistent-volumes/).
```
yorugac, line 221: Shouldn't this part be in the [Add inside a PersistentVolume](#add-inside-a-persistentvolume) subsection, above? When I read through, it seems like a repeat of info in that subsection but more precisely formulated...

src/data/markdown/translated-guides/en/07 Testing Guides/05 Running distributed tests.md
```diff
@@ -0,0 +1,379 @@
...
+spec:
+  parallelism: 4
+  # Removes all resources upon completion
+  cleanup: post
+  script:
+    configMap:
+      name: my-test
+      file: test.js
+  arguments:  -o experimental-prometheus-rw
+```
```
yorugac, line 362: This example is great, thanks! Btw, it's the first time we have it documented: its absent even in Readme ATM :smile:

yorugac (changes requested):
@javaducky, after more reading I've found a couple of typos and possible improvement suggestions - please see the comments. The most funny part is likely the "Custom Resource" terminology: it seems to me that even Kubernetes docs can be seen as confusing on the topic...

src/data/markdown/translated-guides/en/07 Testing Guides/05 Running distributed tests.md
```diff
@@ -0,0 +1,379 @@
...
+Ensure that your `kubectl` tool is set for the appropriate Kubernetes cluster.
+Then, from the `k6-operator` directory, you may now perform the installation:
+```shell
+make deploy
+```
+By default, the operator will be installed into a new namespace, `k6-operator-system`.
+You can verify the successful installation by listing available resources within the namespace, ensuring the status is `Running`:
+```shell
+kubectl get pod -n k6-operator-system
```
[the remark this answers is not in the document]
javaducky: I do like showing the output of some commands within the code blocks, however, readers are able to use the copy button from the code block, which would then copy the result as well. I can create a second code block with the result while removing the copy button from the UI.

src/data/markdown/translated-guides/en/07 Testing Guides/05 Running distributed tests.md
javaducky: > It seems you've ended up adding the k8s link in the next paragraph 😄

😄 yes...I continued to work with it after the initial reply.

src/data/markdown/translated-guides/en/07 Testing Guides/05 Running distributed tests.md
javaducky: I'll include mention of `PersistentVolumeClaim` when accessing data on a `PersistentVolume`.

I didn't originally include the need for PVC in the earlier section because that particular section strictly deals with _adding_ your test scripts versus _using_ your test scripts. Since the latter section requires the need for PVC, I add the more precise usage.

I've played around with the text a bit for the former section...let me know what you think.

javaducky:
@yorugac, here are some more updates as we discussed.

yorugac (approved):
Subsections about PVC flow better now, IMHO. Though I wonder if we should pull volumes out completely as a separate guide some time and keep this guide simple, with only one "path of execution" via configmaps...

Overall, LGTM :raised_hands: Thanks for the updates!

PR #1260, opened 2023-07-19
Fix Public Key

Breee:


codebien:
Hi @Breee,
if I'm getting right your suggestion here the attempt is to use the long `keyid` instead of using the `fingerprint` that we are providing. The expectation here is that most of the people will just copy and paste this command so having a longer string shouldn't be a real issue. Instead, using the keyid will raise the chance of a collision (these days but especially in the future).

For this reason, I close this pull request. Feel free to re-open or provide details in a comment if you feel that we should still address a specific issue.

PR #1264, opened 2023-07-21
Document restrictions with port forwarding

pablochacin:


src/data/markdown/docs/40 xk6-disruptor/01 Get started/04 Expose  Your Application.md
```diff
@@ -7,19 +7,49 @@ To access your application from the test scripts, you must assign it an external
...
-</CodeGroup>
+### Limitations using port forwarding
+
+To be able to inject faults, `xk6-disruptor` must [install an agent on each target](/javascript-api/xk6-disruptor/explanations/how-xk6-disruptor-works) that intercepts the requests and applies the desired disruptions. This process requires any existing connection to the targets to be redirected to the agent.
+
+Due to a existing bug in `kubectl`, the process of installing the disruptor can potentially [break the port forwarding](https://github.com/grafana/xk6-disruptor/issues/254). Notice that this issue arises only if the faults are injected in the service that is exposed using port forward. If the faults are injected in another service not exposed by port-forwarding, there should not be any issue.
+
+Until this issue is solved in `kubectl`, tests using port forwarding to access a service should ensure the agent is installed in the targets before any traffic is sent by the test.
+
+The simplest way to accomplish this is to ensure the scenario that executes the load (2) starts after the scenario that injects the faults (1):
```
roobre, line 30: Another way to work around the current issue could be to run the `port-forward` command in a loop. I'm not a huge fan of proposing it, but it does not require changes on the scripts, which I think is less intrusive. WDYT?

pablochacin: I  wouldn't recommend that approach, as many requests will fail while the loop completes, distorting the results of the test.

roobre (approved):

src/data/markdown/docs/40 xk6-disruptor/01 Get started/04 Expose  Your Application.md
```diff
@@ -7,19 +7,49 @@ To access your application from the test scripts, you must assign it an external
...
+kubectl port-forward svc/my-service 8080:80
+Forwarding from 127.0.0.1:8080 -> 80
 ```
 
-</CodeGroup>
+### Limitations using port forwarding
+
+To be able to inject faults, `xk6-disruptor` must [install an agent on each target](/javascript-api/xk6-disruptor/explanations/how-xk6-disruptor-works) that intercepts the requests and applies the desired disruptions. This process requires any existing connection to the targets to be redirected to the agent.
+
+Due to a existing bug in `kubectl`, the process of installing the disruptor can potentially [break the port forwarding](https://github.com/grafana/xk6-disruptor/issues/254). Notice that this issue arises only if the faults are injected in the service that is exposed using port forward. If the faults are injected in another service not exposed by port-forwarding, there should not be any issue.
```
heitortsergent, line 26:
```suggestion
Due to a existing bug in `kubectl`, the process of installing the disruptor can potentially [break the port forwarding](https://github.com/grafana/xk6-disruptor/issues/254). Notice that this issue happens only if the faults are injected in the service that is exposed using port forward. If the faults are injected in another service not exposed by port-forwarding, there shouldn't be any issue.
```

nit: just a couple of small suggestions here.

src/data/markdown/docs/40 xk6-disruptor/01 Get started/04 Expose  Your Application.md
```diff
@@ -43,6 +73,20 @@ kubectl -n <name space> patch svc <service name> -p '{\"spec\": {\"type\": \"Loa
 </CodeGroup>
 
 
+You can retrieve the external IP address and store in an environment variable (`SVC_IP` in this example) using the following command:
```
heitortsergent, line 76:
```suggestion
You can retrieve the external IP address and store it in an environment variable (`SVC_IP` in this example) using the following command:
```

heitortsergent (approved):
lgtm!

PR #1267, opened 2023-07-24
Add my xk6-read extension to extensions registry.

acuenca-facephi:
xk6-read is an extension to read files and directories. 
I created it because in my use case I need to use thousands of different images during test execution, and k6 only can read files in the init stage. This force me to load all the images using a JSON file with an array of file paths and load in memory ALL the images.

### DIFF
--- a/src/data/doc-extensions/extensions.json
+++ b/src/data/doc-extensions/extensions.json
@@ -1081,6 +1081,21 @@
       "categories": ["Misc"],
       "tiers": ["Community"],
       "cloudEnabled": false
+    },
+    {
+      "name": "xk6-read",
+      "description": "Read files and directories",
+      "url": "https://github.com/acuenca-facephi/xk6-read",
+      "logo": "",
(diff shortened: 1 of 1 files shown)


javaducky:
Hello @acuenca-facephi ! Thank you for your contribution.

After checking into the project, I was curious about the particular license selection of CC0-1.0. Checking into this particular type, there is some concern given that the Open Source Initiative still needs to approve the license. Would it be acceptable to change the license to Apache 2.0?

acuenca-facephi:
> Hello @acuenca-facephi ! Thank you for your contribution.
> 
> After checking into the project, I was curious about the particular license selection of CC0-1.0. Checking into this particular type, there is some concern given that the Open Source Initiative still needs to approve the license. Would it be acceptable to change the license to Apache 2.0?

No problem, I update the license to Apache 2.0.

javaducky (approved):
LGTM! Thank you so much!

PR #1278, opened 2023-08-02
Fix Load Testing snippet comment

nmpcunha:
It seems that there is an inconsistency in some snippets in terms of code and then the comments there.
Hope I got it right 😉


ppcano:
You got it right. Thanks so much for opening the PR to correct this comment!

PR #1288, opened 2023-08-08
fix: remove unnecessary sleep import and blank line

SpiffyEight77:
According to the document note, there is no need to use a sleep() function at the end of the VU code.

> Don't put sleep at the end of an iteration.
The arrival-rate executors already pace the iteration rate through the rate and timeUnit properties. So it's unnecessary to use a sleep() function at the end of the VU code.


heitortsergent (approved):
Thanks for catching that and opening a PR. 🙇

PR #1305, opened 2023-08-21
Add xk6-input-prometheus

JorTurFer:
Hi,
I have created an extension for reading metrics from prometheus. There are use cases where direct metrics aren't available during the tests and the tests case has to be evaluated thanks to indirect measurements.

Our use case for this extension is to measure [KEDA](keda.sh) lag during the load. To measure the operator processing lag, we can't use any direct metric from anywhere because things happen internally, but thanks to this extension, we can export the operator metrics to Grafana Cloud, and query the metrics from there during the test. This brings us the option to use thresholds using custom gauges populated from prometheus queries.

The repo isn't public yet but it'll be public soon (when we finish some pending stuff), so there you will be able to see how the extension works

### DIFF
--- a/src/data/doc-extensions/extensions.json
+++ b/src/data/doc-extensions/extensions.json
@@ -1141,6 +1141,21 @@
       "categories": ["Observability"],
       "tiers": ["Community"],
       "cloudEnabled": false
+    },
+    {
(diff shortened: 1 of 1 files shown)


javaducky (approved):
LGTM! Thank you!

PR #1310, opened 2023-08-31
Fix the browser examples which work with check

ankur22:
`check` expects a function that returns a `boolean`. There is enough of a difference in what the k6 type definitions expect vs how the example k6 browser tests work with `check`. If a user works with the example browser tests as documented along with the type definitions then they will find an error in the example code. This commit fix those examples so that users will not see that error.

Before:
Screenshot 2023-08-31 at 14 48 04

### DIFF
--- a/src/data/markdown/docs/02 javascript api/07 k6-experimental/01 browser/04-element-handle.md
+++ b/src/data/markdown/docs/02 javascript api/07 k6-experimental/01 browser/04-element-handle.md
@@ -82,7 +82,7 @@ export default async function () {
 
     await Promise.all([page.waitForNavigation(), submitButton.click()]);
     check(page, {
-      'header': page => page.locator('h2').textContent() == 'Welcome, admin!',
+      'header': p => p.locator('h2').textContent() == 'Welcome, admin!',
     });
   } finally {
     page.close();
--- a/src/data/markdown/docs/02 javascript api/07 k6-experimental/01 browser/04-element-handle.md
+++ b/src/data/markdown/docs/02 javascript api/07 k6-experimental/01 browser/04-element-handle.md
@@ -82,7 +82,7 @@ export default async function () {
 
     await Promise.all([page.waitForNavigation(), submitButton.click()]);
     check(page, {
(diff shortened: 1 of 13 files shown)


inancgumus (approved):
LGTM 👍 

Here's the same suggestion from [the other PR](https://github.com/grafana/xk6-browser/pull/1022#discussion_r1312626108):

> WDYT about saying just p to distinguish from the actual page object?

src/data/markdown/docs/02 javascript api/07 k6-experimental/01 browser/04-element-handle.md
```diff
@@ -82,7 +82,7 @@ export default async function () {
 
     await Promise.all([page.waitForNavigation(), submitButton.click()]);
     check(page, {
-      header: page.locator('h2').textContent() == 'Welcome, admin!',
+      'header': page => page.locator('h2').textContent() == 'Welcome, admin!',
```
inancgumus, line 85: Asking because of curiosity: I wonder why do we need to make the `header` property to `'header'`?

ankur22: Only so that it matches what is in the [docs](https://k6.io/docs/javascript-api/k6/check/). Happy to change it to `header`, what's the "correct" way?

inancgumus: I looked into it and discovered the correct way is 'header' :)

ankur22: Ah, ok 😅 I just changed it, but I'll change it back 👍

ankur22: I've resolved the comment on `page` to `p` in a105c8c8c36d205770443d992c7cf2b58d5db5a7. I have left `'header'` as `header` 👍

ka3de (approved):
LGTM.

PR #1320, opened 2023-09-11
fix: replace grpbc.in links with grpcbin.test.k6.io

heitortsergent:


ppcano:
@heitortsergent 

It is being a while since these examples were written and we don't verify whether theses examples work. It would be good if you can test if the current examples still work. 

Also, feel free to add any relevant example on the test folder at https://github.com/grafana/k6-grpcbin

LGTM, merge whenever you find convenient.

heitortsergent:
@ppcano that's a good point. I tested the example at https://k6.io/docs/javascript-api/k6-net-grpc/, and it's working, but it also wasn't clear where to get the definitions...

I'm going to merge this for now just to fix the link, and I'll open an issue to review the examples.

PR #1331, opened 2023-09-20
Update metric name restrictions

mstoykov:
Changes in code done in https://github.com/grafana/k6/pull/3335

### DIFF
--- a/src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
+++ b/src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
@@ -109,9 +109,9 @@ For details of all metrics, refer to the [Metrics reference](/using-k6/metrics/r
 
 ## Metric name restrictions
 
-Metric names must comply with OpenTelemetry and [Prometheus limitations](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels).
+Metric names must comply with OpenTelemetry and [Prometheus limitations](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels). The character limit is the same limit k6 had before restricting the character set down.
 
-That means they must not start with a number and a metric name can be 1 to 63 symbols of:
+That means they must not start with a number and a metric name can be 1 to 128 symbols of:
 1. any Unicode Letters
 2. any Unicode Number
 3. `_` (an underscore)


codebien (approved):

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
```diff
@@ -109,9 +109,9 @@ For details of all metrics, refer to the [Metrics reference](/using-k6/metrics/r
 
 ## Metric name restrictions
 
-Metric names must comply with OpenTelemetry and [Prometheus limitations](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels).
+Metric names must comply with OpenTelemetry and [Prometheus limitations](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels). The character limit is the same limit k6 had before restricting the character set down.
```
heitortsergent, line 112: I think we can remove this sentence. If someone is new to k6 and reads this, it doesn't really give them any meaningful information and can just make this more confusing. If they are already familiar with k6, they might also not remember the limit.

heitortsergent (changes requested):
Left one comment for a change I think we should make. Also, is this change available right away, or is it part of 0.47?

src/data/markdown/translated-guides/en/02 Using k6/02 Metrics.md
mstoykov: I think it is better if it stays - as it explains the reason for the specific limit.

heitortsergent: I think it's letting the reader know there were changes to the character limit, but it doesn't really give any details on why this has changed, unless we add more information. And I feel like that could quickly become irrelevant in the future...

Is this important for users if they're upgrading from previous versions to 0.47? If it is, we could also switch this to a `<blockquote mod='attention'>`.

mstoykov: > know there were changes to the character limit

There no loonger *is* a change.

We currently limit to 128, we started giving warning for characters outside of what Otel and Promehtesu support and found out that Otel has a limit of 63, so we added that to the new warning as well.

Now we foud out the 63 limit for Otel is lifted. So the warning got 128 character as the limti - which is also the current *error* limit. 

So in practice the 128 character limit isn't *new* it is *very* old, we just didn't end up adding a new one 63 character limti

Issue #1302, opened 2023-08-17
Further explain vusActive vs vusInitialized in the docs


immavalls:
As reported in the community slack, the docs https://k6.io/docs/javascript-api/k6-execution/#instance don't specify the difference between `vusActive` and `vusInitialized`. It would be good to add an explanation.

> If you have a scenario with 10 VUs and a ramp-up duration of 10 seconds, then at the beginning of the test, `vusInitialized` will be 10 and `vusActive` will be 0. After 5 seconds, `vusActive` will be 5 and `vusInitialized` will still be 10. After 10 seconds, both `vusActive` and `vusInitialized` will be 10

<img width="673" alt="image" src="https://github.com/grafana/k6-docs/assets/4788634/a2d3aaaa-f730-441f-8043-93de0cf6fa53">

heitortsergent:
I think the Constant arrival rate page definitely helps, but it'd still be helpful to have something in the execution page.

As far as the explanation on the Constant arrival rate, I think it doesn't mention vusInitialized, but my guess is that number would be `50` since that's what's set on the example script. But as the test runs, it only needs up to `17` VUs to reach the iteration rate, and that would be the `vusActive` value at that point in time.

PR #1337, opened 2023-09-22
fix change vusInitalized example

zimmj:
mention that the initialised VUs are ready or running, as they finished setup call, which is only run ones per creation of a VU.

Fixes grafana/k6-docs/issues#1302
https://github.com/grafana/k6-docs/issues/1302

### DIFF
--- a/src/data/markdown/docs/02 javascript api/06 k6-execution.md
+++ b/src/data/markdown/docs/02 javascript api/06 k6-execution.md
@@ -48,7 +48,7 @@ The instance object provides information associated with the load generator inst
 | iterationsInterrupted  | integer | The number of prematurely interrupted iterations in the current instance. |
 | iterationsCompleted    | integer | The number of completed iterations in the current instance. |
 | vusActive              | integer | The number of active VUs. |
-| vusInitialized         | integer | The number of currently initialized VUs. |
+| vusInitialized         | integer | The number of VUs which finished setup and are ready or running. If more VUs are required, more will be Initialized. |
 | currentTestRunDuration | float   | The time passed from the start of the current test run in milliseconds. |
 
 ### scenario


heitortsergent:
Hi @zimmj! Thanks for your contribution! I took a look at the issue, and I think adding the example from the issue would probably be a more helpful explanation than updating the parameter description. Would you mind making that change?

zimmj:
I wrote something on the issue.
I think it needs more clarification as we talk about instances, and it could have more than one instance in a test run?

src/data/markdown/docs/02 javascript api/06 k6-execution.md
```diff
@@ -48,7 +48,7 @@ The instance object provides information associated with the load generator inst
 | iterationsInterrupted  | integer | The number of prematurely interrupted iterations in the current instance. |
 | iterationsCompleted    | integer | The number of completed iterations in the current instance. |
 | vusActive              | integer | The number of active VUs. |
-| vusInitialized         | integer | The number of currently initialized VUs. |
+| vusInitialized         | integer | The number of VUs which finished setup and are ready or running. If more VUs are required, more will be Initialized. |
```
heitortsergent, line 51:
```suggestion
| vusInitialized         | integer | The number of currently initialized VUs. |
```

src/data/markdown/docs/02 javascript api/06 k6-execution.md
```diff
@@ -48,7 +48,7 @@ The instance object provides information associated with the load generator inst
...
 | currentTestRunDuration | float   | The time passed from the start of the current test run in milliseconds. |
```
heitortsergent, line 53:
```suggestion

`vusInitialized` vs `vusActive`

The `vusInitialized` and `vusActive` might seem similar, but they are different properties and their values might vary during a test run. For example, if you have a scenario with 10 VUs and a ramp-up duration of 10 seconds, then at the beginning of the test, `vusInitialized` will be 10 and `vusActive` will be 0. After 5 seconds, `vusActive` will be 5 and `vusInitialized` will still be 10. After 10 seconds, both `vusActive and `vusInitialized` will be 10.

This is a simple example, but their values can also vary depending on what type of scenario you're using, and your test script.

```

heitortsergent (commented):
@zimmj I left a couple comments with some suggestions for how we can add the information to this page. :)

PR #1338, opened 2023-09-22
Update options config file section to not be collapsible

heitortsergent:
Just a small update to make it clearer there's a default configuration file, and update the config file section to not be collapsible so it displays like the other sections.

I struggled with figuring out where to find the configuration file when I was trying to configure a default projectID, so hopefully this small change makes it easier for the next person.


mdcruz (approved):
LGTM! 🚀

PR #1348, opened 2023-09-28
Document the aws jslib's EventBridge support

oleiade:
This Pull Request documents the [newly introduced support for the AWS EventBridge service](https://github.com/grafana/k6-jslib-aws/pull/55).

⚠️ this PR should be merged only after the jslib-aws's version 0.10 was released ⚠️

### DIFF
--- a/src/data/markdown/docs/20 jslib/01 jslib/01 aws/00 EventBridge.md
+++ b/src/data/markdown/docs/20 jslib/01 jslib/01 aws/00 EventBridge.md
@@ -1,0 +1,65 @@
+---
+title: 'EventBridgeClient'
+head_title: 'EventBridgeClient'
+description: 'EventBridgeClient allows interacting with AWS EventBridge service'
+excerpt: 'EventBridgeClient class allows sending custom events to Amazon EventBridge so that they can be matched to rules.'
+---
+
+`EventBridgeClient` interacts with the AWS EventBridge service.
+
+With it, you can send custom events to Amazon EventBridge. These events can then be matched to rules defined in EventBridge. For a full list of supported operations, see [Methods](#methods).
+
+Both the dedicated `event-bridge.js` jslib bundle and the all-encompassing `aws.js` bundle include the `EventBridgeClient`.
+
+### Methods
+
+| Function                                                                                       | Description                                        |
+| :--------------------------------------------------------------------------------------------- | :------------------------------------------------- |
+| [putEvents(input)](/javascript-api/jslib/aws/eventbridgeclient/eventbridgeclient-putevents/) | Send custom events to Amazon EventBridge |
+
+### Throws
+
+EventBridgeClient methods will throw errors in case of failure.
+
+| Error                    | Condition                                                                   |
+| :----------------------- | :-------------------------------------------------------------------------- |
+| InvalidSignatureError    | when invalid credentials were provided or the request signature is invalid. |
+| EventBridgeServiceError  | when AWS replied to the requested operation with an error.                  |
+
+### Examples
+
+<CodeGroup labels={[]}>
+
+```javascript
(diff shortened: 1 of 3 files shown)


src/data/markdown/docs/20 jslib/01 jslib/01 aws/00 EventBridge.md
```diff
@@ -0,0 +1,65 @@
...
+
+With it, you can send custom events to Amazon EventBridge. These events can then be matched to rules defined in EventBridge. For a full list of supported operations, see [Methods](#methods).
+
+Both the dedicated `event-bridge.js` jslib bundle and the all-encompassing `aws.js` bundle include the `EventBridgeClient`.
+
+### Methods
+
+| Function                                                                                       | Description                                        |
+| :--------------------------------------------------------------------------------------------- | :------------------------------------------------- |
+| [putEvents(input)](/javascript-api/jslib/aws/eventbridgeclient/eventbridgeclient-putevents/) | Send custom events to Amazon EventBridge |
```
heitortsergent, line 18:
```suggestion
| [putEvents(input)](/javascript-api/jslib/aws/eventbridgeclient/eventbridgeclient-putevents/) | Send custom events to Amazon EventBridge. |
```

nit 🕵️

src/data/markdown/docs/20 jslib/01 jslib/01 aws/EventBridgeClient/putEvents.md
```diff
@@ -0,0 +1,70 @@
...
+excerpt: 'EventBridgeClient.putEvents sends custom events to Amazon EventBridge'
+---
+
+`EventBridgeClient.putEvents` sends custom events to Amazon EventBridge so that they can be matched to rules.
+
+### Parameters
+
+| Parameter     | Type            | Description                                                                                                              |
+| :------------ | :-------------- | :----------------------------------------------------------------------------------------------------------------------- |
+| input         | [PutEventsInput](#puteventsinput) | An array of objects representing events to be submitted. |
```
heitortsergent, line 13: I'm looking at both `input` and `Entries` with the same description, and I'm wondering if this would be more accurate if it said: "An object with the key `Entries` and the value of an array of EventBridgeEntry, representing events to be submitted.", or something along these lines.

Wdyt? Or is this more confusing?

heitortsergent (approved):
Left a comment and a very minor nit, overall LGTM!

One thing, can you please rename the `00 EventBridge.md` file to `00 EventBridgeClient.md`? It looks like that would fix issue in the preview of the pages being in two different sections.

src/data/markdown/docs/20 jslib/01 jslib/01 aws/EventBridgeClient/putEvents.md
oleiade: I see what you mean, but the `PutEventsInput` is a data interface defined by AWS, which we also define in the code, and which also has other attributes such as `EndpointId` (and I believe more that we haven't included yet).

Thus, I believe it is better kept this way.

However, I'm not happy with how we deal with these nested type definitions in the docs at the moment. The problem is that if the argument to a function is an object with a strict form and shape that we expect, it's hard to describe it to the user in a concise way. Having multiple tables, as I did here, was my best bet. We could also have dedicated separate pages, but that makes documenting a library like AWS very cumbersome.

What do you think? Maybe you see an alternative that would be more friendly to both the user and the documentation writer? 😄 🙇🏻

heitortsergent: That's a great point... And I think tables are an ok solution for right now, feel free to merge this PR.

I'll open a separate issue so we can discuss/test some different ways of how to better display nested objects, maybe we can look at some other docs and get some ideas. :)

PR #1356, opened 2023-10-03
Reorganize disruptor get-started

pablochacin:
Make the get started section more to the point and actionable.
Move the introduction to a new "About" page


ppcano:
@pablochacin, 

1 - should this link go to the `About` or `First steps` page?

Screenshot 2023-10-03 at 22 56 26

2- Also, first-steps URL is 404
https://mdr-ci.staging.k6.io/docs/refs/pull/1356/merge/javascript-api/xk6-disruptor/get-started/first-steps/

pablochacin:
@ppcano 

1. The main purpose of the PR is to make the "landing" page of the disruptor API more "actionable" than descriptive. Therefore I think it should point to `First Steps` instead of `About`. Please notice that `About` is basically the same previous `Welcome` page that was the entry page for the disruptor. 

2. The page is there and it is possible to navigate to ti from the side panel and the drop-down menu. Any idea how to fix this? I'm not sure how this navigation works.

roobre:
I'm not 100% convinced about how the new landing page helps avoid making the user jump around. I think another possible way to address the feedback would be to create an additional, less detailed end to end journey (we could call that "quick start", or "get started"). This page would not include any description (I think it is well captured in the "Welcome" page), and instead I'd put on it all the steps that are required to get a first test running. Something like:
1. Requirements: Command to check you have a valid kubeconfig, maybe come `kubectl auth can-i`
2. Installation: Quickest way to get the binary, maybe a curl/wget to the latest release
3. Expose application: Simplest possible way, probably port-forward
4. Run the test: Including a ready-to-use snippet.
5. Learn more: Links to the API

Each of these sections could have a link to their expanded, own-page version.

WDYT about this approach?

pablochacin:
> I'm not 100% convinced about how the new landing page helps avoid making the user jump around.

I agree. For me this is a first step, reorganizing the content that lands in a list of first steps.

 The changes you suggest could be done in a follow-up PR, replacing the current first steps.

ppcano:
I didn't remember this but it seems the issue is that redirects don't work with the `mdr-ci.staging.k6.io`.

```bash
curl -I https://mdr-ci.staging.k6.io/docs/refs/pull/1356/merge/es/guias-de-prueba/automatizacion-de-pruebas-de-rendimiento/ HTTP/2 404

curl -I https://staging.k6.io/docs/es/guias-de-prueba/automatizacion-de-pruebas-de-rendimiento/ HTTP/2 301 
```

So the redirect should work when deployed to `staging` or `production`.

Also, changed https://github.com/grafana/k6-docs/pull/1356/commits/fa09ab9031211d091876c8b28af888b42dfe13f6 which relates to https://k6.io/docs/get-started/welcome/

PR #1376, opened 2023-10-20
Add browser.closeContext() docs

ka3de:
Adds the associated documentation for the new `browser.closeContext()` method, which is a handier option to close the currently active browser context instead of the more convoluted `browser.context().close()`.


inancgumus (approved):
LGTM

PR #1381, opened 2023-10-25
Document Pod termination fault

pablochacin:


roobre (approved):
Left small suggestions, mostly opinionated. Feel free to disregard them if you're not convinced.

roobre (approved):
🎉

PR #1389, opened 2023-10-31
Update 00 NewRelic.md

mmllc-jsilverman:
Without removing the restart policy option, docker run gives error: "docker: Conflicting options: --restart and --rm"


heitortsergent:
Hi @mmllc-jsilverman! Thanks for the PR, and sorry for the delay in reviewing it.

You're right that using those two flags gives an error, would you mind signing the CLA so I can go ahead and run the checks and then merge this? 🙏

PR #1385, opened 2023-10-29
docs: fix small typos

gonmmarques:

PR #1336, opened 2023-09-22
Add example for azure b2c OAuth

zimmj:
This example allows fetching tokens for a given user from azure B2C.

I just started to write a load test for a project I am working on, and needed exactly this kind of OAuth scenario.

Luckily, I found an example in JMeter and took the time to rewrite it in JavaScript to have it run with k6.
As I saw, that already some other people ask for it here, I would love to share this example to get token from Azure B2C.
To help someone on the way.
I think this is the best place to do so.

Thanks for the awesome tool.

PR #1393, opened 2023-11-01
Add PRs from last week to Grafana docs

heitortsergent:
Bring in the changes we've merged to k6-docs in the last week to the Grafana docs version.

#1385, #1336, #1381.

### DIFF
--- a/docs/sources/v0.47.x/examples/oauth-authentication.md
+++ b/docs/sources/v0.47.x/examples/oauth-authentication.md
@@ -61,6 +61,207 @@ export function authenticateUsingAzure(tenantId, clientId, clientSecret, scope, 
 
 {{< /code >}}
 
+### Azure B2C
+
+The following example shows how you can authenticate with Azure B2C using the [Client Credentials Flow](https://docs.microsoft.com/en-us/azure/active-directory-b2c/active-directory-b2c-reference-oauth-code#client-credentials-flow).
+
+This example is based on a JMeter example found at the [azure-ad-b2c/load-tests](https://github.com/azure-ad-b2c/load-tests) repository.
+
+To use this script, you need to:
+
+1. [Set up your own Azure B2C tenant](https://learn.microsoft.com/en-us/azure/active-directory-b2c/tutorial-create-tenant)
+    * Copy the tenant name, it will be used in your test script.
+1. [Register a web application](https://learn.microsoft.com/en-us/azure/active-directory-b2c/tutorial-register-applications?tabs=app-reg-ga)
+    * Register a single page application with the redirect URL of: https://jwt.ms. That's needed for the flow to receive a token.
+    * After the creation, you can get the Application (client) ID, and the Directory (tenant) ID. Copy both of them, they'll be used in your test script.
+1. [Create a user flow so that you can sign up and create a user](https://docs.microsoft.com/en-us/azure/active-directory-b2c/tutorial-create-user-flows)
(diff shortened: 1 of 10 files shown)


heitortsergent:
@ppcano @pablochacin tagging you both here cause there were updates to the xk6-disruptor docs that we merged to k6-docs last week. I tried to bring them over in this PR, if you want to review them and make sure everything looks good. :)

docs/sources/v0.47.x/examples/oauth-authentication.md
jdbaldry, line 1: Looks like I need to add Oauth and JMeter to the Vale dictionary: https://github.com/grafana/writers-toolkit/pull/366.

docs/sources/v0.47.x/javascript-api/xk6-disruptor/faults/pod-termination.md
```diff
@@ -0,0 +1,33 @@
...
+
+# Pod Termination
+
+A Pod Termination Fault allows terminating either a fixed number or a percentage of the pods that matching a selector or back a service.
+
+A Pod Termination fault is defined by the following attributes:
+
+| Attribute     | Type   | Description |
+| ------------- | ------ | --------|
+| count         | integer or percentage | the number of pods to be terminated. It can be specified as a integer number or as a percentage (e.g. `30%`) that defines the fraction of target pods to be terminated|
```
jdbaldry, line 15: > pod-termination.md    15 136 error    Google.Latin Use 'for example' instead of 'e.g.'. (vale-error)

```suggestion
| count         | integer or percentage | the number of pods to be terminated. It can be specified as a integer number or as a percentage, for example `30%`, that defines the fraction of target pods to be terminated|
```

docs/sources/v0.47.x/results-output/end-of-test/custom-summary.md
jdbaldry, line 1: I've also added JUnit to the dictionary from this page: https://github.com/grafana/writers-toolkit/pull/366.

docs/sources/v0.47.x/using-k6/scenarios/executors/shared-iterations.md
jdbaldry, line 1: And VU: https://github.com/grafana/writers-toolkit/pull/366.

It looks like it is commonly misspelt  as 'vu' in this file but that's outside of the scope of this PR.
Perhaps lowercase is also an acceptable spelling though?

jdbaldry (approved):

docs/sources/v0.47.x/using-k6/scenarios/executors/shared-iterations.md
heitortsergent: I think we use lowercase 'vu' when we're referring to the property name in scripts, since that's how it's supposed to be used. It would probably be good to add it as an acceptable spelling as well. 🤓

jdbaldry: Ah nice. And that shouldn't be wrapped in backticks to make it code?

Vale doesn't spellcheck code so we wouldn't need the dictionary entry in that case.

heitortsergent: Ah, that's a good point. I updated the one instance of `vus` in the paragraph to use backticks, the other one is in a table and I think we should keep that as regular text... I think most of the other tables where we have properties listed they are just regular text, not code.

PR #1400, opened 2023-11-07
Clean artifact

olegbespalov:
# What?

It's removing an artifact that is presented on some pages.

image

# Why?

It's probably left during the migration.


heitortsergent:
Thanks for catching that @olegbespalov! 🙇

PR #1402, opened 2023-11-08
Add "next" version folder

heitortsergent:
Add "next" version folder so we can use it for the next k6 release.


jdbaldry (approved):
LGTM assuming it's a copy-paste of v0.47.x.

Are we OK with publishing next? It's consistent with other projects but just wanted to double check.

oleiade (approved):
LGTM 🚀 

@jdbaldry Let's assume it's okay until we're told otherwise 🤝 (@heitortsergent we should probably make it explicit in the README somehow, so that people are aware that changes made in next will be available, albeit not referenced, online)

PR #1406, opened 2023-11-09
Grafana docs: fix links

### LINKED ISSUE #1396: Fix broken links
With the Grafana docs migration, we tried to convert and test as many links as we could with the new structure, but some links might still be broken.

We should do an audit of the links on all the pages to make sure everything is working correctly and fix any broken links we find, especially on the JavaScript API section.

heitortsergent:
Closes #1396.

Fix any remaining links that still have the `(/`format (instead of using the full Grafana URL).


heitortsergent:
@jdbaldry I added you as a reviewer but this PR is obviously too big. 😅  But if you wanted to take a quick look, I think I got all the remaining links that we missed during the migration... It would be great if we could run some sort of link crawler to see if there's anything that might have slipped through or might just have the wrong URL.

jdbaldry:
I agree that an ad hoc link crawler would be great.
I don't have anything to hand right now but have opened https://github.com/grafana/technical-documentation/issues/845 to track it

jdbaldry (approved):
At a scan this seems to fix a significant number of links without any negative side effects.

Thank you so much.

I can request another site crawl from Anne (maybe just even for this section of docs) to see what links are still left after this is merged.

heitortsergent:
Ah yes, a site crawl would be great! And thanks for reviewing it 🙇

PR #1415, opened 2023-11-14
Add page.throttleCPU and page.throttleNetwork docs

ankur22:
# What

This adds the documentation for `page.throttleCPU` and `page.throttleNetwork` which are both new APIs which are part of the browser module.

# Why

To inform users how to work with these new APIs.


docs/sources/next/javascript-api/k6-experimental/browser/page/throttlenetwork.md
```diff
@@ -0,0 +1,59 @@
...
+# throttleNetwork(networkProfile)
+
+Throttles the network in Chrome/Chromium to slow it down by the specified fields in the `networkProfile` object.
+
+| Parameter               | Type           | Default | Description                                                                            |
+|-------------------------|----------------|---------|----------------------------------------------------------------------------------------|
+| networkProfile          | NetworkProfile | `null`  | This is a mandatory parameter.                                                         |
+| networkProfile.latency  | number         | `0`     | Minimum latency from request sent to response headers received (ms).                   |
+| networkProfile.download | number         | `-1`    | Maximal aggregated download throughput (bytes/sec). `-1` disables download throttling. |
+| networkProfile.upload   | number         | `-1`    | Maximal aggregated upload throughput (bytes/sec). `-1` disables upload throttling.     |
```
ka3de, line 15: I'm not too sure about this format, I think we are not using this in other places of the documentation, but can't really suggest a better one, and I can understand it. The only option that I can see is to define the `NetworkProfile` under browser, in the same manner we do for `ConsoleMessage` and then link to that from this doc. But no strong opinion.

The same applies for the CPU profile.

ka3de (approved):
Overall LGTM. Just a consideration, but I think @heitortsergent will provide better feedback on the organization. As far as the content goes it is correct to me.

docs/sources/next/javascript-api/k6-experimental/browser/page/throttlenetwork.md
inancgumus: There is also [`Cookie`])(https://k6.io/docs/javascript-api/k6-experimental/browser/browsercontext/cookie) that we put under `BrowserContext`. I believe it makes sense to keep it under `Page` as it's specific to a `Page`. Like you, no strong opinion.

inancgumus (approved):
👍

docs/sources/next/javascript-api/k6-experimental/browser/page/throttlenetwork.md
ankur22: Good point, I hadn't spotted that pattern.

I think for `cookies` it makes sense to have the class defined in a separate file since it gets referenced by more than one API. I'd argue that it might be best to keep the `ConsoleMessage` with `page.on` for now until it's needed by more than one API... although maybe at the time it made sense to separate them.

For the throttling APIs i feel that there's more value in having the input argument classes defined with the APIs, which also means less navigation back and forth between the API and the class definition. The downside is that it breaks away from how we have done this with the cookies and page.on APIs.

Any thoughts @heitortsergent?

ankur22: Actually, we have other APIs (such as `page.setViewportSize`) where the `viewportSize` class is defined with the API. So I think it's ok to have the throttle input classes defined with the APIs too.

inancgumus: I'm happy with the current implementation.

PR #1417, opened 2023-11-17
Option --profiling-enabled

olegbespalov:
# What?

It documents a new `--profiling-enabled` option from the https://github.com/grafana/k6/pull/3370/files.

# Why?

This is a new option and it should be documented.


docs/sources/next/using-k6/k6-options/reference.md
```diff
@@ -904,6 +905,22 @@ export const options = {
 
 {{< /code >}}
 
+## Profiling Enabled
+
+Enables [pprof](https://pkg.go.dev/net/http/pprof) profiling endpoints under the k6's REST API [address](#address). These endpoints help debug and profile k6 itself. k6's REST API should be enabled as well.
```
heitortsergent, line 910: @olegbespalov I think this looks good, I just have one question here, what does "k6's REST API should be enabled as well" mean? Is that something the user needs to do as well for profiling to work?

heitortsergent (approved):

docs/sources/next/using-k6/k6-options/reference.md
joanlopez: I guess it refers to this: https://k6.io/docs/misc/k6-rest-api/

However, the docs say: *`By default, that server listens on localhost:6565...`*, so I guess that comes by default, and just enabling profiling would be enough? 🤔

olegbespalov: The point is that enabled REST API is required for the profiling endpoints. It's enabled by default but could be manually turned off; in that case, profiling endpoints won't be exposed. So, this sentence tries to document that case.

joanlopez: All good and clear to me :)

joanlopez (approved):

oleiade (approved):

docs/sources/next/using-k6/k6-options/reference.md
heitortsergent: Gotcha, that makes sense. How would one manually turn that off? 🤔

PR #1419, opened 2023-11-20
Document `k6/experimental/fs` module

oleiade:
The `k6/experimental/fs` module will land in v0.48.0 and will empower users to read files efficiently and conveniently within their k6 scripts.

This Pull Request documents the module's APIs.

### DIFF
--- a/docs/sources/next/javascript-api/k6-experimental/_index.md
+++ b/docs/sources/next/javascript-api/k6-experimental/_index.md
@@ -1,6 +1,6 @@
 ---
 title: 'k6/experimental'
-excerpt: 'k6 experimental APIs'
+description: 'k6 experimental APIs'
 weight: 07
 ---
 
@@ -11,10 +11,10 @@ weight: 07
 | Modules                                                                                          | Description                                                                                                              |
 | ------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------ |
 | [browser](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/browser)       | Provides browser-level APIs to interact with browsers and collect frontend performance metrics as part of your k6 tests. |
+| [fs](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/fs)                 | Provides a memory-efficient way to handle file interactions within your test scripts.                                    |
+| [grpc](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/grpc)             | Extends `k6/net/grpc` with the streaming capabilities.                                                                   |
 | [redis](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/redis)           | Functionality to interact with [Redis](https://redis.io/).                                                               |
 | [timers](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/timers)         | `setTimeout`, `clearTimeout`, `setInterval`, `clearInterval`                                                             |
 | [tracing](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/tracing)       | Support for instrumenting HTTP requests with tracing information.                                                        |
 | [webcrypto](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/webcrypto)   | Implements the [WebCrypto API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API).                         |
 | [websockets](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/websockets) | Implements the browser's [WebSocket API](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket).                    |
-| [grpc](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/grpc)             | Extends `k6/net/grpc` with the streaming capabilities.                                                                   |
-| [fs](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/fs)                 | Provides a memory-efficient way to handle file interactions within your test scripts.                                    |
--- a/docs/sources/next/javascript-api/k6-experimental/_index.md
+++ b/docs/sources/next/javascript-api/k6-experimental/_index.md
@@ -1,6 +1,6 @@
 ---
-title: "k6/experimental"
-excerpt: "k6 experimental APIs"
+title: 'k6/experimental'
+excerpt: 'k6 experimental APIs'
 weight: 07
 ---
 
@@ -8,8 +8,8 @@ weight: 07
 
 {{< docs/shared source="k6" lookup="experimental-module.md" version="<K6_VERSION>" >}}
 
-| Modules                                                                           | Description                                                                                                              |
-| --------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------ |
+| Modules                                                                                          | Description                                                                                                              |
+| ------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------ |
 | [browser](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/browser)       | Provides browser-level APIs to interact with browsers and collect frontend performance metrics as part of your k6 tests. |
 | [redis](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/redis)           | Functionality to interact with [Redis](https://redis.io/).                                                               |
 | [timers](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/timers)         | `setTimeout`, `clearTimeout`, `setInterval`, `clearInterval`                                                             |
@@ -17,3 +17,4 @@ weight: 07
 | [webcrypto](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/webcrypto)   | Implements the [WebCrypto API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API).                         |
 | [websockets](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/websockets) | Implements the browser's [WebSocket API](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket).                    |
 | [grpc](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/grpc)             | Extends `k6/net/grpc` with the streaming capabilities.                                                                   |
+| [fs](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/fs)                 | Provides a memory-efficient way to handle file interactions within your test scripts.                                    |
--- a/docs/sources/next/javascript-api/k6-experimental/fs/FileInfo.md
+++ b/docs/sources/next/javascript-api/k6-experimental/fs/FileInfo.md
@@ -33,7 +33,6 @@ export default async function () {
   if (fileinfo.name != 'bonjour.txt') {
     throw new Error('Unexpected file name');
   }
-
 }
 ```
 
--- a/docs/sources/next/javascript-api/k6-experimental/fs/FileInfo.md
+++ b/docs/sources/next/javascript-api/k6-experimental/fs/FileInfo.md
@@ -1,0 +1,67 @@
+---
+title: 'FileInfo'
+excerpt: 'FileInfo represents information about a file.'
+weight: 30
+---
+
+# FileInfo
+
+The `FileInfo` class represents information about a [file](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/fs/file).
+
+## Properties
+
+| Property | Type   | Description                    |
+| :------- | :----- | :----------------------------- |
+| name     | string | The name of the file.          |
+| size     | number | The size of the file in bytes. |
+
+## Example
+
+{{< code >}}
+
+```javascript
+import { open, SeekMode } from 'k6/experimental/fs';
+
+let file;
+(async function () {
+  file = await open('bonjour.txt');
+})();
+
+export default async function () {
+  // About information about the file
+  const fileinfo = await file.stat();
+  if (fileinfo.name != 'bonjour.txt') {
+    throw new Error('Unexpected file name');
+  }
+
+  const buffer = new Uint8Array(4);
+
+  let totalBytesRead = 0;
+  while (true) {
+    // Read into the buffer
+    const bytesRead = await file.read(buffer);
+    if (bytesRead == null) {
+      // EOF
+      break;
+    }
+
+    // Do something useful with the content of the buffer
+    totalBytesRead += bytesRead;
+
+    // If bytesRead is less than the buffer size, we've read the whole file
+    if (bytesRead < buffer.byteLength) {
+      break;
+    }
+  }
+
+  // Check that we read the expected number of bytes
+  if (totalBytesRead != fileinfo.size) {
+    throw new Error('Unexpected number of bytes read');
+  }
+
+  // Seek back to the beginning of the file
+  await file.seek(0, SeekMode.Start);
+}
+```
+
+{{< /code >}}
--- a/docs/sources/next/javascript-api/k6-experimental/fs/FileInfo.md
+++ b/docs/sources/next/javascript-api/k6-experimental/fs/FileInfo.md
@@ -34,33 +34,6 @@ export default async function () {
     throw new Error('Unexpected file name');
   }
 
-  const buffer = new Uint8Array(4);
-
-  let totalBytesRead = 0;
-  while (true) {
-    // Read into the buffer
-    const bytesRead = await file.read(buffer);
-    if (bytesRead == null) {
-      // EOF
-      break;
-    }
-
-    // Do something useful with the content of the buffer
-    totalBytesRead += bytesRead;
-
-    // If bytesRead is less than the buffer size, we've read the whole file
-    if (bytesRead < buffer.byteLength) {
-      break;
-    }
-  }
-
-  // Check that we read the expected number of bytes
-  if (totalBytesRead != fileinfo.size) {
-    throw new Error('Unexpected number of bytes read');
-  }
-
-  // Seek back to the beginning of the file
-  await file.seek(0, SeekMode.Start);
 }
 ```
 
--- a/docs/sources/next/javascript-api/k6-experimental/fs/FileInfo.md
+++ b/docs/sources/next/javascript-api/k6-experimental/fs/FileInfo.md
@@ -1,6 +1,6 @@
 ---
 title: 'FileInfo'
-excerpt: 'FileInfo represents information about a file.'
+description: 'FileInfo represents information about a file.'
 weight: 30
 ---
 
@@ -28,7 +28,7 @@ let file;
 })();
 
 export default async function () {
-  // About information about the file
+  // Retrieve information about the file
   const fileinfo = await file.stat();
   if (fileinfo.name != 'bonjour.txt') {
     throw new Error('Unexpected file name');
--- a/docs/sources/next/javascript-api/k6-experimental/fs/SeekMode.md
+++ b/docs/sources/next/javascript-api/k6-experimental/fs/SeekMode.md
@@ -1,0 +1,43 @@
+---
+title: 'SeekMode'
+excerpt: 'SeekMode is used to specify the position from which to seek in a file.'
+weight: 40
+---
+
+# SeekMode
+
+The `SeekMode` enum is used to specify the position from which to [seek](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/fs/file/seek) in a file.
+
+## Members
+
+| Member  | Value | Description                                 |
+| :------ | :---- | :------------------------------------------ |
+| Start   | 0     | Seek from the start of the file.            |
+| Current | 1     | Seek from the current position in the file. |
+| End     | 2     | Seek from the end of the file.              |
+
+## Example
+
+{{< code >}}
+
+```javascript
+import { open, SeekMode } from 'k6/experimental/fs';
+
+let file;
+(async function () {
+  file = await open('bonjour.txt');
+})();
+
+export default async function () {
+  // Seek 6 bytes from the start of the file
+  await file.seek(6, SeekMode.Start);
+
+  // Seek 2 more bytes from the current position
+  await file.seek(2, SeekMode.Current);
+
+  // Seek backwards 2 bytes from the end of the file
+  await file.seek(-2, SeekMode.End);
+}
+```
+
+{{< /code >}}
--- a/docs/sources/next/javascript-api/k6-experimental/fs/SeekMode.md
+++ b/docs/sources/next/javascript-api/k6-experimental/fs/SeekMode.md
@@ -1,12 +1,12 @@
 ---
 title: 'SeekMode'
-excerpt: 'SeekMode is used to specify the position from which to seek in a file.'
+description: 'SeekMode is used to specify the position from which to seek in a file.'
 weight: 40
 ---
 
 # SeekMode
 
-The `SeekMode` enum is used to specify the position from which to [seek](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/fs/file/seek) in a file.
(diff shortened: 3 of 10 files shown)


docs/sources/next/javascript-api/k6-experimental/fs/open.md
```diff
@@ -0,0 +1,63 @@
+---
+title: 'open'
+excerpt: 'open opens a file and returns a promise resolving to a File instance.'
+weight: 20
+---
+
+# open
+
+The `open` function opens a file and returns a promise that resolves to a [File](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/fs/file) instance. This function loads the file into shared memory space accessible by all Virtual Users (VUs). This approach significantly reduces the memory footprint, enabling the handling of large files without overwhelming memory resources.
```
codebien, line 9:
```suggestion
The `open` function opens a file and returns a promise that resolves to a [File](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/fs/file) instance. This function loads the file into shared memory space accessible by all Virtual Users (VUs). This approach significantly ~reduces the memory footprint~, enabling the handling of large files without overwhelming memory resources.
```
Of course, it compares with the old `open` function but maybe the reader is not aware of it.

docs/sources/next/javascript-api/k6-experimental/fs/open.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Asynchronous Nature
+
+It's important to note that `open` is asynchronous and returns a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise). Due to k6's current limitation with the init context (which doesn't support asynchronous functions directly), you need to use an asynchronous wrapper like this:
```
codebien, line 13:
```suggestion
It's important to note that `open` is asynchronous and returns a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise). Due to k6's current limitation with the Init context (which doesn't support asynchronous functions directly), you need to use an asynchronous wrapper like this:
```

docs/sources/next/javascript-api/k6-experimental/fs/file/read.md
```diff
@@ -0,0 +1,72 @@
...
+})();
+
+export default async function () {
+  // About information about the file
+  const fileinfo = await file.stat();
+  if (fileinfo.name != 'bonjour.txt') {
+    throw new Error('Unexpected file name');
+  }
+
+  const buffer = new Uint8Array(4);
```
codebien, line 42:
```suggestion
  const buffer = new Uint8Array(4);
```
I don't think it's the usual case to know the size of the file to read, we need a more dynamic example here.

docs/sources/next/javascript-api/k6-experimental/fs/file/read.md
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  // Check that we read the expected number of bytes
+  if (totalBytesRead != fileinfo.size) {
+    throw new Error('Unexpected number of bytes read');
+  }
+
+  // Seek back to the beginning of the file
+  await file.seek(0, SeekMode.Start);
+}
```
codebien, line 69: It could be useful to wrap this block code in a helper function `readAll`. It should be more expressive for an example and it could be a starting point for users as a copy-and-paste snippet.

codebien (commented):
@oleiade does it work `make docs` for you? I wasn't able to see the preview with the current changes included.

oleiade:
@codebien `make docs` does work on my side, what error are you experiencing?

```bash
$ cd docs/

$ make docs
docker pull -q grafana/docs-base:latest
docker.io/grafana/docs-base:latest
/Users/theocrevon/Dev/grafana/documentation/k6-docs/docs/make-docs k6:UNVERSIONED:k6-docs
node scripts/docs/versions.js
k6 -> v0.47.x
/usr/local/bin/hugo server --environment=docs -p 3003 --buildDrafts=false --buildFuture=false --bind 0.0.0.0 --noHTTPCache --minify --poll 700ms --renderToDisk -d dist | grep -Ev 'REF_NOT_FOUND: Ref "/(go|blog)/'
Start building sites …
hugo v0.115.2 linux/arm64 BuildDate=unknown


                   |  EN  |  DE  |  FR  |  ES  | PT-BR | ZH-CN
-------------------+------+------+------+------+-------+--------
  Pages            | 2328 |   12 |   12 |   12 |    12 |    12
  Paginator pages  |    0 |    0 |    0 |    0 |     0 |     0
  Non-page files   |    0 |    0 |    0 |    0 |     0 |     0
  Static files     | 1274 | 1274 | 1274 | 1274 |  1274 |  1274
  Processed images |    0 |    0 |    0 |    0 |     0 |     0
  Aliases          |    4 |    0 |    0 |    0 |     0 |     0
  Sitemaps         |    2 |    1 |    1 |    1 |     1 |     1
  Cleaned          |    0 |    0 |    0 |    0 |     0 |     0

Built in 2541 ms
Watching for changes in /hugo/{content,data,i18n,layouts,package.json,static}
Use watcher with poll interval 700ms
Watching for config changes in /hugo/config/_default, /hugo/config/docs
Environment: "docs"
Serving pages from disk
Running in Fast Render Mode. For full rebuilds on change: hugo server --disableFastRender

View documentation locally:
                             http://localhost:3002/docs/k6/

                                                           Press Ctrl+C to stop the server
```

docs/sources/next/javascript-api/k6-experimental/fs/open.md
oleiade: I compared it to root-level/native (I don't know how to call it) `open` function somewhere indeed, I could probably also do it here  👍🏻

docs/sources/next/javascript-api/k6-experimental/fs/file/read.md
oleiade: That's a good point, I'll adjust the example 👍🏻

docs/sources/next/javascript-api/k6-experimental/fs/file/read.md
oleiade: I would have preferred a `fs.readAll` function for doing that. But I see where you're coming from, I'll add a second example demonstrating it then, as I think it's important to display the "low-level" usage of the `read` function, too 👍🏻

docs/sources/next/javascript-api/k6-experimental/fs/open.md
codebien: @oleiade I see you applied the suggestion, was not my intent to add it. It was mostly for signaling the place that I was referencing. Sorry, for not being clear.

Doing directly a concrete suggestion:
```
s/memory footprint/memory footprint from the same usage done using [open](link to builtin)/
```

docs/sources/next/javascript-api/k6-experimental/fs/file/read.md
```diff
@@ -0,0 +1,126 @@
...
+  // Obtain the size of the file
+  const fileInfo = await file.stat();
+  const fileSize = fileInfo.size;
+
+  // Prepare a buffer to store the file content
+  const fileContent = new Uint8Array(fileInfo.size);
+  let fileOffset = 0;
+
+  // Prepare a buffer to read the chunks of the file into
+  const buffer = new Uint8Array(Math.min(bufferSize, fileSize));
```
codebien, line 91: Is it convenient to read in chunks? If yes, probably we should detail _why_. Isn't the implementation optimized for doing it?


I expect `readAll` wouldn't be something more than the following. Am I wrong?
```
async function readAll(file) {
    const finfo = await file.stat();
    const buffer = new Uint8Array(finfo.size+1);

    let bytesRead = await file.read(buffer);
    if (bytesRead !== fsize) {
        // we don't expect to have more to read
	throw new Error("returned bytes read are not as expected");
    }
    return buffer;
}
```

docs/sources/next/javascript-api/k6-experimental/fs/_index.md
```diff
@@ -0,0 +1,84 @@
...
+
+# fs
+
+{{< docs/shared source="k6" lookup="experimental-module.md" version="<K6_VERSION>" >}}
+
+The k6 filesystem experimental module provides a memory-efficient way to handle file interactions within your test scripts. It currently offers support for opening files, reading their content, and seeking through their content, and retrieving metadata about them.
+
+### Memory efficiency
+
+One of the key advantages of the filesystem module is its memory efficiency. Unlike the traditional [open]() function, which loads a file multiple times into memory, the filesystem module reduces memory usage by loading the file as little possible, and sharing the same memory space between all VUs. This approach signally reduces the memory footprint of your test script, and lets you load and process large files without running out of memory.
```
olegbespalov, line 15: >and lets you load and process large files without running out of memory.

I'm wondering if we should adjust this place to something like: 

```suggestion
One of the key advantages of the filesystem module is its memory efficiency. Unlike the traditional [open]() function, which loads a file multiple times into memory, the filesystem module reduces memory usage by loading the file as little possible, and sharing the same memory space between all VUs. This approach reduces the number of running out-of-memory errors while processing large files.
```

Since I am not sure if the new module without file-streams support allows customers to leverage the issues with the large files fully

WDYT?

docs/sources/next/javascript-api/k6-experimental/fs/file/_index.md
```diff
@@ -0,0 +1,75 @@
+---
+title: 'File'
+excerpt: 'File represents a file with methods for reading, seeking, and obtaining file stats.'
+weight: 10
+weight: 10
+---
+
+# File
+
+The `File` class represents a file with methods for reading, seeking, and obtaining file stats. It's returned by the [open](link-to-open-doc) function.
```
olegbespalov, line 10: Is this `link-to-open-doc` a placeholder?

docs/sources/next/javascript-api/k6-experimental/fs/file/_index.md
```diff
@@ -0,0 +1,75 @@
+---
+title: 'File'
+excerpt: 'File represents a file with methods for reading, seeking, and obtaining file stats.'
+weight: 10
+weight: 10
```
olegbespalov, line 5:
```suggestion
weight: 10
```

docs/sources/next/javascript-api/k6-experimental/fs/_index.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Notes on usage
+
+An important consideration when using the filesystem module is its handling of external file modifications. Once a file is loaded in k6, it behaves like a "view" over the content of the file. This means that if the underlying file is modified externally during a test, those changes won't be reflected in the loaded [File]() instance.
```
olegbespalov, line 19: I guess a link in `[File]()` is missing. Also, it seems like the other files have a similar issue.

docs/sources/next/javascript-api/k6-experimental/fs/file/_index.md
oleiade: I forgot it behind 👍🏻 The new linking system relies on absolute URLs, which is a bit of a pain, so I put placeholders while I write, and make a pass once I'm done to fill the links. Obviously forgot that one, so I might need a better system...

docs/sources/next/javascript-api/k6-experimental/fs/_index.md
oleiade: 👍🏻 

I've rephrased to "This approach reduces the risk of encountering out-of-memory errors, especially in load tests involving large files."

docs/sources/next/javascript-api/k6-experimental/fs/file/_index.md
```diff
@@ -0,0 +1,75 @@
...
+  if (totalBytesRead != fileinfo.size) {
+    throw new Error('Unexpected number of bytes read');
+  }
+
+  // Seek back to the beginning of the file
+  await file.seek(0, SeekMode.Start);
+}
+```
+
+{{< /code >}}
```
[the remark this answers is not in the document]
oleiade: I prefer if we do, considering that users might land on this page directly, coming from a search engine, and not have the outer context. That way, if they landed there directly, they have an overview of the usage of file already 👍🏻

docs/sources/next/javascript-api/k6-experimental/fs/file/read.md
oleiade: We could also do it this way indeed, I don't really mind. I just wanted to illustrate again how the `read` API can be used at a low-level 👍🏻 

PS: why the `fileinfo.size + 1`?

docs/sources/next/javascript-api/k6-experimental/fs/file/read.md
```diff
@@ -0,0 +1,126 @@
...
+
+    // Do something useful with the content of the buffer
+    fileOffset += bytesRead;
+
+    // If bytesRead is less than the buffer size, we've read the whole file
+    if (bytesRead < buffer.byteLength) {
+      break;
+    }
+  }
+
```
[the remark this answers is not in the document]
oleiade: I disagree on this one. It is not the responsibility of a read operation to also seek. At least in the context like this one where the `readAll` function takes a `File` as input. 

If we were to change it take a path as input instead (which we can't do because `fs.open` is limited to the Init context) I would say that's okay because the `readAll` then would be responsible for managing the `File` object state too.

docs/sources/next/javascript-api/k6-experimental/fs/file/read.md
```diff
@@ -0,0 +1,126 @@
...
+  return fileContent;
+}
+
+export default async function () {
+  // Read the whole file
+  const fileContent = await readAll(file);
+  console.log(JSON.stringify(fileContent));
+
+  // Seek back to the beginning of the file
+  await file.seek(0, SeekMode.Start);
```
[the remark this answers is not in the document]
oleiade: See https://github.com/grafana/k6-docs/pull/1419/files#r1400445346 👆

docs/sources/next/javascript-api/k6-experimental/fs/file/read.md
codebien: >We could also do it this way indeed, I don't really mind.

I think our examples are the main place, where we can have opinionated k6 code. Where users can find idiomatic k6, best practices and they have help regarding how to write code that scales and is not error-prone.

>PS: why the fileinfo.size + 1?

In this way, we can be really sure if we have read the entire file or otherwise there is more in the file to read. I'm not sure if there are better alternatives.

docs/sources/next/javascript-api/k6-experimental/fs/_index.md
```diff
@@ -0,0 +1,84 @@
...
+
+# fs
+
+{{< docs/shared source="k6" lookup="experimental-module.md" version="<K6_VERSION>" >}}
+
+The k6 filesystem experimental module provides a memory-efficient way to handle file interactions within your test scripts. It currently offers support for opening files, reading their content, and seeking through their content, and retrieving metadata about them.
+
+### Memory efficiency
+
+One of the key advantages of the filesystem module is its memory efficiency. Unlike the traditional [open](https://grafana.com/docs/k6/latest/javascript-api/init-context/open/) function, which loads a file multiple times into memory, the filesystem module reduces memory usage by loading the file as little possible, and sharing the same memory space between all VUs. This approach reduces the risk of encountering out-of-memory errors, especially in load tests involving large files.
```
heitortsergent, line 15:
```suggestion
One of the key advantages of the filesystem module is its memory efficiency. Unlike the traditional [open](https://grafana.com/docs/k6/latest/javascript-api/init-context/open/) function, which loads a file multiple times into memory, the filesystem module reduces memory usage by loading the file as little as possible and sharing the same memory space between all VUs. This approach reduces the risk of encountering out-of-memory errors, especially in load tests involving large files.
```

Small nit here, but also out of curiosity, what does "as little as possible" mean? Does it mean k6 is able to only load parts of the file as needed, or does it mean it can load the file fewer times than the traditional open function? And if it's the latter, would it always load the file fewer times?

docs/sources/next/javascript-api/k6-experimental/fs/_index.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Notes on usage
+
+An important consideration when using the filesystem module is its handling of external file modifications. Once a file is loaded in k6, it behaves like a "view" over the content of the file. This means that if the underlying file is modified externally during a test, those changes won't be reflected in the loaded [File](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/fs/file/) instance.
```
heitortsergent, line 19:
```suggestion
An important consideration when using the filesystem module is its handling of external file modifications. Once k6 loads a file, it behaves like a "view" over its contents. If you modify the underlying file during a test, k6 will not reflect those changes in the loaded [File](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/fs/file/) instance.
```

docs/sources/next/javascript-api/k6-experimental/_index.md
```diff
@@ -1,19 +1,20 @@
 ---
-title: "k6/experimental"
-excerpt: "k6 experimental APIs"
+title: 'k6/experimental'
+excerpt: 'k6 experimental APIs'
```
heitortsergent, line 3:
```suggestion
description: 'k6 experimental APIs'
```

docs/sources/next/javascript-api/k6-experimental/fs/FileInfo.md
```diff
@@ -0,0 +1,39 @@
+---
+title: 'FileInfo'
+excerpt: 'FileInfo represents information about a file.'
```
heitortsergent, line 3:
```suggestion
description: 'FileInfo represents information about a file.'
```

docs/sources/next/javascript-api/k6-experimental/fs/SeekMode.md
```diff
@@ -0,0 +1,43 @@
+---
+title: 'SeekMode'
+excerpt: 'SeekMode is used to specify the position from which to seek in a file.'
```
heitortsergent, line 3:
```suggestion
description: 'SeekMode is used to specify the position from which to seek in a file.'
```

docs/sources/next/javascript-api/k6-experimental/fs/_index.md
```diff
@@ -0,0 +1,84 @@
+---
+title: 'fs'
+excerpt: 'k6 fs experimental API'
```
heitortsergent, line 3:
```suggestion
description: 'k6 fs experimental API'
```

docs/sources/next/javascript-api/k6-experimental/fs/file/_index.md
```diff
@@ -0,0 +1,74 @@
+---
+title: 'File'
+excerpt: 'File represents a file with methods for reading, seeking, and obtaining file stats.'
```
heitortsergent, line 3:
```suggestion
description: 'File represents a file with methods for reading, seeking, and obtaining file stats.'
```

docs/sources/next/javascript-api/k6-experimental/fs/file/_index.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Methods
+
+| Method                                                                                       | Description                                                                                                                            |
+| :------------------------------------------------------------------------------------------- | :------------------------------------------------------------------------------------------------------------------------------------- |
+| [read](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/fs/file/read) | Reads up to `buffer.byteLength` bytes from the file into the passed `buffer`. Returns a promise resolving to the number of bytes read. |
+| [seek](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/fs/file/seek) | Sets the file position indicator for the file to the passed `offset` bytes. Returns a promise resolving to the new offset.             |
+| [stat](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/fs/file/stat) | Returns a promise resolving to a [FileInfo](link-to-fileinfo-doc) object with information about the file.                              |
```
heitortsergent, line 23:
```suggestion
| [stat](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/fs/file/stat) | Returns a promise resolving to a [FileInfo](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/fs/fileinfo/) object with information about the file.                              |
```

docs/sources/next/javascript-api/k6-experimental/fs/file/read.md
```diff
@@ -0,0 +1,109 @@
+---
+title: 'read'
+excerpt: 'the read method is used to read a chunk of the file.'
```
heitortsergent, line 3:
```suggestion
description: 'the read method is used to read a chunk of the file.'
```

docs/sources/next/javascript-api/k6-experimental/fs/file/read.md
```diff
@@ -0,0 +1,109 @@
...
+
+A [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) resolving to the number of bytes read or `null` if the end of the file has been reached.
+
+## Example
+
+### Reading a file
+
+In the following example, we open a file and read it in chunks of 128 bytes until we reach the end of the file.
+
+{{< code >}}
```
heitortsergent, line 29: Just as a comment, the `{{< code >}}` shortcode main purpose is to support multiple code snippets (how to do X in JavaScript, and in Ruby, for example). It still works if you have a single code snippet inside of it, but we don't necessarily need it here. 😄

docs/sources/next/javascript-api/k6-experimental/fs/file/seek.md
```diff
@@ -0,0 +1,52 @@
+---
+title: 'seek'
+excerpt: 'seek sets the file position indicator for the file to the passed offset bytes.'
```
heitortsergent, line 3:
```suggestion
description: 'seek sets the file position indicator for the file to the passed offset bytes.'
```

docs/sources/next/javascript-api/k6-experimental/fs/file/stat.md
```diff
@@ -0,0 +1,38 @@
+---
+title: 'stat'
+excerpt: 'stat returns a promise resolving to a FileInfo object with information about the file.'
```
heitortsergent, line 3:
```suggestion
description: 'stat returns a promise resolving to a FileInfo object with information about the file.'
```

docs/sources/next/javascript-api/k6-experimental/fs/open.md
```diff
@@ -0,0 +1,63 @@
+---
+title: 'open'
+excerpt: 'open opens a file and returns a promise resolving to a File instance.'
```
heitortsergent, line 3:
```suggestion
description: 'open opens a file and returns a promise resolving to a File instance.'
```

docs/sources/next/javascript-api/k6-experimental/fs/open.md
```diff
@@ -0,0 +1,63 @@
...
+weight: 20
+---
+
+# open
+
+The `open` function opens a file and returns a promise that resolves to a [File](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/fs/file) instance. Unlike the traditional [open](https://grafana.com/docs/k6/latest/javascript-api/init-context/open/) function, which loads a file multiple times into memory, the filesystem module reduces memory usage by loading the file as little possible, and sharing the same memory space between all VUs. This approach reduces the risk of encountering out-of-memory errors, especially in load tests involving large files.
+
+### Asynchronous Nature
```
heitortsergent, line 11:
```suggestion
### Asynchronous nature
```

heitortsergent (commented):
Left a few comments, mostly small nits!

docs/sources/next/javascript-api/k6-experimental/fs/_index.md
oleiade: In practice, the file will be opened twice for the whole test run under the hood. But that's an implementation detail that users should probably not be aware of, and they should treat it as if it was opened only once (O(1) space).

Thus, "as little as possible" was a way to say, "it's optimized, and the detail doesn't matter to 99.9% of users, plus they might change in the future" 😄 But I'm happy to modify it if you have a suggestion 🙇🏻

docs/sources/next/javascript-api/k6-experimental/fs/file/read.md
oleiade: Oh, that's good to know! I think it still makes sense to have it even with a single example? Makes it more explicit too as it will display "Javascript" above the code right?

codebien (approved):

olegbespalov (approved):

heitortsergent (approved):

PR #1426, opened 2023-11-23
Document a new metrics restriction

olegbespalov:
# What?

Documenting a new restriction that was introduced in https://github.com/grafana/k6/pull/3448

# Why?

It's important to have in docs.

### DIFF
--- a/docs/sources/next/using-k6/metrics/_index.md
+++ b/docs/sources/next/using-k6/metrics/_index.md
@@ -21,10 +21,10 @@ To make a test fail a certain criteria, you can write a [Threshold](https://graf
 To filter metrics, you can use [Tags and groups](https://grafana.com/docs/k6/<K6_VERSION>/using-k6/tags-and-groups).
 You can also export metrics in various summary and granular formats, as documented in [Results output](https://grafana.com/docs/k6/<K6_VERSION>/results-output).
 
-| On this page...                                                                           | Read about...                                                                                   |
-| ----------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------- |
+| On this page...                                                                                          | Read about...                                                                                                   |
+| -------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


docs/sources/next/using-k6/metrics/_index.md
```diff
@@ -111,8 +111,4 @@ For details of all metrics, refer to the [Metrics reference](https://grafana.com
 
 Metric names must comply with OpenTelemetry and [Prometheus limitations](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels). The character limit is the same limit k6 had before restricting the character set down.
 
-That means they must not start with a number and a metric name can be 1 to 128 symbols of:
-
-1. any Unicode Letters
-2. any Unicode Number
-3. `_` (an underscore)
+That means metrics names must only include up to 128 ASCII letters, numbers, or underscores and start with a letter or an underscore.
```
heitortsergent, line 114:
```suggestion
That means metrics names must:

- Start with a letter or an underscore.
- Include up to 128 symbols (ASCII letters, numbers, or underscores).
```

I like keeping it as a list, I think it makes it easier to parse. What do you think? 🤔

olegbespalov: TBH, in that particular case, for me, a single sentence seems more straightforward than a list of two items. 

On the other hand, I'm not firm about that, so if you think that the list is better for a reader, I'll trust you and will be happy to go with the list :relaxed:

olegbespalov: @mstoykov could also bring an input :+1:

oleiade (approved):
I'm happy with either of the way discussed in https://github.com/grafana/k6-docs/pull/1426/files#r1406782372 🙇🏻

joanlopez (approved):

docs/sources/next/using-k6/metrics/_index.md
mstoykov: I am fine with both, but I would reverse the order of the bullet points, as `letter` in the first is not specified to be ASCII one, so IMO makes it harder to grasp.

heitortsergent: I reversed the order of the bullet points in my suggestion. :)

It's a small thing, but I do think, generally, lists are easier to grasp for readers, especially when you're talking about conditions.

PR #1427, opened 2023-11-27
fix JSON for run example test

lucasjct:
Hello everyone!

### What  
Removed comma on the last line from JSON example. 
Follow link related: 
https://k6.io/docs/testing-guides/api-load-testing/#data-parameterization

### Why  
Because the format is incorrect and the test broken when running.  

Thanks!


heitortsergent:
Good catch, thanks for opening a PR @lucasjct! Obrigado 🇧🇷 :shipit: .

PR #1431, opened 2023-11-29
Update version banner

heitortsergent:
## What?

Update the version banner for the JavaScript API section to display on latest version, and update link to point to Grafana docs.

## Checklist

Please fill in this template:

### DIFF
--- a/src/components/shared/page-info/page-info.module.scss
+++ b/src/components/shared/page-info/page-info.module.scss
@@ -1,6 +1,6 @@
 .container {
   margin-bottom: 20px;
-  margin-top: 115px;
+  margin-top: 165px;
 
   @include xl-down {
     margin-top: 160px;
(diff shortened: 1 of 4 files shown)


heitortsergent:
@ppcano I think I was able to figure out the version banner, let me know how it looks! 🙏

src/components/shared/page-info/page-info.module.scss
```diff
@@ -1,6 +1,6 @@
 .container {
   margin-bottom: 20px;
-  margin-top: 115px;
+  margin-top: 165px;
```
ppcano, line 3: Screenshot 2023-11-29 at 23 51 02


On my screen, the margin seems too large.

ppcano (commented):
Commented about the `margin-top` changes. The rest looks perfect!

src/components/shared/page-info/page-info.module.scss
heitortsergent: Ah good catch, thanks @ppcano! I just updated the PR, I think it's fixed now so the higher `margin-top` only applies to the JS API section.

Issue #1435, opened 2023-11-30
Update error message for patch script when running it outside of root folder


heitortsergent:
If you try running the [`apply-patch`](https://github.com/grafana/k6-docs/blob/main/scripts/apply-patch) outside of the root folder, you get an error saying:

```
error: No valid patches in input (allow with "--allow-empty")
```

It would be nice to make the error message clearer for this case to guide users to re-run it from the root folder instead.

heitortsergent:
@jdbaldry is this something that would be easy to do?

PR #1438, opened 2023-12-04
Add check for running in the root of the repository

jdbaldry:
```console
$ ../scripts/apply-patch HEAD~ sources/next sources/v0.47.x
ERROR: Script has not been run from the repository root ('/Users/jdb/ext/grafana/k6-docs').
Please change directory there and adjust argument paths to match the new working directory.
```

Closes: #1435

### DIFF
--- a/scripts/apply-patch
+++ b/scripts/apply-patch
@@ -3,6 +3,7 @@
 function usage {
   cat <<EOF
 Apply the changes for directory SRC for commit COMMIT to directory DST.
+The script assumes it is run from the root of the repository.
 
 Usage:
   $0 <COMMIT> <SRC> <DST>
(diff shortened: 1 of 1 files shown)


scripts/apply-patch
```diff
@@ -50,5 +51,13 @@ function path_components {
   echo "${n}"
 }
 
+GIT_ROOT="$(git rev-parse --show-toplevel)"
+if [[ "$(pwd)" != "${GIT_ROOT}" ]]; then
+  echo "ERROR: Script has not been run from the repository root ('${GIT_ROOT}')."
+  echo "Please change directory there and adjust argument paths to match the new working directory."
```
heitortsergent, line 57:
```suggestion
  echo "Please move to the root directory and re-run the script."
```

What if we shorten this a little bit? At least from the feedback, I think the main issue was just running from the wrong folder, maybe adjusting the argument path is implied once we ask the user to change directories.

jdbaldry: It certainly is implied. Sometimes I find being explicit is necessary, and sometimes it's more confusing. I'm happy with your wording too!

heitortsergent (approved):

PR #1440, opened 2023-12-04
Update npm start command to run make docs

heitortsergent:
## What?

Update the `npm start` command to run `make docs`, so we have one command that can be run from the root of the repo and simplify the instructions.

Also create a new `npm run start-gatsby` so we can still build/preview the Gatsby version of the docs.


heitortsergent:
@jdbaldry I was thinking it'd be nice to be able to build the docs from the root of the repo, is this a silly way to do it? What do you think? 🤓

jdbaldry (approved):
Nice! Looks good to me and works for me locally!

ppcano (approved):

PR #1457, opened 2023-12-20
Update browser module isVisible and isHidden docs

ankur22:
## What?

This updates the docs for the `isVisible` and `isHidden` APIs in all associated classes.

## Checklist

Please fill in this template:

If updating the documentation for the next release of k6:

## Related PR(s)/Issue(s)

Related issue: https://github.com/grafana/xk6-browser/issues/981


docs/sources/next/javascript-api/k6-experimental/browser/frame/ishidden.md
```diff
@@ -0,0 +1,62 @@
...
+
+{{% /admonition %}}
+
+Checks if the element is `hidden`.
+
+<TableWithNestedRows>
+
+| Parameter      | Type    | Default | Description                                                                                                                                                |
+| -------------- | ------- | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| selector       | string  | `''`    | A selector to search for an element. If there are multiple elements satisfying the selector, the first will be used.                                       |
```
ka3de, line 20: Small suggestion:
```suggestion
| selector       | string  | `''`    | A selector to search for an element. If there are multiple elements satisfying the selector, the first one will be used.                                       |
```

docs/sources/next/javascript-api/k6-experimental/browser/frame/ishidden.md
```diff
@@ -0,0 +1,62 @@
...
+| options        | object  | `null`  |                                                                                                                                                            |
+| options.strict | boolean | `false` | When `true`, the call requires selector to resolve to a single element. If given selector resolves to more than one element, the call throws an exception. |
```
ka3de, line 22: Not sure if both options might be correct, but I was missing `the` in these cases:
```suggestion
| options.strict | boolean | `false` | When `true`, the call requires the selector to resolve to a single element. If the given selector resolves to more than one element, the call throws an exception. |
```

docs/sources/next/javascript-api/k6-experimental/browser/frame/ishidden.md
```diff
@@ -0,0 +1,62 @@
...
+
+</TableWithNestedRows>
+
+### Returns
+
+| Type | Description                                      |
+| ---- | ------------------------------------------------ |
+| bool | `true` if the element is `hidden`, else `false`. |
```
ka3de, line 30: Small suggestion, feel free to skip it if you think the current form is OK:
```suggestion
| bool | `true` if the element is `hidden`, `false` otherwise. |
```

docs/sources/next/javascript-api/k6-experimental/browser/frame/isvisible.md
```diff
@@ -0,0 +1,65 @@
...
+
+{{% /admonition %}}
+
+Checks if the element is `visible`.
+
+<TableWithNestedRows>
+
+| Parameter      | Type    | Default | Description                                                                                                                                                |
+| -------------- | ------- | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| selector       | string  | `''`    | A selector to search for an element. If there are multiple elements satisfying the selector, the first will be used.                                       |
```
ka3de, line 20: Small suggestion:
```suggestion
| selector       | string  | `''`    | A selector to search for an element. If there are multiple elements satisfying the selector, the first one will be used.                                       |
```

docs/sources/next/javascript-api/k6-experimental/browser/frame/isvisible.md
```diff
@@ -0,0 +1,65 @@
...
+| options        | object  | `null`  |                                                                                                                                                            |
+| options.strict | boolean | `false` | When `true`, the call requires selector to resolve to a single element. If given selector resolves to more than one element, the call throws an exception. |
```
ka3de, line 22:
```suggestion
| options.strict | boolean | `false` | When `true`, the call requires the selector to resolve to a single element. If the given selector resolves to more than one element, the call throws an exception. |
```

docs/sources/next/javascript-api/k6-experimental/browser/frame/isvisible.md
```diff
@@ -0,0 +1,65 @@
...
+
+</TableWithNestedRows>
+
+### Returns
+
+| Type | Description                                       |
+| ---- | ------------------------------------------------- |
+| bool | `true` if the element is `visible`, else `false`. |
```
ka3de, line 30:
```suggestion
| bool | `true` if the element is `visible`, `false` otherwise. |
```

docs/sources/next/javascript-api/k6-experimental/browser/page/ishidden.md
```diff
@@ -9,18 +9,17 @@ excerpt: 'Browser module: page.isHidden(selector[, options) method'
...
-| options.timeout | number  | `30000` | Maximum time in milliseconds. Pass `0` to disable the timeout. Default is overridden by the `setDefaultTimeout` option on [BrowserContext](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/browser/browsercontext/) or [Page](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/browser/page/). |
+| Parameter      | Type    | Default | Description                                                                                                                                                |
+| -------------- | ------- | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| selector       | string  | `''`    | A selector to search for an element. If there are multiple elements satisfying the selector, the first will be used.                                       |
```
ka3de, line 20:
```suggestion
| selector       | string  | `''`    | A selector to search for an element. If there are multiple elements satisfying the selector, the first one will be used.                                       |
```

docs/sources/next/javascript-api/k6-experimental/browser/page/ishidden.md
```diff
@@ -9,18 +9,17 @@ excerpt: 'Browser module: page.isHidden(selector[, options) method'
...
+| options        | object  | `null`  |                                                                                                                                                            |
+| options.strict | boolean | `false` | When `true`, the call requires selector to resolve to a single element. If given selector resolves to more than one element, the call throws an exception. |
```
ka3de, line 22:
```suggestion
| options.strict | boolean | `false` | When `true`, the call requires the selector to resolve to a single element. If the given selector resolves to more than one element, the call throws an exception. |
```

docs/sources/next/javascript-api/k6-experimental/browser/page/isvisible.md
```diff
@@ -9,18 +9,17 @@ excerpt: 'Browser module: page.isVisible(selector[, options]) method'
...
-| options.timeout | number  | `30000` | Maximum time in milliseconds. Pass `0` to disable the timeout. Default is overridden by the `setDefaultTimeout` option on [BrowserContext](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/browser/browsercontext/) or [Page](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/browser/page/). |
+| Parameter      | Type    | Default | Description                                                                                                                                                |
+| -------------- | ------- | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| selector       | string  | `''`    | A selector to search for an element. If there are multiple elements satisfying the selector, the first will be used.                                       |
```
ka3de, line 20:
```suggestion
| selector       | string  | `''`    | A selector to search for an element. If there are multiple elements satisfying the selector, the first one will be used.                                       |
```

docs/sources/next/javascript-api/k6-experimental/browser/page/isvisible.md
```diff
@@ -9,18 +9,17 @@ excerpt: 'Browser module: page.isVisible(selector[, options]) method'
...
+| options        | object  | `null`  |                                                                                                                                                            |
+| options.strict | boolean | `false` | When `true`, the call requires selector to resolve to a single element. If given selector resolves to more than one element, the call throws an exception. |
```
ka3de, line 22:
```suggestion
| options.strict | boolean | `false` | When `true`, the call requires the selector to resolve to a single element. If the given selector resolves to more than one element, the call throws an exception. |
```

docs/sources/next/javascript-api/k6-experimental/browser/elementhandle/isvisible.md
```diff
@@ -0,0 +1,48 @@
...
+
+# isVisible()
+
+Checks if the element is `visible`.
+
+### Returns
+
+| Type | Description                                       |
+| ---- | ------------------------------------------------- |
+| bool | `true` if the element is `visible`, else `false`. |
```
ka3de, line 14: Small suggestion, feel free to discard it if you think the current form is OK:
```suggestion
| bool | `true` if the element is `visible`, `false` otherwise. |
```

docs/sources/next/javascript-api/k6-experimental/browser/elementhandle/ishidden.md
```diff
@@ -0,0 +1,48 @@
...
+
+# isHidden()
+
+Checks if the element is `hidden`.
+
+### Returns
+
+| Type | Description                                      |
+| ---- | ------------------------------------------------ |
+| bool | `true` if the element is `hidden`, else `false`. |
```
ka3de, line 14: Small suggestion, feel free to discard it if you think the current form is OK:
```suggestion
| bool | `true` if the element is `hidden`, `false` otherwise. |
```

ka3de (commented):
Overall LGTM, just some small suggestions, but feel free to discard them if you think the current version is OK, these are no blockers by any means.

Also, I have a comment, maybe more for @heitortsergent:
Now we have a mix of methods for `elementHandle` and `frame`, some we do have our own specific documentation page, and some are just links to Playwright's method, as they are equivalent. This makes it look a little bit weird in the navigation sidebar. See:
Screenshot from 2023-12-21 08-21-15

The `elementHandle` has many methods listed in the table, on the right side doc, but on the left navigation sidebar it only lists the ones that we have documentation for. The same happens for `frame`. I guess this is OK and expected until we progressively add our own specific docs for each method, but it seemed a little bit weird, as it might look like it only has these two methods which are documented.

inancgumus (approved):
LGTM after Daniel's suggestions.

docs/sources/next/javascript-api/k6-experimental/browser/frame/ishidden.md
```diff
@@ -0,0 +1,62 @@
+---
+title: 'isHidden(selector[, options])'
+excerpt: 'Browser module: frame.isHidden(selector[, options]) method'
+---
+
+# isHidden(selector[, options])
+
+{{% admonition type="warning" %}}
+
+Use locator-based [`locator.isHidden([options])`](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/browser/locator/ishidden/) instead.
```
heitortsergent, line 10: Just out of curiosity, is there a specific reason why we recommend using the locator-based method?

heitortsergent (approved):
Just left one comment with a question, LGTM! :shipit: 

@ka3de that's a good point about the left-hand navigation only including a few methods, that can be confusing... One thing we could do is to fill out the section with pages for all the methods, even if the content in those pages is just linking to the playwright docs. But I'd also be curious if any users find it confusing, or think we don't might not support more methods if they only look at the left-hand navigation.

docs/sources/next/javascript-api/k6-experimental/browser/frame/ishidden.md
ankur22: `Locators` are seen to be better since they will auto wait and retry the search for the element with the given `selector`. This means that we can set a locator, navigate to a page, and then still work with the locator after the navigation occurs (as long as the element is actually on the new page). Whereas with the `frame` APIs, once a navigation occurs, we cannot use the old reference to the element that was setup before the navigation. In a nutshell, `locators` are safer.

heitortsergent: Ah, that does seem better, thanks for the explanation! 🙇

inancgumus: We explain locators with a Page Object Model example [here](https://k6.io/docs/using-k6-browser/recommended-practices/page-object-model/). We might expand on it.

PR #1507, opened 2024-02-12
chore: add redirect from stress testing to grafana page

heitortsergent:
## What?

Add a redirect from the stress testing page to the new Grafana blog, to test impact on SEO over canonical.


heitortsergent:
@ppcano I chatted with @AnneB-SEO2 today about the pillar content project, and some of the results we're seeing, and we wanted to test out if adding a redirect instead of the canonical that we currently have would impact it. We thought we'd test out on one page before adding the redirect to the other Load test type pages.

What do you think?

ppcano:
@heitortsergent, 

If I recall well, you have to remove the page in order the redirect to take effect. You can test this in `staging`.

heitortsergent:
@ppcano you're right, I deleted the page and I think it should work now.

I was going to just do a test deploy first with the Stress testing page first to make sure it's working, and then set up the rest of the pages in that section, do you think that's ok?

The one thing I'm not sure about is we'll lose that section of the left-hand side menu.

ppcano:
> The one thing I'm not sure about is we'll lose that section of the left-hand side menu.

@heitortsergent Yes, I think you will lose it. You can see this running the docs locally. 

If you want to include it on the sidebar, you can have an empty `md` page and  set the `redirect` frontmatter property. See [this](https://raw.githubusercontent.com/grafana/k6-docs/main/src/data/markdown/docs/05%20Examples/02%20Tutorials/03%20k6-learn.md) as an example.

heitortsergent:
Thanks @ppcano! I'll merge this and see if that helps us with the SEO before I do the rest. 🙏

PR #1515, opened 2024-02-27
chore: remove null return value from page.$$ docs

heitortsergent:
## What?

The browser module `page.$$` method returns an empty array when it can't find any elements, not a `null` value.

## Checklist

Please fill in this template:
- [x] I have reflected my changes in the relevant (*e.g.*  when correcting a documentation error) folders of the previous k6 versions of the documentation.

### DIFF
--- a/docs/sources/next/javascript-api/k6-experimental/browser/page/doubledollar.md
+++ b/docs/sources/next/javascript-api/k6-experimental/browser/page/doubledollar.md
@@ -16,9 +16,9 @@ The method finds all elements matching the specified selector within the page. I
 
 ### Returns
 
-| Type                                                                                                                      | Description                                                                                    |
-| ------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------- |
-| null \| [ElementHandle](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/k6-experimental/browser/elementhandle/)[] | Returns an array of `ElementHandle` when multiple elements are found. Else, it returns `null`. |
+| Type                                                                                                              | Description                       |
+| ----------------------------------------------------------------------------------------------------------------- | --------------------------------- |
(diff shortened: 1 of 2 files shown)


inancgumus (approved):
👍

ankur22 (commented):
I'm not 100% sure about this change. It does seem that in some cases we can get a `null` response due to [this part of the code](https://github.com/grafana/xk6-browser/blob/69e1f33417ebff0f9a0aee27de44b65236102eb3/common/element_handle.go#L1082-L1085). What was the motivation for the change? Is there a test script that exhibits this behaviour where it returns an empty array instead of null?

inancgumus:
> I'm not 100% sure about this change. It does seem that in some cases we can get a `null` response due to [this part of the code](https://github.com/grafana/xk6-browser/blob/69e1f33417ebff0f9a0aee27de44b65236102eb3/common/element_handle.go#L1082-L1085). What was the motivation for the change? Is there a test script that exhibits this behaviour where it returns an empty array instead of null?

But we still return an empty array: https://github.com/grafana/xk6-browser/blob/69e1f33417ebff0f9a0aee27de44b65236102eb3/browser/mapping.go#L332-L337.

ankur22 (approved):

heitortsergent:
@ankur22 this was the sample script that @tom-miseur shared that shows the behavior:

```javascript
import { browser } from 'k6/experimental/browser'

export const options = {
  scenarios: {
    scenario1: {
      executor: 'shared-iterations',
      vus: 1,
      iterations: 1,
      options: {
        browser: {
          type: 'chromium',
        }
      }
    }
  }
}

export default async function () {
  const context = browser.newContext();
  const page = context.newPage();

  await page.goto('https://test-api.k6.io/public/crocodiles/'); // implicitly waits for the page to load

  const prettyElements = page.$$('.prettyprint')

  console.log('.prettyprint elements found: ' + prettyElements.length);

  const nonElements = page.$$('.nonexistent-element')

  console.log('.nonexistent elements found: ' + nonElements.length);
}
```

PR #1517, opened 2024-02-28
chore: fix aliases

heitortsergent:
## What?

Fix a few aliases that were added on #1513.

## Checklist

Please fill in this template:


heitortsergent:
Tagging @jdbaldry this time to make sure the aliases look ok. 😅

jdbaldry (approved):
I think each alias should have a comment explaining the absolute URL it is trying to redirect because it makes it a lot easier when they need to be updated in the future.

Aliases are working though 👍

heitortsergent:
I added the comment with the absolute URL path next to the aliases, thanks for the review @jdbaldry! 🙏

PR #1589, opened 2024-05-16
chore: fix ESLint errors

heitortsergent:
## What?

Update code samples with ESLint errors by making code samples valid or disabling ESLint rules.

## Checklist


heitortsergent:
@codebien this should take care of the errors you saw when creating the v0.51.x release. 🙇

codebien (commented):
I have the feeling that in most of the examples where we are going to use  `<!-- eslint-disable no-undef  -->`, instead, we could fix it using an import statement. Should it be a preferred practice?

If we prefer to continue with the current essential approach, should we consider switching off this rule for the entire repository? I guess it will be a very popular warning.

heitortsergent:
@codebien that's a fair point. I think for tutorials, it makes sense to have snippets that are shorter and don't have all the required imports or variables. Right now, it's only a few files that require us to disable the rule, so I think it's ok to keep it this way, and it would unblock us from seeing the errors in this next release.

I want to spend a bit more time looking at the overall ESLint configuration for the repo and seeing if there's anything else we can change to make sure we have valid examples, while keeping the tutorials clear and concise.

codebien (approved):

PR #1645, opened 2024-07-09
Document HTTP headers structure

olegbespalov:
## What?

This change documents the fact that headers are arrays and should address https://github.com/grafana/k6/issues/2559

## Checklist

- [x] I have reflected my changes in the relevant folders of the two previous k6 versions of the documentation (if still applicable to previous versions).

## Related PR(s)/Issue(s)

Closes: https://github.com/grafana/k6/issues/2559


codebien (approved):
LGTM with Heitor's suggestions applied.

docs/sources/next/javascript-api/k6-http/response/_index.md
```diff
@@ -18,7 +16,7 @@ Note that in the case of redirects, all the information in the Response object w
...
+| `Response.headers`                                                                                                              | object   | Key-value pairs representing all HTTP headers sent by the server. Note that the header names are in [canonical form](https://pkg.go.dev/net/http#CanonicalHeaderKey); for example, if the server responds with "accept-encoding", the key will be "Accept-Encoding". When requesting a header by a specific name, an array of strings is returned since the header can hold multiple values. In the case of a single-valued header, you can access it by index, such as `Response.headers["my_key"][0]`.                                                                                                  |
```
mstoykov, line 19:
```suggestion
| `Response.headers`                                                                                                              | object   | Key-value pairs representing all HTTP headers sent by the server. Note that the header names are in [canonical form](https://pkg.go.dev/net/http#CanonicalHeaderKey); for example, if the server responds with "accept-encoding", the key will be "Accept-Encoding". When requesting a header by a specific name, an array of strings is returned since the header can have multiple values. So you can access each value by its index - for the first value (or for single-valued header) that will be `Response.headers["my_key"][0]`.          

                                                                                                                                                                                |
```

> In the case of a single-valued header, you can access it by index, 

sounds like otherwise you can not - as if they were multiple values, you need to do something else.

mstoykov (approved):

PR #1688, opened 2024-08-08
Update after native ESM support

mstoykov:
## What?

Update after ESM is now natively supported.

## Checklist

- [ ] I have used a meaningful title for the PR.
- [ ] I have described the changes I've made in the "What?" section above.
- [ ] I have performed a self-review of my changes.

- [ ] I have made my changes in the `docs/sources/next` folder of the documentation.
- [ ] I have reflected my changes in the `docs/sources/v{most_recent_release}` folder of the documentation.

- [ ] I have made my changes in the `docs/sources/next` folder of the documentation.

### DIFF
--- a/docs/sources/next/using-k6/javascript-typescript-compatibility-mode.md
+++ b/docs/sources/next/using-k6/javascript-typescript-compatibility-mode.md
@@ -9,24 +9,20 @@ weight: 19
 
 # JavaScript and TypeScript compatibility mode
 
-You can write k6 scripts in various ECMAScript versions:
+You can write k6 tests in various ECMAScript versions:
 
 - ES6+ JavaScript with ES modules (ESM).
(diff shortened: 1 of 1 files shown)


ppcano:
LGTM, I share some thoughts for later consideration:

1 - The guides are versioned, so I am unsure if  it is desired to mention specific version changes `After v0.53.0...`.
On the other hand, it may be desired b/c users tend to read the latest version. This is why we decided time ago to remove these references after two releases.

2 - `--compatibility-mode=base` is now a very rare case. I think we could drop the documentation about this option, and leave the versioned docs for older references.

3 - About `--compatibility-mode=experimental_enhanced`. I shared this before and don't know exactly the future plan, but we could rename this option to just `typescript`... or `mode=typescript`

ppcano (approved):

joanlopez (approved):

PR #1702, opened 2024-08-23
Fix invalid option `enablecolor`

stas00:
Fix the doc `s/enableColor/enableColors/`


heitortsergent (approved):
Thanks for catching that error and for opening a PR @stas00! 🙇

PR #1712, opened 2024-08-30
Typo: Browser Recorder should be DevTools Recorder

allansson:
Just a small typo I noticed.

EDIT: Also fixed broken links.


heitortsergent (approved):
Thanks @allansson! 🙇

PR #1732, opened 2024-09-19
Document k6 cloud run --local-execution

oleiade:
## What?

We have introduced the `k6 cloud run --local-execution` flag to supersede `k6 run -o cloud`, which is now considered (but not communicated as) deprecated.

I have made the changes I found obvious, let me know if you think we should go farther, and modify further docs, or add more info 🙇🏻 

## Checklist

### DIFF
--- a/docs/sources/next/results-output/real-time/cloud.md
+++ b/docs/sources/next/results-output/real-time/cloud.md
@@ -12,11 +12,11 @@ When streaming the results to the cloud, the machine - where you execute the k6 
 
 ## Streaming results vs. running on cloud servers
 
-Don't confuse `k6 run --out cloud script.js` (what this page is about) with `k6 cloud run script.js`.
+Don't confuse `k6 cloud run --local-execution script.js` (what this page is about) with `k6 cloud run script.js`.
 
 Fundamentally the difference is the machine that the test runs on:
 
-- `k6 run --out cloud` runs k6 locally and streams the results to the cloud.
+- `k6 cloud run --local-execution` runs k6 locally and streams the results to the cloud.
 - `k6 cloud run`, on the other hand, uploads your script to the cloud solution and runs the test on the cloud infrastructure. In this case you'll only see status updates in your CLI.
 
 In all cases you'll be able to see your test results at [k6 Cloud](https://app.k6.io) or [Grafana Cloud](https://grafana.com/products/cloud/).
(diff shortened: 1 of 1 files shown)


docs/sources/next/results-output/real-time/cloud.md
```diff
@@ -102,6 +102,20 @@ data, setting the run state to `Timed out`. You can still analyze a timed out
...
+command will also upload the test archive k6 produces from your test script and resources to our cloud service.
+
+This behavior can be deactivated by adding the `--no-archive-upload` option to your
+command: `k6 cloud run --local-execution --no-archive-upload script.js`.
+
+{{< /admonition >}}
+
+{{< admonition type="note" >}}
+
+{{< /admonition >}}
```
heitortsergent, line 117:
```suggestion
{{< admonition type="caution" >}}

The `k6 cloud run --local-execution script.js` command uploads the test archive k6 produces from your test script and resources to Grafana Cloud.

This behavior can be deactivated by adding the `--no-archive-upload` option to your
command: `k6 cloud run --local-execution --no-archive-upload script.js`.

{{< /admonition >}}
```

Is there any reason why a user might want to disable this or keep it enabled? I think we should include it here as well if we want to use the `caution` admonition type. Otherwise, we can just switch it to `note`.

heitortsergent (commented):
Looks good! I left a small comment about the admonition, and I found another instance of the command on this file that we should probably change as well: https://github.com/grafana/k6-docs/blob/document_local_execution/docs/sources/next/set-up/set-up-distributed-k6/usage/k6-operator-to-gck6.md.

heitortsergent:
I created a PR to update the Grafana Cloud k6 docs as well: https://github.com/grafana/website/pull/21806

docs/sources/next/results-output/real-time/cloud.md
oleiade: A user might want to use the `--no-archive-upload` option primarily for privacy reasons. A user might also use the `k6 cloud run—-local-execution` mode specifically to keep all their context, assets, and resources local while still benefiting from the Grafana Cloud k6 experience.

It's activated by default because uploading the archive to the cloud, like with the `cloud run` command by default, gives the Grafana Cloud k6 much more context about the execution and leads to a better UX now and in the future.

I'll commit your suggestion as-is; cheers for that, and I'll keep the "caution" state as I think users who care about that do **really** care about it, and we should ensure they don't miss that bit.

oleiade:
Regarding the operator changes, I would prefer to not touch it for the time being and reasses later if the operator does support the new mode of operation 🙇🏻

docs/sources/next/results-output/real-time/cloud.md
heitortsergent: That makes sense, thanks for the explanation! 🙇

heitortsergent (approved):
LGTM! :shipit:

PR #1734, opened 2024-09-23
Document the `k6 cloud upload` command

oleiade:
## What?

This PR documents the newly introduced `k6 cloud upload` command. We also document the "Upload Only" option as deprecated.

I was really unsure where the right place to document this was, please feel free to propose a better place or even move it around. 

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/pull/3949/files#r1767831917

### DIFF
--- a/docs/sources/next/using-k6/k6-options/reference.md
+++ b/docs/sources/next/using-k6/k6-options/reference.md
@@ -1476,7 +1476,13 @@ The possible keys with their meanings and default values:
 | proto               | the protocol to use when connecting with the traces backend                                                            | `grpc`                  |
 | header.`headerName` | adds an additional HTTP header with the provided header name and value to each HTTP request made to the traces backend | N/A                     |
 
-## Upload Only
+## Upload Only (deprecated)
+
+{{< admonition type="caution" >}}
+
+The "Upload Only" option is deprecated and will be removed in a future release. Please use the `k6 cloud upload` command instead.
+
+{{< /admonition >}}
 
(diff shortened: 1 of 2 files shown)


docs/sources/next/using-k6/k6-options/reference.md
```diff
@@ -1476,7 +1476,13 @@ The possible keys with their meanings and default values:
 | proto               | the protocol to use when connecting with the traces backend                                                            | `grpc`                  |
 | header.`headerName` | adds an additional HTTP header with the provided header name and value to each HTTP request made to the traces backend | N/A                     |
 
-## Upload Only
+## Upload Only (deprecated)
+
+{{< admonition type="caution" >}}
+
+The "Upload Only" option is deprecated and will be removed in a future release. Please use the `k6 cloud upload` command instead.
```
heitortsergent, line 1483:
```suggestion
## Upload only (deprecated)

{{< admonition type="caution" >}}

The "Upload only" option is deprecated and will be removed in a future release. Use the `k6 cloud upload` command instead.
```

heitortsergent (approved):
Left one small comment there, I think this works for now! I'm wondering if we might need a new reference page for the k6 CLI commands... What do you think?

docs/sources/next/using-k6/k6-options/reference.md
oleiade: I think it would be worth it indeed, especially as some of those have very specific usage or behaviors that we currently don't really document here.

PR #1755, opened 2024-09-30
Fix quoting in heredoc

jdbaldry:
I wasn't really thinking when I added the quotes here. With the quotes, `ed` is trying to open a file that has quotes in the name.


szkiba (approved):
LGTM

szkiba:
@jdbaldry 

> I wasn't really thinking when I added the quotes here. With the quotes, ed is trying to open a file that has quotes in the name.

and I wasn't really thinking when I accepted your suggestion :) the lesson is that I should have tried it

PR #1803, opened 2024-11-07
Update group page to include link to utils check function

heitortsergent:
## What?

The group link is referenced from a k6 error message when using the check function with async. This PR includes a link to where users can find the correct information about using the jslib utils check function.

## Checklist

- [ ] I have reflected my changes in the `docs/sources/v{most_recent_release}` folder of the documentation.

### DIFF
--- a/docs/sources/k6/next/javascript-api/k6/group.md
+++ b/docs/sources/k6/next/javascript-api/k6/group.md
@@ -6,6 +6,12 @@ description: 'Runs code inside a group. Used to organize results in a test.'
 
 # group( name, fn )
 
+{{< admonition type="note" >}}
+
+For details about using the `check()` function with async values, refer to the jslib utils [check](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/jslib/utils/check/).
+
+{{< /admonition >}}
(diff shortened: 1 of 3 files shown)


docs/sources/k6/next/javascript-api/k6/group.md
```diff
@@ -6,6 +6,12 @@ description: 'Runs code inside a group. Used to organize results in a test.'
 
 # group( name, fn )
 
+{{< admonition type="note" >}}
+
+For details about using the `check()` function with async values, refer to the jslib utils [check](https://grafana.com/docs/k6/<K6_VERSION>/javascript-api/jslib/utils/check/).
```
inancgumus, line 11: Would it be better to link to the original `check` function here as well?

inancgumus (approved):
LGTM 👍

docs/sources/k6/next/javascript-api/k6/group.md
heitortsergent: There's a link to the original check function on the jslib utils check page, so I think just having the one link here is better. 🤓

PR #1781, opened 2024-10-23
Split publishing workflows

jdbaldry:
## What?

Split k6 and k6-studio documentation so they can be published independently.

Because I've moved all the k6-docs, there's a big patch. Look at the individual commits in https://github.com/grafana/k6-docs/pull/1781/commits. 

Self-review included running `make docs` and browsing to both http://localhost:3002/docs/k6/ and http://localhost:3002/docs/k6-studio/.

## Checklist

- [x] I have used a meaningful title for the PR.
- [x] I have described the changes I've made in the "What?" section above.
- [x] I have performed a self-review of my changes.
- [x] I have run the `npm start` command locally and verified that the changes look good.

## Related PR(s)/Issue(s)

Closes https://github.com/grafana/website/issues/22312

Signed-off-by: Jack Baldry <jack.baldry@grafana.com>

PR #1816, opened 2024-11-27
Fix path to extension registry output file

jdbaldry:
It was moved in https://github.com/grafana/k6-docs/pull/1781 and the script has been failing in https://github.com/grafana/k6-docs/actions/workflows/extension-registry-changed.yml since then.

Thanks to @heitortsergent for spotting the failing workflow.


heitortsergent (approved):
Thanks for fixing this @jdbaldry! 🙇

PR #1828, opened 2024-12-23
Add md-k6 script + new workflow

federicotdn:
## What?

Adds a new script, `md-k6.py`, that is run via the `run-code-blocks.yml` Workflow. The workflow is run only when .md files are modified in the `k6/next/**` directory via a PR. The script then takes the paths of all .md files that were modified in the PR (in comparison to `main`) and then run all JS codeblocks within those files. If any of the scripts causes k6 to exit with 1, or logs any type of error, then the workflow will fail.

It is possible to skip the running of the script by adding a magic comment on top of the script: `<!--md-k6:skip-->`.

## Checklist

### DIFF
--- a/CONTRIBUTING/README.md
+++ b/CONTRIBUTING/README.md
@@ -171,7 +171,7 @@ export default async function () {
 
 In addition to linting code snippets, we also run the snippets using k6 OSS. This is done automatically on PRs, only for Markdown files that have been changed in the `docs/sources/next` directory when compared to `main`. See the `scripts/md-k6.py` script for details on how this works internally.
 
-Code snippets are run using the `-w` k6 OSS flag. If the code snippet causes k6 to exit with a nonzero status, then the script (and therefore the workflow) will fail. If any error is logged by k6 (for example, because an exception was raised), this will also fail the execution.
+Code snippets are run using the `-w` k6 OSS flag. If the code snippet causes k6 to exit with a nonzero status, then the script (and, therefore, the workflow) will fail. If any error is logged by k6, for example, because an exception was raised, this will also fail the execution.
 
 You can control the behaviour of `md-k6.py` via magic `md-k6` HTML comments placed above the code snippets. The format is the following:
 
--- a/CONTRIBUTING/README.md
+++ b/CONTRIBUTING/README.md
@@ -16,6 +16,9 @@ When you contribute to the docs, it helps to know how things work.
     - [Use the `apply-patch` script](#use-the-apply-patch-script)
     - [Style Guides](#style-guides)
     - [Shortcodes](#shortcodes)
+    - [Shortcodes](#shortcodes)
+    - [Code snippets and ESLint](#code-snippets-and-eslint)
+    - [Code snippets evaluation](#code-snippets-evaluation)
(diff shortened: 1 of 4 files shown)


heitortsergent:
This is awesome, thanks so much for creating this @federicotdn! 🎉 

Tagging @jdbaldry here as well if you'd like to take a look 🤓 . I'm not sure if this could be applicable to other repos, but what do you think of having this in the k6 repo?

federicotdn:
I would rather keep it here since this is only documentation-related, but also I have a feeling that adding it to the main k6 repo would take more time in terms of reviewing.

CONTRIBUTING/README.md
```diff
@@ -164,6 +167,49 @@ export default async function () {
 ```
 ````
 
+### Code snippets evaluation
+
+In addition to linting code snippets, we also actually run the snippets using k6 OSS. This is done automatically for all newly opened PRs, only for Markdown files that have been changed (when compared to `main`) in the `docs/sources/next` directory. See the `scripts/md-k6.py` script for details on how this works internally.
```
heitortsergent, line 172:
```suggestion
In addition to linting code snippets, we also run the snippets using k6 OSS. This is done automatically on PRs, only for Markdown files that have been changed in the `docs/sources/next` directory when compared to `main`. See the `scripts/md-k6.py` script for details on how this works internally.
```

CONTRIBUTING/README.md
```diff
@@ -164,6 +167,49 @@ export default async function () {
...
+
+Code snippets are run using the `-w` k6 OSS flag. If the code snippet causes k6 to exit with a nonzero status, then the script (and therefore the workflow) will fail. If any error is logged by k6 (for example, because an exception was raised), this will also fail the execution.
```
heitortsergent, line 174:
```suggestion
Code snippets are run using the `-w` k6 OSS flag. If the code snippet causes k6 to exit with a nonzero status, then the script (and, therefore, the workflow) will fail. If any error is logged by k6, for example, because an exception was raised, this will also fail the execution.
```

heitortsergent (approved):
Made some small edits to the README, thanks for including a new section there! 🙇 

I tested this locally running the script on a few files, and it works as expected 🤓 . I think this will be really helpful to keep the doc samples working and up to date!

jdbaldry:
Cool script, thanks for making me aware of it. It's a neat solution to ensuring the validity of code blocks. Another mechanism, used by the Mimir team, is to write tests using the code or configuration files directly and then embed that code or configuration in the documentation using something like `embedmd`.

federicotdn:
@jdbaldry thanks! And yes, that sounds like a good idea with `embedmd`. Maybe we could do that in the future, it would require some initial effort for going through all the files (unless we do it progressively). After that we could delete this script and run the k6 .js files directly.

jdbaldry:
> @jdbaldry thanks! And yes, that sounds like a good idea with `embedmd`. Maybe we could do that in the future, it would require some initial effort for going through all the files (unless we do it progressively). After that we could delete this script and run the k6 .js files directly.

No need to migrate if this is working for you :)

PR #1924, opened 2025-04-08
Update the "Using k6 browser" section

heitortsergent:
## What?

Updates to the "Using k6 browser" section.

## Checklist

- [x] I have reflected my changes in the `docs/sources/k6/v{most_recent_release}` folder of the documentation.
- [x] I have reflected my changes in the relevant folders of the two previous k6 versions of the documentation (if still applicable to previous versions).


docs/sources/k6/next/using-k6-browser/how-to-write-browser-tests/asynchronous-operations.md
```diff
@@ -0,0 +1,63 @@
...
+weight: 100
+---
+
+# Asynchronous operations
+
+Since many browser operations happen asynchronously, and to follow the Playwright API more closely, we are working on migrating most of the browser module methods to be asynchronous as well.
+
+At the moment, methods such as `page.goto()`, `page.waitForNavigation()` and `Element.click()` return [JavaScript promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises), and scripts must be written to handle this properly.
+
+To avoid timing errors or other race conditions in your script, if you have actions that load up a different page, you need to make sure that you wait for that action to finish before continuing.
```
oleiade, line 13: I believe we're mostly done with this move, but I might be wrong, might be worth checking with the Browser folks 😇

oleiade (approved):
Left a tiny clarification question, but it's good to merge-as-is 👏🏻

docs/sources/k6/next/using-k6-browser/how-to-write-browser-tests/asynchronous-operations.md
```diff
@@ -0,0 +1,63 @@
+---
+title: Asynchronous operations
+description: 'Learn how the k6 browser module uses asynchronous operations.'
+weight: 100
+---
+
+# Asynchronous operations
+
+Since many browser operations happen asynchronously, and to follow the Playwright API more closely, we are working on migrating most of the browser module methods to be asynchronous as well.
```
ankur22: This has been completed. Most of the browser module APIs are now async and require working with `await` keyword.

The reason we made them async was:
1. To be more aligned with PW;
2. In JS to avoid blocking the single event loop, any long running processes or IO operations should be made async so that they can be handled in the background. Most of the browser module APIs are IO operations since they communicate with chromium over a websocket.

There were other reasons, but these two are the most important.

EDIT: More details in the release notes before we made the change: https://github.com/grafana/k6/releases/tag/v0.51.0

docs/sources/k6/next/using-k6-browser/how-to-write-browser-tests/asynchronous-operations.md
```diff
@@ -0,0 +1,63 @@
+---
+title: Asynchronous operations
+description: 'Learn how the k6 browser module uses asynchronous operations.'
+weight: 100
+---
+
+# Asynchronous operations
```
ankur22, line 7: I think for this file we need to structure it a bit differently since we've made the move to async. Here are my thoughts on what might be useful for the reader (i'm assuming the reader isn't so proficient with JS):

Super rough suggestion:

---

Most methods in the browser module return [JavaScript promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises), and scripts must be written to handle this properly. This usually means using the `await` keyword to wait for the async operation to complete.

It may seem counterintuitive to have async APIs when you are then required to `await` on them, why not make them synchronous? The reasons are:

1. JS is single threaded with a single event loop, we want to avoid blocking the thread and event loop with long running processes or IO based tasks.
2. To align with Playwright.
3. etc. (Other reasons outlined in https://github.com/grafana/k6/releases/tag/v0.51.0)

Let's take a look at a simple example:

```js
const page = await browser.newPage();

await page.goto('https://quickpizza.grafana.com/');

const locator = page.locator('button[name="pizza-please"]');

await locator.click();
```

All the API calls that perform some action against chromium are async, and in the example above we've placed an `await` on them to allow for the action to complete before proceeding to the next step. The `page.locator` API call doesn't call out to chromium, so this is a synchronous API and we don't need to place an `await` on that API call -- no problems will occur even if you did place an `await` when calling `page.locator` since the JS runtime will handle that automatically realising that there is no promise to handle and just return the return value straight away.

If we didn't add an `await` on the async APIs then the iteration is likely to end before the test script completes and we will also likely see errors like "Uncaught (in promise) TypeError: Object has no member 'goto'", which is because the page object from `browser.newPage()` without an `await` is actually a `JavaScript promise`. Give it a try and see the errors for yourself:

```js
const page = browser.newPage();

page.goto('https://quickpizza.grafana.com/');

const locator = page.locator('button[name="pizza-please"]');

locator.click();
```

## Promise.all

// Now we can talk about the Promise.all and just add the paragraph that you already have

## waiting for specific elements

We also would encourage the use of `loactor.waitFor` where possible. When you navigate to a website, there is usually one or more elements that are of importance. Once those elements load we should be able to progress to the next step of the test. Think about it from a search perspective:

1. We navigate to the search site;
2. wait for the search bad and submit button to appear;
3. fill in the search bar with the query;
4. click on the button;
5. wait for the search results.

We should be able to do these actions like so:

```js
await page.goto('https://my-search-engine.com');

const searchBar = page.locator('.search-bar');
const submitBUtton = page.locator('.submit-button');

await searchBar.waitFor();
await submitButton.waitFor();

await searchBar.fill('k6');
await submitButton.click();

const searchResults = page.locator('.search-results-table');

await searchResults.waitFor();
```

This avoids the use of `Promise.all` which can be confusing to work with, and instead makes the script a bit more easy to follow.

---

WDYT?

heitortsergent: Sorry for the delay in reviewing this @ankur22. 🙇 

I tried to incorporate the changes you suggested into the page. I made some small edits, but one thing I did was shift the first section explaining why the browser module works with asynchronous APIs to the bottom of the page.

Can you take another look at the page and see if the changes I made make sense? Also @oleiade if you wouldn't mind taking a look as well. 🙏

docs/sources/k6/next/using-k6-browser/how-to-write-browser-tests/asynchronous-operations.md
```diff
@@ -0,0 +1,136 @@
...
+
+Most methods in the browser module return JavaScript promises, and k6 scripts must be written to handle this properly. This usually means using the `await` keyword to wait for the async operation to complete.
```
oleiade: We could maybe link `Javascript Promises` to the "why the browser apis are async at the bottom of the page? (don't know how to do that in github's pr review interface)

docs/sources/k6/next/using-k6-browser/how-to-write-browser-tests/asynchronous-operations.md
```diff
@@ -0,0 +1,136 @@
...
+
+There are two recommended methods to handle async operations in browser scripts: using `Promise.all`, or using the `waitFor` method.
+
```
oleiade, line 26: I think it would be worth, to bring back a basic example of a browser test to showcase the `await` keyword, before we move on to the advanced use-case below?

docs/sources/k6/next/using-k6-browser/how-to-write-browser-tests/asynchronous-operations.md
```diff
@@ -6,9 +6,11 @@ weight: 100
 
 # Asynchronous operations
 
-Since many browser operations happen asynchronously, and to follow the Playwright API more closely, we are working on migrating most of the browser module methods to be asynchronous as well.
+Most methods in the browser module return JavaScript promises, and k6 scripts must be written to handle this properly. This usually means using the `await` keyword to wait for the async operation to complete.
 
-At the moment, methods such as `page.goto()`, `page.waitForNavigation()` and `Element.click()` return [JavaScript promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises), and scripts must be written to handle this properly.
+There are two recommended methods to handle async operations in browser scripts: using `Promise.all`, or using the `waitFor` method.
```
ankur22: These two functions/methods are mainly used to wait for page navigations to occur.

docs/sources/k6/next/using-k6-browser/how-to-write-browser-tests/asynchronous-operations.md
```diff
@@ -61,3 +63,74 @@ export default async function () {
...
+API calls that interact with Chromium are asynchronous and require `await` to ensure completion before proceeding. Synchronous APIs, such as `page.locator`, do not require `await`, but using it does not cause issues since the JavaScript runtime will simply return the value immediately.
+
+If you don't add `await` on asynchronous APIs, it can cause the script to finish before the test completes, resulting in errors like `"Uncaught (in promise) TypeError: Object has no member 'goto'"`. That can happen because the page object from `browser.newPage()` without an `await` is actually a JavaScript promise. You can try and see the error using the following code snippet:
+
+<!-- eslint-skip -->
+
+```js
+const page = browser.newPage();
+
+page.goto('https://quickpizza.grafana.com/');
```
ankur22:
```suggestion
page.goto('https://quickpizza.grafana.com/'); // An error should occur since we're not using await in the line above.
```

Or something along those lines.

docs/sources/k6/next/using-k6-browser/how-to-write-browser-tests/asynchronous-operations.md
heitortsergent: Would it be better to reword this @ankur22? I want it to be accurate. 🤓 

Maybe:

```suggestion
In addition to using `await`, another important part of writing k6 browser tests is handling page navigations. There are two recommended methods for doing that: using `Promise.all` or using the `waitFor` method.
```

docs/sources/k6/next/using-k6-browser/how-to-write-browser-tests/asynchronous-operations.md
heitortsergent: I added a link to that section. 👍

docs/sources/k6/next/using-k6-browser/how-to-write-browser-tests/asynchronous-operations.md
heitortsergent: I agree. I added the small example from Ankur's comment back in after the first paragraph.

docs/sources/k6/next/using-k6-browser/how-to-write-browser-tests/asynchronous-operations.md
ankur22: Yeah, that's better. Sorry, i should've suggested something.

heitortsergent: No worries, I appreciate you reviewing the PR! 🙇

PR #1944, opened 2025-05-03
Set javascript syntax highlight on `http-debugging.md`

sh-cho:
## What?

Some code block in HTTP debugging page is highlighted as bash, but it's actually javascript code.

So changed syntax highlight with javascript.

## Checklist

- I have updated all versions

## Additional Question
I noticed that this HTTP debugging page is unlisted.
Is it intended?

### DIFF
--- a/docs/sources/k6/next/using-k6/http-debugging.md
+++ b/docs/sources/k6/next/using-k6/http-debugging.md
@@ -19,7 +19,7 @@ Given the following script:
 
 {{< code >}}
 
-```bash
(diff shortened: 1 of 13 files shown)


heitortsergent (approved):
Thanks for opening this PR @sh-cho!

I think this page was hidden because we had plans to update it, along with the blog link that was at the end of it. We'll keep it as is until we have time to tackle this. 🤓 

Thank you again for fixing this! 🙇

PR #1974, opened 2025-07-02
Update /docs/k6 redirect to latest

heitortsergent:
## What?

Update the /docs/k6 redirect to go to the latest version instead of next.

## Checklist

### DIFF
--- a/docs/sources/k6/_index.md
+++ b/docs/sources/k6/_index.md
@@ -3,7 +3,7 @@ title: Grafana k6 documentation
 menuTitle: Grafana k6
 description: The k6 documentation covers everything you need to know about k6 OSS, load testing, and performance testing.
 type: redirect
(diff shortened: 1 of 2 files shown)


heitortsergent:
@jdbaldry do you remember why this change was made? I can see that if I try to build this with the `redirectURL` pointing to `latest`, the docs don't render correctly... But I think it would be best to redirect users to latest if they just go to `/docs/k6`.

cc @robbymilo

robbymilo (approved):
Makes sense to me as this is the only project that does not redirect to latest.

heitortsergent:
Thanks @jdbaldry! Could you sign the CLA again so I can merge this? 😆 I think it's been asking everyone to resign it since we had to tweak it to make it work with the GitHub bot.

PR #1979, opened 2025-07-04
Document k6-jslib-aws v0.14.0

oleiade:
## What?

This pull request introduces updates to the documentation for the `k6` JavaScript library, focusing on enhancements to AWS service clients (`S3Client` and `SQSClient`). Key changes include the addition of new features, improved examples, and updates to existing functionality.

### Updates to `S3Client` documentation:

* **New documentation for `S3UploadedObject`:**
  - Added a new page describing the `S3UploadedObject` class, which represents metadata from S3 upload operations. This includes properties like checksums and object size, along with examples for single-part and multipart uploads.

* **Enhanced examples for `putObject` and `completeMultipartUpload`:**
  - Updated `putObject` to return an `S3UploadedObject` and added examples for accessing metadata such as ETag and checksums.
  - Updated `completeMultipartUpload` to return an `S3UploadedObject` and added examples for accessing metadata and verifying multipart uploads.

### Updates to `SQSClient` documentation:

* **Expanded functionality:**
  - Added new methods (`sendMessageBatch`, `receiveMessages`, and `deleteMessage`) to the `SQSClient` documentation, enabling batch message sending, message retrieval, and deletion.

* **New documentation for `deleteMessage`:**
  - Added a dedicated page for the `deleteMessage` method, including parameter descriptions, return type, and an example demonstrating message deletion after processing.

* **Improved example for queue operations:**
  - Enhanced the example to include receiving messages, processing them, and deleting them using the newly documented `deleteMessage` method.

### General updates:

* **Version bump for `JSLIB_AWS_VERSION`:**
  - Updated the AWS library version from `0.13.0` to `0.14.0` to reflect the latest changes.

## Checklist

-


docs/sources/k6/next/javascript-api/jslib/aws/SQSClient/_index.md
```diff
@@ -61,6 +66,20 @@ export default async function () {
 
   // Send message to test queue
   await sqs.sendMessage(testQueue, JSON.stringify({ value: '123' }));
+
```
joanlopez, line 69: Nit; should we also show a short example of how to call `sendMessageBatch`, for the sake of completeness?

joanlopez (commented):
Orthogonal to this PR, but all the changes in older versions of k6 triggered my mind: should we eventually also add versioning for jslibs docs? 

For instance, how a k6 users would recognize that these methods are part of v0.14 and not part of v0.13, or viceversa? 🤔 Although, maybe we just don't care that much for now.

joanlopez:
> How might we achieve and present that? What would it look like?

I don't feel very inspired, but if I eventually have any idea, I will tell you.

But being honest, I don't feel like I'm the person who should decide how the docs should look like for the end user nor how to organize the codebase that backs them 🤷🏻

jdbaldry:
> @jdbaldry or @robbymilo wondering if I can tag you in for your feedback here. Do you know if we have any other Grafana projects that are versioned, but also have another sub-project that's versioned as well?
> 
> In our case, we have the k6 version, and then this set of libraries that users can use, which have their own versions.

I think the answer depends on the relationships between the two. For example, GEM and Mimir have a one-to-one relationship, so they are both separately versioned projects with links between them. [Version substitution](https://grafana.com/docs/writers-toolkit/write/shortcodes/#about-version-substitution) is adequate to maintain that one-to-one relationship and it uses a front matter parameter similar to your JS lib version.

I'm not sure that multiple versions of one thing should be nested within multiple versions of another. It duplicates content and muddies the idea of a canonical page. If we want to start maintaining multiple versions of the library documentation, I think we should consider splitting it into its own project so there's a canonical source.

We'd then need to consider the matrix of relationships between k6 and the library and think about how best to serve the user the most useful information. We have a general stance of preferring latest and discouraging the use of older versions with the intent of encouraging users to upgrade to the latest version rather than stick with the old version.

oleiade:
I will merge as-is to keep things moving.

Regarding the jslibs I suggest we open a dedicated issue 🙇🏻

PR #1986, opened 2025-07-08
k6 browser: add locator.all

AgnesToulet:
## What?
Add documentation for the browser module new function `locator.all()` implemented in https://github.com/grafana/k6/pull/4899

## Checklist

## Related PR(s)/Issue(s)
https://github.com/grafana/k6/pull/4899


docs/sources/k6/next/javascript-api/k6-browser/locator/_index.md
```diff
@@ -12,12 +12,13 @@ The Locator API makes it easier to work with dynamically changing elements. Some
 - Helps with writing robust tests by finding an element even if the underlying frame navigates.
 - Makes it easier to work with dynamic web pages and SPAs built with Svelte, React, Vue, etc.
 - Enables the use of test abstractions like the Page Object Model (POM) pattern to simplify and organize tests.
-- `strict` mode is enabled for all `locator` methods, which means that if more than one element matches the given selector it will throw an error.
+- `strict` mode is enabled for some `locator` methods, which means that if more than one element matches the given selector it will throw an error.
```
AgnesToulet: I'm not sure if we should keep a list here for the methods that support matching multiple elements. This is not only about the new `all` function but also `count`, `nth`, `first` and `last`. It also makes sense that these functions are meant to work with multiple elements?

AgnesToulet:
I noticed that there is a check to run the code block and it fails because the change isn't merged yet. How does it work once this is merged? Does it take the latest released version or do we have a dev release that is updated on a regular basis? Or do I need to wait for the next release before merging this doc PR?

docs/sources/k6/next/javascript-api/k6-browser/locator/_index.md
heitortsergent: Do you think the list of supported methods would see a lot of changes in the future? If it's stable, I'd say let's list the methods here.

heitortsergent:
@AgnesToulet that is something we still need to fix with that workflow. 😅 

It's non-blocking, I do think it's okay to merge for an unreleased feature if you have run that code sample and know it works with the next version.

docs/sources/k6/next/javascript-api/k6-browser/locator/_index.md
AgnesToulet: Here's what's written in the Playwright documentation: https://playwright.dev/docs/locators#strictness
We could write something similar like:
```suggestion
- `strict` mode is enabled for all `locator` methods that imply some target DOM element, which means that if more than one element matches the given selector it will throw an error.
```

There are a few more in Playwright that we don't support yet like `allInnerTexts`, `allTextContents`, `filter`, `evaluateAll`. 

@ankur22 Any opinion?

ankur22: Would it work if we ensured that we let the reader know `strict` doesn't work on each individual API docs page instead of in the `locator` APIs index page?

Also, how about:

```suggestion
- `strict` mode is enabled for all `locator` methods that are expected to target a single DOM element, meaning that if more than one element matches the given selector, an error will be thrown.
```

heitortsergent: I think that's a good solution @ankur22. We can do that in a separate PR.

I also like the suggested change. 👍

AgnesToulet: I've updated the wording following Ankur's suggestion :+1: We can add more details in each function page in a separate PR (I guess adding the strict mode warning in each function that does use it?).

heitortsergent (approved):

PR #2028, opened 2025-08-11
k6-operator: add troubleshooting to image with extension

yorugac:
## What?

People hit issues with custom images in k6-operator, when making a mistake in `Dockerfile`. This PR adds some small bits of info to improve handling that case.

## Checklist

- [X] I have made my changes in the `docs/sources/k6/next` folder of the documentation.
- [X] I have reflected my changes in the `docs/sources/k6/v{most_recent_release}` folder of the documentation.
- [X] I have reflected my changes in the relevant folders of the two previous k6 versions of the documentation (if still applicable to previous versions).


docs/sources/k6/next/set-up/set-up-distributed-k6/usage/extensions.md
```diff
@@ -59,3 +59,12 @@ spec:
 ```
 
 Note that this examples overrides the default image with `k6-extended:latest`, and it includes environment variables that are required by the `xk6-output-influxdb` extension.
+
+## Troubleshooting
+
+If a `TestRun` with a custom image does not start the test and the k6 Operator [logs an error about the initializer](https://grafana.com/docs/k6/latest/set-up/set-up-distributed-k6/troubleshooting/#an-error-on-reading-output-of-the-initializer-pod), check the following command:
+```sh
+docker run --entrypoint sh k6-extended:local -c "ls -l && k6 version"
+```
```
heitortsergent:
```suggestion
If a `TestRun` with a custom image doesn't start the test and the k6 Operator [logs an error about the initializer](https://grafana.com/docs/k6/latest/set-up/set-up-distributed-k6/troubleshooting/#an-error-on-reading-output-of-the-initializer-pod), run the following command:

```sh
docker run --entrypoint sh k6-extended:local -c "ls -l && k6 version"
```
```

File: docs/sources/k6/next/set-up/set-up-distributed-k6/usage/extensions.md [hunk 2]
```diff
@@ -59,3 +59,12 @@ spec:
...
+
+This command should execute without an error.
```
heitortsergent: Is there any guidance we could add here in case the user sees an error?

yorugac: If there's an error, it means they misconfigured their Dockerfile. So basically, it's a ~ look at the example Dockerfile at the top of this webpage and see the diff in you Dockerfile.
Next (by logic) would be explaining Dockerfile line by line, which is probably out-of-scope...

yorugac: Also, the main error they might hit is literally `k6 command not found`. This error should be pretty self-explanatory, IMO.

IOW, I don't know what else to add here without being unnecessary verbose :smile:

heitortsergent: I think those two things you mentioned should be enough, and we don't have to explain the Dockerfile line by line. Maybe something like:

"The command should execute without an error. If there's an error, make sure that the Dockerfile is correctly configured, and that k6 is correctly installed in your cluster."

yorugac: It's not about cluster though :sweat_smile: 
```suggestion
This command should execute without an error. If there's an error, resolve it by making sure that the `Dockerfile` is correctly configured. Use the sample above as a reference.
```
So something like this?

heitortsergent (approved):

PR #2038, opened 2025-08-18
Mount actual latest semantic version at /docs/k6/latest/

jdbaldry:
The introduction of this workflow in #2014 mounted the "next" documentation at the `/docs/k6/latest/` location and @heitortsergent pointed out that this could confuse people.

This version of the workflow mounts the latest version semantic-version-like directory at that location instead for an experience that matches the website behavior.

The behavior was implemented in https://github.com/grafana/writers-toolkit/pull/1217.

I'll drop the CI triggering test commit before merge

### DIFF
--- a/.github/workflows/deploy-pr-preview.yml
+++ b/.github/workflows/deploy-pr-preview.yml
@@ -12,7 +12,7 @@ on:
 jobs:
   deploy-pr-preview:
     if: "!github.event.pull_request.head.repo.fork"
-    uses: grafana/writers-toolkit/.github/workflows/deploy-preview.yml@main
+    uses: grafana/writers-toolkit/.github/workflows/deploy-preview.yml@jdb/deploy-preview
     with:
       branch: ${{ github.head_ref }}
       event_number: ${{ github.event.number }}
--- a/.github/workflows/deploy-pr-preview.yml
+++ b/.github/workflows/deploy-pr-preview.yml
@@ -12,7 +12,7 @@ on:
 jobs:
   deploy-pr-preview:
     if: "!github.event.pull_request.head.repo.fork"
-    uses: grafana/writers-toolkit/.github/workflows/deploy-preview.yml@jdb/deploy-preview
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-pr-preview.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           {
             "relative_prefix": "/docs/k6/latest",
             "repo": "k6-docs",
-            "source_directory": "docs/sources/k6/next",
+            "source_directory": "docs/sources/k6/<LATEST>",
```
ankur22, line 32: What's the impact of this change? I can't tell just by looking at the change itself.

ankur22: I can see "TESTING AGAIN" in the previews... i think that this wouldn't show up if it was pointing to `next`? When we change something in `next` and not in the "`latest`" directory, is it possible to preview that too?

oleiade:
I'm confused as to if I should review this (`TESTING AGAIN`), and why 😄 
I'll respectfully dismiss myself from this review 🙇🏻

heitortsergent (approved):
Thank you so much for updating the deploy preview for this repo @jdbaldry. 🙇

jdbaldry:
Apologies reviewers for the poor PR description, it assumed the context from the previous PR. I've updated it now and things should make more sense.

.github/workflows/deploy-pr-preview.yml
```diff
@@ -12,7 +12,7 @@ on:
 jobs:
   deploy-pr-preview:
     if: "!github.event.pull_request.head.repo.fork"
-    uses: grafana/writers-toolkit/.github/workflows/deploy-preview.yml@main
+    uses: grafana/writers-toolkit/.github/workflows/deploy-preview.yml@jdb/deploy-preview
```
jdbaldry:
```suggestion
    uses: grafana/writers-toolkit/.github/workflows/deploy-preview.yml@main
```

PR #2044, opened 2025-08-28
[docs] Update k6 Studio usage collection docs

going-confetti:
## What?

- Add a table with the description of every event tracked in Grafana k6 Studio
- Add info about the new `script_copied` event and the new `isLoggedIn` event property

## Checklist

### DIFF
--- a/docs/sources/k6-studio/set-up/usage-collection.md
+++ b/docs/sources/k6-studio/set-up/usage-collection.md
@@ -17,18 +17,18 @@ A usage report includes the following information:
 - The k6 Studio version (for example, `1.7.0`).
 - The timestamp of when the data was collected (for example, `2024-10-22T16:02:56.261Z`).
 
-| Event name           | Description                                           | Event payload                     |
-| -------------------- | ----------------------------------------------------- |---------------------------------- |
-| `app_installed`      | k6 Studio is installed and opened for the first time  |                                   |
-| `user_logged_in`     | The user logs in to Grafana Cloud                     |                                   |
-| `recording_created`  | A new recording is created                            |                                   |
-| `recording_imported` | A recording is imported                               |                                   |
-| `generator_created`  | A new generator is created                            |                                   |
-| `generator_updated`  | An existing generator is updated                      | The total number of test rules per rule type and the total number of disabled rules                     |
-| `script_copied`      | A script is copied into the clipboard                 |                                   |
-| `script_exported`    | A script is exported                                  |                                   |
-| `script_validated`   |  A script is validated                                |                                   |
-| `script_run_in_cloud`|  A script is run in the cloud                         |                                   |
+| Event name            | Description                                          | Event payload                                                                       |
+| --------------------- | ---------------------------------------------------- | ----------------------------------------------------------------------------------- |
+| `app_installed`       | k6 Studio is installed and opened for the first time |                                                                                     |
+| `user_logged_in`      | The user logs in to Grafana Cloud                    |                                                                                     |
+| `recording_created`   | A new recording is created                           |                                                                                     |
+| `recording_imported`  | A recording is imported                              |                                                                                     |
+| `generator_created`   | A new generator is created                           |                                                                                     |
+| `generator_updated`   | An existing generator is updated                     | The total number of test rules per rule type and the total number of disabled rules |
+| `script_copied`       | A script is copied into the clipboard                |                                                                                     |
+| `script_exported`     | A script is exported                                 |                                                                                     |
+| `script_validated`    | A script is validated                                |                                                                                     |
+| `script_run_in_cloud` | A script is run in the cloud                         |                                                                                     |
 
 We use [Sentry](https://sentry.io/) to collect crash reports and error data. That includes:
 


docs/sources/k6-studio/set-up/usage-collection.md
```diff
@@ -8,23 +8,29 @@ weight: 300
...
+| `recording_imported` | A recording is imported                               |                                   | 1.6.0              |
+| `generator_created`  | A new generator is created                            |                                   | 1.6.0              |
+| `generator_updated`  | An existing generator is updated                      | `rules` - object containing the total number of test rules per rule type and the total number of disabled rules                     | 1.6.0              |
+| `script_copied`      | A script is copied into the clipboard                 |                                   | 1.7.0              |
+| `script_exported`    | A script is exported                                  |                                   | 1.6.0              |
+| `script_validated`   |  A script is validated                                |                                   | 1.6.0              |
+| `script_run_in_cloud`|  A script is run in the cloud                         |                                   | 1.6.0              |
```
going-confetti: @heitortsergent - do you have any suggestions on how we can present this in a better way?

I believe listing every event in the list above will eventually becomes unsustainable (especially if we were to include the event properties), so I moved the detailed event info into a table. I'm not quite sure about the `Additional event properties` column:
- Does this format work or is there a better alternative?
- in the case of `generator_updated` – is this level of detail enough or should I be more explicit by describing the internal structure of the `rules` object`?

heitortsergent: I think the table works well @going-confetti.

One suggestion here is, since only the `generator_updated` has information in the "Additional event properties" column, I think you could also remove the column and just update the description for `generator_updated` to something like:

"An existing generator is updated. It also includes a `rules` object that contains the total number of test rules per rule type and the total number of disabled rules".

Another alternative would be adding an extra row to the table for `generator_updated.rules` to have a dedicated row for the `rules` object.

What do you think?

going-confetti: @heitortsergent 
> since only the generator_updated has information in the "Additional event properties" column

This will likely change soon - one of the changes we're considering is tracking whether the script is a browser- or a protocol-level script in all script-related events.

After some thought, I decided to:
- Remove the `Version introduced` column as it's not very useful IMO (plus, the docs are only for the latest version of k6 Studio)
- Describe the additional "event properties" in plain text without mentioning of property names:

  Replace this:
  > `rules` - object containing the total number of test rules per rule type and the total number of disabled rules
  
  with this:
  
  > The total number of test rules per rule type and the total number of disabled rules


Would these changes make sense?

heitortsergent: Yeah, that makes sense. 👍

docs/sources/k6-studio/set-up/usage-collection.md
```diff
@@ -10,21 +10,26 @@ By default, k6 Studio sends anonymous usage reports, as well as crash reports an
...
-  - `user_logged_in`: when a user logs in to Grafana Cloud.
-  - `recording_created`: when a new recording is created.
-  - `recording_imported`: when a recording is imported.
-  - `generator_created`: when a new generator is created.
-  - `generator_updated`: when a generator is updated.
-  - `script_exported`: when a script is exported.
-  - `script_validated`: when a script is validated.
-  - `script_run_in_cloud`: when a script is run in the cloud.
+- The event name (see the table below for a list of possible event names and their details).
+- The event payload - additional information about the event (every event includes the boolean property `isLoggedIn`).
```
heitortsergent:
```suggestion
- The event name. Refer to the table below for a list of event names and details.
- The event payload - additional information about the event. Every event includes the boolean property `isLoggedIn`.
```

docs/sources/k6-studio/set-up/usage-collection.md
```diff
@@ -10,21 +10,26 @@ By default, k6 Studio sends anonymous usage reports, as well as crash reports an
...
+| `user_logged_in`     | The user logs in to Grafana Cloud                     |                                   |
+| `recording_created`  | A new recording is created                            |                                   |
+| `recording_imported` | A recording is imported                               |                                   |
+| `generator_created`  | A new generator is created                            |                                   |
+| `generator_updated`  | An existing generator is updated                      | The total number of test rules per rule type and the total number of disabled rules                     |
+| `script_copied`      | A script is copied into the clipboard                 |                                   |
+| `script_exported`    | A script is exported                                  |                                   |
+| `script_validated`   |  A script is validated                                |                                   |
+| `script_run_in_cloud`|  A script is run in the cloud                         |                                   |
```
heitortsergent:
```suggestion
| Event name            | Description                                          | Event payload                                                                       |
| --------------------- | ---------------------------------------------------- | ----------------------------------------------------------------------------------- |
| `app_installed`       | k6 Studio is installed and opened for the first time |                                                                                     |
| `user_logged_in`      | The user logs in to Grafana Cloud                    |                                                                                     |
| `recording_created`   | A new recording is created                           |                                                                                     |
| `recording_imported`  | A recording is imported                              |                                                                                     |
| `generator_created`   | A new generator is created                           |                                                                                     |
| `generator_updated`   | An existing generator is updated                     | The total number of test rules per rule type and the total number of disabled rules |
| `script_copied`       | A script is copied into the clipboard                |                                                                                     |
| `script_exported`     | A script is exported                                 |                                                                                     |
| `script_validated`    | A script is validated                                |                                                                                     |
| `script_run_in_cloud` | A script is run in the cloud                         |                                                                                     |
```

I just formatted the table here.

PR #2074, opened 2025-09-23
Dynamic docs from catalog

Lantero:
This is the OSS equivalent of https://github.com/grafana/website/pull/27448, which works fine.

The problem is that these shared layouts are only hosted in the website, which means the local preview is broken here, even though the final website would be ok.

@heitortsergent what is the recommended approach for using a shared layout in a place like this?

### DIFF
--- a/docs/sources/k6/next/extensions/explore.md
+++ b/docs/sources/k6/next/extensions/explore.md
@@ -7,20 +7,17 @@ weight: 01
 
 # Explore extensions
 
-Extensions can be used to extend the core k6 features to fit your performance-testing use cases. There are two main categories of extensions:
-
-| Category      | Maintainer   | Audited | Cloud support |
-| ------------- | ------------ | ------- | ------------- |
-| **Official**  | Grafana Labs | ✅ Yes  | ✅ Yes        |
-| **Community** | Community    | ✅ Yes  | ⚠️ Partial\*  |
-
-\* _Partial_ cloud support means that the extensions can be used in the CLI with the `--local-execution` mode, and on Private Load Zones with a custom image. They don't work on Grafana Cloud k6 Public Load Zones.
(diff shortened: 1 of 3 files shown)


docs/sources/k6/next/extensions/explore.md
```diff
@@ -18,7 +18,7 @@ Extensions can be used to extend the core k6 features to fit your performance-te
 
 ## Official extensions
 
-{{< docs/shared source="k6" lookup="official-extensions.md" version="<K6_VERSION>" >}}
+{{< docs/shared source="k6" lookup="official-extensions.html" version="<K6_VERSION>" >}}
```
codebien: Image

I guess this change is generating this error.

Lantero: Yeah, I thought it was working in one of the previous previews, but it's not, I will put this back as a draft until I get it working

heitortsergent:
@jdbaldry or @robbymilo, do you have any suggestions on how we can make the shared files work in the k6-docs without breaking the local/preview builds? 🤔

jdbaldry:
Are they not working with the `docs/shared` lookup? Did they break in https://github.com/grafana/k6-docs/pull/2038?

Lantero:
> Or is the intent that you want to maintain the shortcode in the website repository as implemented in [grafana/website#27448](https://github.com/grafana/website/pull/27448)?
> 
> If that's the case, when we release a new version of the `grafana/docs-base` image, the shortcode will be available to local and preview builds.

@jdbaldry does that mean that if I merge https://github.com/grafana/website/pull/27448 (Which works in its current form), a new `grafana/docs-base` image will be built, and then the builds/renders for this PR (`k6-docs`) will start working?

codebien (approved):

PR #2203, opened 2026-03-06
docs: use async functions in gRPC streaming examples

bonnywelsford-source:
## Summary

- Adds a new "Streaming requires async functions" section explaining that gRPC streaming event handlers can only fire when the call stack is empty, so the default function must be `async`.
- Updates the server streaming and client streaming examples to use `async`/`await` with `Promise`, replacing the previous synchronous `sleep()`-based approach.
- Adds `stream.on('error', ...)` handlers to both examples.

Addresses customer feedback that the gRPC streaming docs didn't explain the need for async functions, making it difficult to get streaming to work.

Applies to `next/` and `v1.6.x`.

## Test plan

- [ ] Verify the page builds locally (`npm start`, visit http://localhost:3002/docs/k6/next/using-k6/protocols/grpc/)
- [ ] Confirm the new "Streaming requires async functions" section renders between "Unary gRPC requests" and "Server gRPC streaming"
- [ ] Review the server and client streaming code examples for correctness
- [ ] Check the v1.6.x version matches the next/ version

Made with [Cursor](https://cursor.com)


heitortsergent:
@bonnywelsford-source this looks good to me, but it would be nice to get one of the engineers to double-check this. @andrewslotin do you know if anyone on the squad might be able to take a quick look at this? 🙇 

We also would need to update the v1.7.x version of the page before merging.

mstoykov (approved):
Thank you for the fixes 🙇

bonnywelsford-source:
Closing in favor of a replacement PR with a single verified/signed commit on current `main` (includes `v2.0.x` + backports to `v1.7.x`/`v1.6.x`). The original branch could not be force-pushed or merged because of the signed-commits requirement.

Replacement: will link in a moment.

bonnywelsford-source:
Replacement PR: https://github.com/grafana/k6-docs/pull/2280

PR #2221, opened 2026-04-22
docs: add k6/x/tcp API documentation

szkiba:
## What

This PR adds comprehensive documentation for the **xk6-tcp** extension (k6/x/tcp import path), providing first-class TCP protocol support for k6 load testing.

## Checklist

- [x] I have described the changes I've made in the What? section above.
- [ ] I have run the `npm start` command locally and verified that the changes look good. (failed to run)


docs/sources/k6/next/javascript-api/k6-x-tcp/socket/connect.md
```diff
@@ -0,0 +1,83 @@
+---
+title: 'Socket.connect( port, [host] )'
```
mstoykov, line 2: Is there a reason why we are providing a blocking API for this or anything else that will block the event loop and will potentially break the script when it is also using `.on` or any other async operations?

szkiba: Several APIs provide synchronous and asynchronous functions for the same task. I think this is good, the user can decide which one to use. In many cases, it is simpler and more convenient to use synchronous operations.

mstoykov: Several of the APIs do this for historical reasons. 

Providing syncrhonous APIs has been deprecatred. We have them because otherwise it will be a breaking change and we haven't gotten to rewriting the remaing parts out. 

I would really prefer if we do not provide ways for users to write code that will be buggy, so they can skip adding `await`.

This also adds to maitainance which IMO is not a good idea

cc @grafana/k6-core

codebien: > it is simpler and more convenient to use synchronous operations

@szkiba I honestly struggle to see how this might be true for a networking-focused extension. Can you provide an example, please?

szkiba: > Providing syncrhonous APIs has been deprecatred. 

Oh I didn't know that, sorry. Could you please send me the link to the document where it is written that the syncrhonous api is deprecated. I would like to link it when I refactor the xk6-tcp API and remove the blocking operations.
BTW can I get rid of the `Async` suffixes too? So it can just be `connect()` which is async?

szkiba: > @szkiba I honestly struggle to see how this might be true for a networking-focused extension. Can you provide an example, please?

@codebien I don't want to argue for convenience, if it's deprecated then it's deprecated.

PR #2225, opened 2026-04-24
add deploy preview permissions

robbymilo:
Adds permissions to fix deploy previews.


jdbaldry (approved):
Approved assuming we drop the docs/sources file

PR #2288, opened 2026-08-04
Document TextEncoder and TextDecoder support

oleiade:
## Summary

- replace manual string/byte conversion workarounds across Web Crypto, Streams, browser, MQTT, and TCP examples with direct encoder and decoder calls
- recognize the standard text encoding globals when linting documentation examples

## Why

k6 now supports the standard text encoding globals. The documentation should expose that capability and stop teaching workarounds that can mishandle Unicode data.

## Validation

- Prettier and ESLint pass for all changed documentation
- `python3 -m unittest scripts/md-k6.py` passes
- all changed `next` Markdown files pass `md-k6.py` with the v2 master build, including the browser example
- `git diff --check` passes


oleiade:
Note that this is red because the code blocks are evaluated against a version of k6 that doesn't have TextEncoder and TextDecoder, yet.

inancgumus (approved):

mstoykov (approved):

PR #2304, opened 2026-08-19
docs: add k6 current release and support lifecycle information

cristianoventura:
## What?

As part of phasing out legacy k6 versions, we'd like to publicly describe to users what versions are currently supported, what type of support users should expect, and when the EoL for a specific version will be.

This PR includes a table that describes all k6 releases and when support ends.

## Checklist


bonnywelsford-source:
Thanks for the draft! I think that putting current versions and EOL in one table is really useful.

For public docs, I think you should drop the badge shortcodes. Those styles are for product and stage labels (OSS/Cloud, GA/Beta), not version support. They read as the wrong meaning here (especially Deprecated using the Cloud color). Grafana’s version support tables use plain text for the support state. Writers’ Toolkit: https://grafana.com/docs/writers-toolkit/write/shortcodes/

Also, the table separator has five columns (|---|---|---|---|---|) against four headers, so preview shows an empty column. 

Maybe, consider moving the table up under **Support and development policies**, and match that section’s wording for “supported.”

Appreciate you writing this. Thank. you!

PR #2317, opened 2026-09-11
docs: add byte encoding and Set operations

inancgumus:
## What?

Documents built-in byte encoding and set comparisons available in k6 v2.3.0.

## Why?

Helps users prepare binary test data and compare expected results without additional libraries.

## Checklist

## Related PR(s)/Issue(s)

- grafana/k6#6382


inancgumus:
@mstoykov Could you review the byte-encoding and Set documentation for your change in grafana/k6#6382?

mstoykov:
We do not in general document (in the documentation) compatibility with JS apart from in very general. So I do not think that this will be a good idea to be the start.

Arguably some examples can be changed to use this operaitons, but a part for it is not really a thing that I woull advocaet for.

THis still should be part fo the release notes.

inancgumus:
@mstoykov Should we close this one then?

